PR #25, opened 2025-02-12
Add `chat_clear()`

### LINKED ISSUE #13: Add a way to clear or reset a chat
I'm looking for something like `chat_reset()` so I can have a way to start a fresh chat without requiring a page reload.

It may already be possible to do this with the right incantation to `chat_message()`, but it seems like this could be a top-level API feature in its own right.

cpsievert:
Closes #13


cpsievert:
@jcheng5 I'm gonna merge before a review since I'd like to demo it tomorrow, but please feel free to review and I'll address in a follow up PR

PR #28, opened 2025-02-20
feat: Add `chat_enable_bookmarking(id, client, ..., update_on_input = TRUE, update_on_response = TRUE)`

### LINKED ISSUE #27: feat: Add `set_chat_client()`
**Updated to `set_chat_client()` given comment below

----------

Method to add your (currently only) `{ellmer}` chat model the UI.

Ex API: 

```r
shinychat::set_chat_model(id, model, ..., bookmark = TRUE)
```

Assuming your {shinychat} ID is `"chat"`...

Goals:
* Error if `model` is not an `{ellmer}` chat model. 
	* Provide error message on submitting GH issue for extension?
* Given `bookmark = TRUE`,
	* Require bookmarking to be enabled: `stopifnot(shiny::getShinyOption("bookmarkStore", "none") != "none")`
		*  Provide example in docs on how to exclude some input in the example app.
			```r
          	# Escape bookmarking all inputs
          	shiny::setBookmarkExclude("big_input_name")
             ```
	* Save minimal chat information: `role`, `contents`
		* `shiny::onBookmark()` callback
		* `content` should be the markdown of each turn's contexts. 
			* Using markdown escapes us from different content types and is more than just text. [`ellmer::contents_markdown()`](https://ellmer.tidyverse.org/reference/contents_text.html) will allow us to escape the s7 serialization difficulty (related https://github.com/RConsortium/S7/issues/225).
	* Restore minimal chat information: `role`, `contents`
		* `shiny::onRestore()` callback
	* Updates the url on user input / server output change
* Opens the door for future integration with the model

Bonus goals:
* Update the URL on markdown stream update
* Send multiple chat messages at once, not one at a time. 
	* Avoids multiple `shinychat::chat_append("chat", stream)` calls.
	* Single scroll to bottom call. This currently is broken when many updates happen within the initial transition.

Possible approaches:
* Use `shiny::getQueryString()` and `shiny::updateQueryString()`/`httpuv::encodeURIComponent()` to quickly insert chat information within the query string.
	* This would not require bookmarking to be enabled.
	* Feels hacky (but a great approach for py-shiny until full bookmarking is enabled!)

----------------

### Rejected ideas

We had working code to restore the UI dynamically, but the user story is much simpler to add a single call in the server. 
* Restoring the UI would also require reaching inside shiny for `shiny:::getCurrentRestoreContext()$values`.
* Restoring the UI would also require both a UI and a server flag (of some kind). With `shinychat::set_chat(ID, model)`, we can restore values quickly and in a single location.
```r
ui <- page_fillable(
	shinychat::chat_ui("chat", height = "100%", bookmark = TRUE)
)

server <- function(input, output) {
    model <- ellmer::...()

    shinychat::set_chat_bookmark("chat", model, TRUE)
    # or 
    shinychat::set_chat_model("chat", model, bookmark = TRUE)
}
```
The complexity didn't outweigh the very minor blip of chat messages. This can be significantly reduced if "send multiple chat messages" is added.

### LINKED ISSUE #55: Shinychat bookmarking serialized information hits url limit
Browsers support ~60k characters. I had asked 2 questions/answers and was at 8k. It is not sustainable.

schloerke:
Fixes #27 
Fixes #55 

Reprex app:

```r
library(shiny)
library(bslib)
library(shinychat)

ui <- function(...) {
  page_fillable(
    chat_ui("chat", fill = TRUE)
  )
}

  chat_client <- ellmer::chat_ollama(
    system_prompt = "Important: Always respond in a limerick",
    model = "qwen2.5-coder:1.5b",
    echo = TRUE
  )
  # Let the UI know about the client
  chat_enable_bookmarking("chat", chat_client)

    chat_append("chat", stream)
  })
}

# Enable bookmarking!
shinyApp(ui, server, enableBookmarking = "url")

```


R/set_chat_client.R
```diff
@@ -0,0 +1,301 @@
...
+  auto_update = TRUE,
+  session = getDefaultReactiveDomain()
+) {
+  rlang::check_dots_empty()
+
+  stopifnot(is.character(id) && length(id) == 1)
+
+  # Verify bookmark store is not "none"
+  bookmarkStore <- shiny::getShinyOption("bookmarkStore", "none")
+  if (bookmarkStore == "none") {
```
cpsievert: AFAIK, there are 3 valid values: `url`, `server`, and `disable`. And we should also error for `disable`.
```suggestion
  bookmarkStore <- shiny::getShinyOption("bookmarkStore", default = "disable")
  if (!bookmarkStore %in% c("url", "server")) {
```

R/set_chat_client.R
```diff
@@ -0,0 +1,301 @@
...
+  # Exclude works with bookmark names
+  excluded_names <- session$getBookmarkExclude()
+  ns_user_input <- paste0(id, "_user_input")
+  if (!(ns_user_input %in% excluded_names)) {
+    session$setBookmarkExclude(c(excluded_names, ns_user_input))
+  }
+
+  # Save
+  cancel_on_bookmark <- session$onBookmark(function(state) {
+    if (id %in% names(state$values)) {
```
cpsievert: I would add a prefix/suffix to the id. I don't think it matters today since there is no `input` value tied to `id`, but there might be someday

R/set_chat_client.R
```diff
@@ -0,0 +1,301 @@
...
+
+    turns_list <- lapply(turns, function(turn) {
+      list(
+        role = turn@role,
+        # Convert everything to a single markdown string (including images!)
+        contents = ellmer::contents_markdown(turn),
+        tokens = turn@tokens
+      )
+    })
+    state$values[[id]] <- turns_list
```
cpsievert: Can we just save the whole `client` object as is when `bookmarkStore` is `"server"`?

Also, even for `"url"`, I think we'll need that approach you had to serializing R6 objects for the `contents` since  `contents_markdown()` will be too lossy (sorry for suggesting it in the 1st place)

R/set_chat_client.R
schloerke: Inputs and values do not mix.

(Marking resolved)

R/set_chat_client.R
schloerke: Yup! Good call!

schloerke: When doing URL formatting, contents_markdown() is the easiest way to get from text, image, or any other type to something that can be serialized.

I'd like to stay with `contents_markdown()` for now when doing URL formatting.

schloerke:
Talking with `@icarusz` and `@wch` , I am in agreement that we should change the API:

This change below (`set_chat_client(id, client, ..., bookmark = "auto")` -> chat_bookmark(id, client, ..., on_user_input = TRUE, on_stream_complete = TRUE)` will only really change the interface. The main logic will remain.

```r
library(shiny)
library(bslib)
# library(shinychat)
pkgload::load_all()

ui <- function(...) {
  page_fillable(
    chat_ui("chat", fill = TRUE)
  )
}

  chat_client <- ellmer::chat_ollama(
    system_prompt = "Important: Always respond in a limerick",
    model = "qwen2.5-coder:1.5b",
    echo = TRUE
  )

  # Let the UI know that shinychat should bookmark the chat_client
  shinychat::chat_bookmark("chat", chat_client) # on_user_input = TRUE, on_stream_complete = TRUE

    stream_promise <- chat_client$stream_async(input$chat_user_input)
    shinychat::chat_append("chat", stream_promise)
  })

}

# Enable bookmarking!
shinyApp(ui, server, enableBookmarking = "server")
```

If other features were to be made, they should have their own `shinychat::chat_FEATURE()` method.

R/set_chat_client.R
cpsievert: It seems that if you serialize tool requests/results in this way then you won't be able to restore them?

R/chat_bookmark.R
```diff
@@ -0,0 +1,321 @@
...
+  if (bookmark_store == "disable") {
+    rlang::abort(
+      paste0(
+        "Error: Shiny bookmarking is not enabled. ",
+        "Please enable bookmarking in your Shiny app either by calling ",
+        "`shiny::enableBookmarking(\"server\")` or by setting the parameter in ",
+        "`shiny::shinyApp(enableBookmarking = \"server\")`"
+      )
+    )
+  }
```
schloerke: Remaining question

schloerke:
@wch Looking for a review on the API: `shinychat::chat_bookmark(ID, CLIENT)` with [default parameters of `update_on_input = TRUE` and `update_on_response = TRUE`](https://github.com/posit-dev/shinychat/pull/28/files#diff-1c780db262b94b3868c84e10f3980abb58dff1c2c02fcfad19f2683287ff9a3cR13-R14).

These default parameters hide away the multiple calls to `session$doBookmark()`.

Ex equivalent server code: 
```r
  shinychat::chat_bookmark("chat", chat_client, update_on_input = FALSE, update_on_response = FALSE)
  
    session$doBookmark()
    stream_promise <- chat_client$stream_async(input$chat_user_input)
	stream_promise$then(function(stream_value) { session$doBookmark() })
    shinychat::chat_append("chat", stream_promise)
  })
```

Final server code:
```r
  shinychat::chat_bookmark("chat", chat_client)
  
    stream_promise <- chat_client$stream_async(input$chat_user_input)
    shinychat::chat_append("chat", stream_promise)
  })
```

wch:
The new API does look better. A few questions:

- The name `chat_bookmark` makes it sound like running that function will do the bookmarking. Maybe `chat_enable_bookmarking`?
- Another thing we should consider: We want to use this mechanism to (A) do bookmarking in the usual way that Shiny does it (B) allow the user to save/restore multiple chats, similar to the ChatGPT web interface. Will this API make sense for the second case as well?
- Suppose an app uses `update_on_input=TRUE` and there's a disconnection before the response is completed. When the chat is restored, is there be a way in the UI re-send the message to the server and get a response? If not, then restoring that conversation could leave the app in an unusable state, I think. (We should have the ability to re-send a message anyway, but that's outside the scope of this PR.)

schloerke:
✅ `chat_enable_bookmarking()`

schloerke:
> Will this API make sense for the second case as well?

If the chat is done within a single chat object, no. The data storage required is more complex than what a single ellmer object  wants.

If the chat is done within a module where given the sidebar's chat selected id, we could have a dynamic chat module that loads a chat given the chat ID. I believe this is a plausible approach

R/chat_enable_bookmark.R
```diff
@@ -0,0 +1,321 @@
...
+        list(
+          role = turn@role,
+          # Convert everything to a single markdown string (including images!)
+          contents = ellmer::contents_markdown(turn),
+          tokens = turn@tokens
+        )
+      })
+      state$values[[id]] <- turns_list
+    }
+  })
```
schloerke: We also need to save the UI here

R/chat_enable_bookmark.R
```diff
@@ -0,0 +1,321 @@
...
+    # TODO-barret-future; In shinychat, make this a single/internal custom message call to send all the messages at once (and then scroll)
+    lapply(client$get_turns(), function(turn) {
+      chat_append(
+        id,
+        # Use `contents_markdown()` as it handles image serialization
+        ellmer::contents_markdown(turn),
+        #  turn_info$contents,
+        role = turn@role
+      )
+    })
```
schloerke: We need to independently restore the UI here (not directly from ellmer)

R/chat_enable_bookmark.R
```diff
@@ -0,0 +1,321 @@
...
+      client$set_turns(turns)
+    }
+
+    # Set the UI
+    # TODO-barret-future; In shinychat, make this a single/internal custom message call to send all the messages at once (and then scroll)
+    lapply(client$get_turns(), function(turn) {
+      chat_append(
+        id,
+        # Use `contents_markdown()` as it handles image serialization
+        ellmer::contents_markdown(turn),
```
schloerke: This should ideally be a `{s7}` JSON "pack" method, similar to `jsonlite:::pack()`

schloerke:
@cpsievert There is no UI to save (and later restore) as the information is sent off and forgotten. 

We could in the future ask the browser about the current state, but I believe it is possibly for another PR down the road.

R/shinychat-package.R
```diff
@@ -5,9 +5,14 @@
 #' @importFrom coro async
 #' @importFrom htmltools tag css HTML
 #' @import rlang
+#' @import S7
 ## usethis namespace: end
 NULL
 
+# enable usage of <S7_object>@name in package code
+#' @rawNamespace if (getRversion() < "4.3.0") importFrom("S7", "@")
```
cpsievert: Is this because it just comes with R now?

R/markdown-stream.R
```diff
@@ -166,7 +168,6 @@ markdown_stream <- function(
   result
 }
 
-
```
cpsievert: Add back whitespace, just to reduce conflict potential with #56

R/client_state.R
```diff
@@ -0,0 +1,133 @@
+client_get_state <- new_generic(
+  "client_get_state",
+  "client",
+  function(client) {
+    result <- S7::S7_dispatch()
```
cpsievert:
```suggestion
    S7::S7_dispatch()
```

R/shinychat-package.R
schloerke: Yes? It is from S7's articles: https://rconsortium.github.io/S7/articles/packages.html#backward-compatibility

R/client_state.R
```diff
@@ -0,0 +1,133 @@
...
+
+method(client_get_state, S7::new_S3_class(c("Chat", "R6"))) <-
+  function(client) {
+    # if (is_server_bookmarkstore()) {
+    #   # Serialize the whole chat client
+    #   # However, this is a security leak
+    #   return(client)
+    # }
+
+    A <<- client$get_turns()
```
cpsievert:
```suggestion
```

R/client_state.R
```diff
@@ -0,0 +1,133 @@
...
+)
+
+is_server_bookmarkstore <- function() {
+  shiny::getShinyOption("bookmarkStore", "server") == "server"
+}
+is_url_bookmarkstore <- function() {
+  shiny::getShinyOption("bookmarkStore", "url") == "url"
+}
+
+method(client_get_state, S7::new_S3_class(c("Chat", "R6"))) <-
```
cpsievert: Interesting that `ellmer::Chat` is a `R6::R6Class()` , but this is `new_S3_class()`? I have know idea how/if this is supposed to work, but I'm willing to take your word for it :)

R/client_state.R
```diff
@@ -0,0 +1,133 @@
...
+    S7::S7_dispatch()
+  }
+)
+
+is_server_bookmarkstore <- function() {
+  shiny::getShinyOption("bookmarkStore", "server") == "server"
+}
+is_url_bookmarkstore <- function() {
+  shiny::getShinyOption("bookmarkStore", "url") == "url"
+}
```
cpsievert: Just in case someone is tempted to use these outside of a "we know bookmarking is enabled" context

```suggestion
is_server_bookmarkstore <- function() {
  shiny::getShinyOption("bookmarkStore", "") == "server"
}
is_url_bookmarkstore <- function() {
  shiny::getShinyOption("bookmarkStore", "") == "url"
}
```

R/client_state.R
```diff
@@ -0,0 +1,133 @@
...
+      client$get_turns(),
+      ellmer::contents_record,
+      chat = client
+    )
+
+    # Pre-serialize the contents so that when shiny:::toJSON() is called, it is stable.
+    # jsonlite::toJSON() is not stable as it is a lossy serialization. In addition, jsonlite::fromJSON() (which shiny:::safeFromJSON() uses) is not stable as it tries to make everything a data.frame.
+    #
+    # * `jsonlite::serializeJSON()` is a stable transformation
+    # * `jsonlite::unserializeJSON()` is a stable transformation
```
cpsievert: Move this comment to be just above the `serializeJSON()` call, and add a comment here about why we drop the `json` prop.

pkgdown/_pkgdown.yml
```diff
@@ -34,3 +34,7 @@ reference:
   contents:
   - markdown_stream
   - output_markdown_stream
+
+- title: Bookmark
```
cpsievert:
```suggestion
- title: Bookmark chat history
```

R/chat_enable_bookmarking.R
```diff
@@ -0,0 +1,245 @@
+#' Add Shiny bookmarking for shinychat
+#'
+#' @description
+#' Adds hooks to the Shiny chat given the LLM client.
```
cpsievert, line 4: We should probably have a big ol' warning/disclaimer somewhere that this currently only works as intended when the UI content matches what is stored in the `ellmer` turns. 

PS. I know we're considering `localstorage` as a potential long-term solution to this issue, which would be great.

Just throw it out there though, something I just thought of is that if the append knew about the `chat_client`:

```r
chat_append_message("id", chat_client$stream("foo"), chat_client)
```

Then, when the message ends, we could do something like:

```r
turns <- chat$get_turns()
# When the stream gets exhausted, the last turn represents "this" assistant turn 
last_turn <- turns[[length(turns)]]
key <- rlang::hash(last_turn@contents)
session$userData[[key]] <- final_ui_content_string
```

Then, when restoring the bookmark, we could lookup the content string from `session$userData`? There is a small risk of hash collisions doing it just based on contents, but we can probably workaround that? Anyway, just wanted to offer it as an idea to consider at this point and see what you think.

R/chat_enable_bookmarking.R
```diff
@@ -0,0 +1,245 @@
...
+#'
+#' If either `bookmark_on_input` or `bookmark_on_response` is `TRUE`, the Shiny
+#' App's bookmark will be automatically updated without showing a modal to the
+#' user.
+#'
```
cpsievert: We should probably have a big ol' warning/disclaimer somewhere around here that this currently only works as intended when the UI content matches what is stored in the `ellmer` turns. 

PS. I know we're considering `localstorage` as a potential long-term solution to this issue, which would be great.

Just throw it out there, something I just thought of is that if the append knew about the `chat_client`:

```r
chat_append_message("id", chat_client$stream("foo"), chat_client)
```

Then, when the message ends, we could do something like:

```r
turns <- chat$get_turns()
# When the stream gets exhausted, the last turn represents "this" assistant turn 
last_turn <- turns[[length(turns)]]
key <- rlang::hash(last_turn@contents)
session$userData[[key]] <- final_ui_content_string
```

Then, when restoring the bookmark, we could do a lookup from `session$userData`? There is a small risk of hash collisions doing it just based on contents, but we can probably workaround that? Anyway, just wanted to offer it as an idea to consider at this point and see what you think.

R/chat_enable_bookmarking.R
cpsievert: This isn't a very informative description

R/shinychat-package.R
cpsievert: Thanks!

R/markdown-stream.R
```diff
@@ -166,7 +168,6 @@ markdown_stream <- function(
   result
 }
```
schloerke: (Add back whitespace)
```suggestion


```

R/client_state.R
```diff
@@ -0,0 +1,133 @@
+client_get_state <- new_generic(
+  "client_get_state",
+  "client",
+  function(client) {
+    result <- S7::S7_dispatch()
+  }
+)
```
schloerke:
```suggestion
client_get_state <- new_generic("client_get_state", "client")
```

R/client_state.R
schloerke: Using suggestion below

R/client_state.R
```diff
@@ -0,0 +1,133 @@
...
+
+client_set_state <- new_generic(
+  "client_get_state",
+  "client",
+  function(client, state) {
+    S7::S7_dispatch()
+  }
+)
```
schloerke:
```suggestion
client_set_state <- new_generic("client_get_state", "client")
```

R/client_state.R
schloerke: Bad autocomplete. Good catch!

R/client_state.R
schloerke: Garrick pointed me to it. Because the `Chat` class is a "class" on the object, then standard s3 dispatch will work. 

Also, `ellmer:::Chat` is not exported as of today: https://github.com/tidyverse/ellmer/blob/2c8364f5efddca492334a2ea9b68ed70cea06f29/NAMESPACE

R/chat_enable_bookmarking.R
schloerke: Warning added. Issue created.

Resolving

R/chat_enable_bookmarking.R
```diff
@@ -0,0 +1,249 @@
...
+#' @description
+#' Adds Shiny bookmarking hooks to save and restore the \pkg{ellmer} chat `client`.
+#'
+#' If either `bookmark_on_input` or `bookmark_on_response` is `TRUE`, the Shiny
+#' App's bookmark will be automatically updated without showing a modal to the
+#' user.
+#'
+#' Note: The chat UI is not saved within the bookmark and is only restored
+#' directly from the `client`'s messages. If any transient state (e.g. Shiny
+#' inputs and outputs) has been added to the chat UI, they will not be properly restored. Please add your own `session$onRestore()` (and possibly `session$onBookmark`) handler to restore any additional state.
```
cpsievert: Probably needs work wrapping

```suggestion
#' Note: Only the `client`'s chat state is saved/restored in the bookmark. If the `client`'s state 
#' doesn't properly capture the chat's UI (i.e., a transformation is applied in-between receiving and displaying
#' the message), then you may need to implement your own `session$onRestore()` (and possibly `session$onBookmark`) handler to restore any additional state.
```

cpsievert (commented):
LGTM once final comment is addressed, thanks!

cpsievert (approved):

PR #36, opened 2025-04-02
feat: Add `chat_app()`, `chat_mod_ui()` and `chat_mod_server()`

gadenbuie:
For tidyverse/ellmer#397

* Moves the implementation of `ellmer::live_browser()` to shinychat as `shinychat::chat_app()`.

* Adds a simple Shiny module, that can be initialized from a chat `client` or can be intialized in the same way as `chat_ui()`. The server-side logic connects the user input and wraps `client$stream_async() |> chat_append()`.

The module also uses an `ExtendedTask` to avoid blocking reactivity; the docs include an example with simultaneous Claude/OpenAI chats that can run in parallel.


cpsievert:
LGTM, but would also be great to have @jcheng5's take, especially in regard to `ExtendedTask` and whether he'd be open to `chat_append()` using it (since it seems like it'll be necessary for prompting the user for input during tool execution)

cpsievert (approved):

R/chat_app.R
```diff
@@ -0,0 +1,216 @@
...
+      promises::promise(function(resolve, reject) {
+        tryCatch(
+          resolve(client$stream_async(user_input)),
+          error = function(err) {
+            reject(err)
+          }
+        )
+      })$then(function(stream) {
+        chat_append(ui_id, stream)
+      })
```
jcheng5:
```suggestion
      promises::then(client$stream_async(user_input), function(stream) {
        chat_append(ui_id, stream)
      })
```

R/chat_app.R
```diff
@@ -0,0 +1,216 @@
...
+
+    shiny::reactive({
+      if (append_stream_task$status() == "success") {
+        client$last_turn()
+      }
+    })
+  })
+}
+
+notify_error <- function(id, session = shiny::getDefaultReactiveDomain()) {
```
jcheng5: Please emit the error/stack to the console as well. I've been frustrated in the past by my chatbot erroring and there being nothing actionable in the console.

R/chat_app.R
```diff
@@ -0,0 +1,216 @@
...
+  function(err) {
+    needs_sanitized <-
+      isTRUE(getOption("shiny.sanitize.errors")) &&
+      !inherits(err, "shiny.custom.error")
+    if (needs_sanitized) {
+      msg <- "**An error occurred.** Please try again or contact the app author."
```
jcheng5, line 195: Do you know if it's actually possible to try again after a failure? I've been thinking, we probably would need to clear out Ellmer's turns until you get back to the last successful, non-too-call API response?

R/chat_app.R
```diff
@@ -0,0 +1,216 @@
...
+      session = session
+    )
+  }
+}
+
+strip_ansi <- function(text) {
+  # Matches codes like "\x1B[31;43m", "\x1B[1;3;4m"
+  ansi_pattern <- "(\x1B|\x033)\\[[0-9;?=<>]*[@-~]"
+  gsub(ansi_pattern, "", text)
+}
```
jcheng5, line 224: Do you know, are these color codes?

R/chat_app.R
gadenbuie: Will this catch errors that happen before `$stream_async()` returns a promise?

I don't think so -- I had something quite similar initially and needed this pattern to ensure we caught very early errors in the `client$stream_async()` call.

R/chat_app.R
gadenbuie: This isn't tool related, in fact ellmer should catch all author-caused errors in tool calls. This is a more general error handler for general API-level errors, e.g. putting `model = "claude-3-5-sonnet"` (not a valid model name).

So trying again may or may not work, it depends on the error. In this case, I'm just trying to emulate the sanitized errors shown in shiny apps generally (in local dev you'll see the actual error message).

R/chat_app.R
gadenbuie: Yeah mostly colors or styles. They show up in the `cli` errors that are common now. They're hard to turn off with options (and we probably wouldn't want to so we get the formatting in the console), so it's just easier to strip them with regex.

R/chat_app.R
gadenbuie: Yeah, the suggested code gives me

```
Caused by error in `as.promise.default()`:
! Don't know how to convert object of class coro_generator_instance into a promise
```

Actually, if everything works as expected you just need

```
stream <- client$stream_async(user_input)
chat_append(ui_id, stream)
```

inside the extended task, but you get into trouble when `$stream_async()` fails.

R/chat_app.R
```diff
@@ -0,0 +1,216 @@
...
+    ...
+  )
+}
+
+#' @describeIn chat_app A simple chat app module server.
+#' @export
+chat_mod_server <- function(id, client) {
+  check_ellmer_chat(client)
+
+  append_stream_task <- shiny::ExtendedTask$new(
```
jcheng5, line 140: I'm hazy on this but I vaguely recall not using `ExtendedTask` for this because I wanted tools to be able to do Shiny stuff. I'm not thinking clearly right now about whether having tools run under ExtendedTask instead of directly in the async observer, meaningfully impedes the tool's ability to do anything.

(I know with ExtendedTask you can't take a reactive dependency, but it's hard for me to imagine how that would be useful for a tool call.)

In Python we don't really have a choice, we must use ExtendedTask because async observers don't unblock other sessions.

gadenbuie: What I found in practice -- by creating a tool whose function adds Shiny UI and blocks evaluation while waiting for an input -- is that without `ExtendedTask` there are no reactive updates until the tool finishes evaluating.

The other thing that happens without `ExtendedTask` is that streaming to chat UI components is synchronous. If you take out the `ExtendedTask`, the app in the example (two side-by-side `claude`/`gpt-4o` chats) streams to each chat synchronously if you submit two chats in quick succession. With `ExtendedTask`, they both stream simultaneously.

R/chat_app.R
jcheng5: Oh, good point. That's unfortunate, can we work harder to catch those?

In the meantime, I believe `promise_resolve` catches the synchronous error. But it's also fine with me if you want to leave it as-is.

```suggestion
      promises::then(promises::promise_resolve(client$stream_async(user_input)), function(stream) {
        chat_append(ui_id, stream)
      })
```

jcheng5 (approved):

R/chat_app.R
gadenbuie: Perfect, thanks!

PR #45, opened 2025-05-13
feat(chat_append): Return stream results as promise value

gadenbuie:
Currently `chat_append()` returns a promise that resolves when the stream is complete, but it resolves to a value of `NULL` or an error if one occurs.

This PR updates `chat_append_stream_impl()` to collect the stream results and return them as the final value of the resolved promise.


R/chat.R
```diff
@@ -448,13 +448,19 @@ rlang::on_load(
       chunk = "start",
       session = session
     )
+
+    res <- list()
```
cpsievert: Seems worth using `fastmap::fastqueue()` instead?

R/chat.R
```diff
@@ -463,13 +469,20 @@ rlang::on_load(
...
+
     chat_append_message(
       id,
       list(role = role, content = ""),
       chunk = "end",
       operation = "append",
       session = session
     )
+
+    if (every(res, is.character)) {
```
cpsievert, line 495: Where is `every` coming from? purrr?

gadenbuie: https://github.com/posit-dev/shinychat/blob/main/R/import-standalone-purrr.R

R/chat.R
gadenbuie: Sure, done in de491ef1ad36115238e9810e5d610c52c6a2958d

Here's an example you can try to see how it works:

```r
library(shiny)
library(bslib)
library(ellmer)
pkgload::load_all()

client <- ellmer::chat_openai(model = "gpt-4.1-nano")

ui <- page_fluid(
  chat_ui("chat"),
)

  # chat_mod_server("chat", client)
    stream <- client$stream_async(input$chat_user_input)
    chat_append("chat", stream)$then(function(value) {
      message("Response: ", value)
    })
  })
}

shinyApp(ui, server)
```

NEWS.md
```diff
@@ -10,6 +10,8 @@
 
 * Added `chat_app()`, `chat_mod_ui()` and `chat_mod_server()`. `chat_app()` takes an `ellmer::Chat` client and launches a simple Shiny app interface with the chat. `chat_mod_ui()` and `chat_mod_server()` replicate the interface as a Shiny module, for easily adding a simple chat interface connected to a specific `ellmer::Chat` client. (#36)
 
+* The promise returned by `chat_append()` now resolves to the content streamed into the chat. (#49)
```
cpsievert, line 13: Would be good to surface this on the Rd page as well

PR #52, opened 2025-05-19
WIP: Tool calling UI

gadenbuie:


js/src/chat/chat-tools.ts
```diff
@@ -0,0 +1,290 @@
...
+
+  render() {
+    const bodyContent = html` ${this.#renderRequest()} ${this.#renderResult()} `
+
+    return this.renderCard(bodyContent)
+  }
+}
+
+const ICONS = {
+  tools: `<svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" fill="currentColor" class="bi bi-tools" viewBox="0 0 16 16">
```
gadenbuie: I used this icon first, but I like the wrench icon better; kept this for testing but we can remove before shipping.

js/src/markdown-stream/markdown-stream.ts
```diff
@@ -58,6 +58,19 @@ markdownRenderer.table = (header: string, body: string) => {
...
+const defaultMarkdownCodeRenderer = markdownRenderer.code
+
+markdownRenderer.code = function (
+  code: string,
+  infostring: string | undefined,
+  escaped: boolean,
+): string {
+  if (infostring === "{=html}") {
+    return code
+  }
```
gadenbuie, line 70: This lets us replicate pandoc's raw html block with markdown like

````
```{=html}
```
````

where we can then use this pseudo-raw-html block to wrap raw HTML and then avoid additional markdown processing inside the HTML (and avoid wrapping in a `<p>` element).

File: pkg-r/R/chat.R [hunk 3]
```diff
@@ -380,8 +380,16 @@ chat_append_message <- function(
     ui <- process_ui(content, session)
   }
 
+  msg_content <- ui[["html"]]
+  if (is_html) {
+    msg_content <- sprintf(
+      "\n\n````````{=html}\n%s\n````````\n\n",
```
gadenbuie, line 394: Pairs with the custom raw-html block in the markdown parser, but with extra backticks to make sure it isn't accidentally closed by internal content.

pkg-r/R/contents_shinychat.R
```diff
@@ -0,0 +1,207 @@
...
+  )
+}
+
+tool_result_display <- function(content) {
+  # Default fallback
+  res <- list(
+    value = content@value,
+    value_type = "code",
+    deps = NULL
+  )
```
cpsievert: If I'm following correctly, this effectively means we do `tag(value = value)` where `value` could be any R object? 

I'm not 100% positive, but I think htmltools will run `as.character()` on `value`, which isn't ideal for more complex values (e.g., data frames).

Maybe this should behave similar to how `renderPrint()` does? Where you effectively get whatever you'd see if you printed the object at the console (i.e., `capture.output(print(value))`)?

gadenbuie: Yeah good catch. There's a helper function in ellmer that turns this value into a string or json, I'll see if there's a way to access that representation. Otherwise we can easily inline that helper

gadenbuie: Fixed in b4aa4fa617396397e06205bd2720ca869a16d27d

js/src/chat/chat.ts
```diff
@@ -250,7 +254,9 @@ class ChatInput extends LightElement {
     if (this.valueIsEmpty) return
     if (this.disabled) return
 
-    window.Shiny.setInputValue!(this.id, this.value, { priority: "event" })
+    window.Shiny.setInputValue!(this.id, this.value, {
+      priority: "event",
+    })
```
cpsievert:
```suggestion
    window.Shiny.setInputValue!(this.id, this.value, { priority: "event" })
```

js/src/chat/chat-tools.ts
```diff
@@ -0,0 +1,302 @@
...
+        data-require-bs-caller="chat_ui()"
+        data-require-bs-version="5"
+      >
+        <button
+          class="card-header"
+          id="${headerId}"
+          @click="${this.#toggleCollapse}"
+          aria-expanded="${this.expanded}"
+          aria-controls="${contentId}"
+        >
```
cpsievert, line 143: This is nice -- maybe worth bringing to bslib as well (not now, but eventually)?

js/src/chat/chat-tools.ts
```diff
@@ -0,0 +1,302 @@
...
+          id="${contentId}"
+          role="region"
+          aria-labelledby="${headerId}"
+          ?inert="${!this.expanded}"
+        >
+          ${bodyContent}
+        </div>
+        <script data-bslib-card-init>
+          bslib.Card.initializeAllCards()
+        </script>
```
cpsievert: What are we gaining by mimicking a `card()`? Full-screen/resize?

js/src/chat/chat-tools.scss
```diff
@@ -0,0 +1,152 @@
...
+
+    &[aria-expanded="false"] {
+      border-bottom-width: 0;
+    }
+
+    // .hstack
+    display: flex;
+    flex-direction: row;
+    align-items: center;
+    align-self: stretch;
```
cpsievert: Personally I like when display/position/size stuff comes first

js/src/chat/chat-tools.ts
gadenbuie: Dropped this script in d24fc48d4e729a7d388700ba4de9edd20b25b98a

It'd be nice to have full screen cards, but it's not worth the effort currently

js/src/chat/chat-tools.scss
gadenbuie: Moved in 45de77c0c4d951e9621c62cf330c519bcb380244

js/src/chat/chat-tools.ts
```diff
@@ -0,0 +1,299 @@
...
+      <div class="tool-icon ${this.classStatus}">${unsafeHTML(this.icon)}</div>
+      <div class="tool-title ${this.classStatus}">${this.formatTitle()}</div>
+      <div class="tool-spacer"></div>
+      ${this.intent ? html`<div class="tool-intent">${this.intent}</div>` : ""}
+    `
+
+    return html`
+      <div
+        class="shiny-tool-card card bslib-card bslib-mb-spacing html-fill-item html-fill-container m-0"
+        data-bslib-card-init
```
cpsievert:
```suggestion
```

js/src/chat/chat-tools.ts
```diff
@@ -0,0 +1,299 @@
...
+        data-require-bs-caller="chat_ui()"
+        data-require-bs-version="5"
```
cpsievert: The intention behind these attributes is to inform/warn about a `fluidPage(card())` situation.  Feels weird if we were to do that since we're in charge of the markup here?

```suggestion
```

js/src/chat/chat-tools.ts
```diff
@@ -0,0 +1,299 @@
...
+    const headerContent = html`
+      <div class="tool-icon ${this.classStatus}">${unsafeHTML(this.icon)}</div>
+      <div class="tool-title ${this.classStatus}">${this.formatTitle()}</div>
+      <div class="tool-spacer"></div>
+      ${this.intent ? html`<div class="tool-intent">${this.intent}</div>` : ""}
+    `
+
+    return html`
+      <div
+        class="shiny-tool-card card bslib-card bslib-mb-spacing html-fill-item html-fill-container m-0"
```
cpsievert: This is maybe a little too aggressive in removing classes, but seems like we don't need filling layout, and also don't need bslib's complex margin/spacing rules?

```suggestion
        class="shiny-tool-card card m-0"
```

js/src/chat/chat-tools.ts
gadenbuie: Something to consider adding when we turn `bslib::card()` into a custom component 😉

js/src/chat/chat-tools.ts
gadenbuie: Yeah, I was thinking we do want filling -- if you return only HTML and turn off tool request display, you could have your tool return a map or plot that fills the card. I have a demo planned to show that.

js/src/chat/chat-tools.ts
```diff
@@ -0,0 +1,299 @@
...
+          id="${headerId}"
+          @click="${this.#toggleCollapse}"
+          aria-expanded="${this.expanded}"
+          aria-controls="${contentId}"
+        >
+          ${headerContent}
+          <div class="collapse-indicator">${unsafeHTML(ICONS.plus)}</div>
+        </button>
+        <div
+          class="card-body bslib-gap-spacing html-fill-item html-fill-container${this
```
cpsievert, line 148: Assuming we don't actually need filling layout or complex gap spacing

```suggestion
          class="card-body ${this
```

js/src/chat/chat-tools.ts
gadenbuie: Yeah, and it's already handled by `tag_require()` in `chat_ui()`. I was thinking we'd want to stick as close to `bslib::card()` as reasonable; but in the end we'll be better off with a custom element in bslib that we'd use here as needed.

js/src/chat/chat-tools.ts
gadenbuie: And just generally the further we get away from `bslib::card()` markup, the more space we leave for updates to `bslib::card()` to miss the nuance of differences in this component. Sounds like a bad time waiting to happen.

cpsievert: Ahh OK, I had missed the `max-height` setting on `.shiny-tool-card`

js/src/chat/chat-tools.ts
```diff
@@ -0,0 +1,296 @@
...
+      return html``
+    }
+
+    const bodyContent = html`
+      <shiny-markdown-stream
+        content="${markdownCodeBlock(this.arguments, "json")}"
+        content-type="markdown"
+        ?streaming=${false}
+      ></shiny-markdown-stream>
+    `
```
cpsievert, line 226: Could simplify this a bit by importing from `markdown-stream.ts` and do something like?

```js
import { parse, markdownRenderer } from "../markdown-stream"

const args = markdownCodeBlock(this.arguments, "json")
const bodyContent = parse(content, { renderer: markdownRenderer })
````

File: js/src/chat/chat-tools.ts [hunk 14]
```diff
@@ -0,0 +1,296 @@
...
+      if (this.valueType !== "markdown") {
+        // If value_type is "code", we format it as a markdown code block
+        result = markdownCodeBlock(this.value, "text")
+      }
+
+      result = html`<shiny-markdown-stream
+        content=${result || this.value}
+        content-type="markdown"
+        ?streaming=${false}
+      ></shiny-markdown-stream>`
```
cpsievert, line 338: Same point as https://github.com/posit-dev/shinychat/pull/52/files#r2246381129

js/src/chat/chat-tools.ts
```diff
@@ -0,0 +1,296 @@
...
+ */
+export class ShinyToolResult extends ShinyToolCard {
+  @property({ type: String, attribute: "request-call" })
+  requestCall?: string
+
+  @property({ type: String })
+  status!: string
+
+  @property({ type: Boolean, attribute: "show-request" })
+  showRequest = true
```
cpsievert: It'd be good to document that what we mean here is showing the request _inside_ the result container (i.e., the request container element should always be hidden).

js/src/chat/chat-tools.ts
```diff
@@ -0,0 +1,296 @@
...
+  #renderRequest() {
+    if (!this.showRequest || !this.requestCall) {
+      return ""
+    }
+
+    const request = html`<shiny-markdown-stream
+      content="${markdownCodeBlock(this.requestCall, "")}"
+      content-type="markdown"
+      ?streaming=${false}
+    ></shiny-markdown-stream>`
```
cpsievert, line 367: Same point as https://github.com/posit-dev/shinychat/pull/52/files#r2246381129

js/src/chat/chat-tools.ts
gadenbuie: Would that bring the code copy buttons?

cpsievert: Ahh, good point, no it wouldn't 😞

js/src/chat/chat-tools.ts
cpsievert: It does seem worth dropping `bslib-mb-spacing` and `bslib-gap-spacing` though since we're in control of the container of the card?

js/src/markdown-stream/markdown-stream.ts
cpsievert: How did you even find this 😄 ? If there's any halfway decent documentation of `infostring`, it seems useful to have a link to it in the source (I hate navigating marked.js docs)

pkg-r/R/chat.R
cpsievert: Let's leave a comment in the source about this?

pkg-r/R/contents_shinychat.R
```diff
@@ -1,14 +1,261 @@
...
+  props <- list(
+    "request-id" = content@id,
+    name = content@name,
+    arguments = jsonlite::toJSON(content@arguments, auto_unbox = TRUE)
+  )
+
+  # Add optional title if present in tool annotations
+  tool <- content@tool
+  if (!is.null(tool) && !is.null(tool@annotations$title)) {
+    props$title <- content@tool@annotations$title
```
cpsievert:
```suggestion
    props$title <- tool@annotations$title
```

pkg-r/R/contents_shinychat.R
```diff
@@ -1,14 +1,261 @@
...
+  }
+
+  # Add optional intent if present in request arguments
+  if (!is.null(content@arguments$.intent)) {
+    props$intent <- content@arguments$.intent
+  }
```
cpsievert:
```suggestion
  props$intent <- content@arguments[[".intent"]]
```

pkg-r/R/contents_shinychat.R
```diff
@@ -1,14 +1,261 @@
...
+
+  # Return structured tag
+  htmltools::tag("shiny-tool-request", props)
+}
+
+S7::method(contents_shinychat, ellmer::ContentToolResult) <- function(content) {
+  # Prepare base props
+  props <- list(
+    request_id = content@request@id,
```
cpsievert: I think technically we need to handle the `is.null(content@request)` case?

pkg-r/R/contents_shinychat.R
```diff
@@ -1,14 +1,261 @@
...
+      ),
+      auto_unbox = TRUE,
+      pretty = 2
+    )
+  }
+
+  # Add optional intent if present
+  if (!is.null(content@request@arguments$.intent)) {
+    props$intent <- content@request@arguments$.intent
+  }
```
cpsievert:
```suggestion
  props$intent <- content@request@arguments[[".intent"]]
```

pkg-r/R/contents_shinychat.R
```diff
@@ -1,14 +1,261 @@
...
+  is_html <-
+    inherits(display, c("html", "shiny.tag", "shiny.tag.list")) ||
+    (is.list(display) && !is.null(display$html))
+
+  if (is_html) {
+    html <- list(
+      value = format(display),
+      value_type = "html",
+      deps = htmltools::findDependencies(display)
+    )
```
cpsievert, line 140: I think you could get away with just returning `value` and `value_type` from this function?

```suggestion
    html <- list(
      value = display,
      value_type = "html",
    )
```

pkg-r/R/contents_shinychat.R
```diff
@@ -1,14 +1,261 @@
...
+    return(html)
```
cpsievert: I think you could get away with just returning `value` and `value_type` from this function?

```suggestion
    return(list(value = display, value_type = "html"))
```

cpsievert: Going even further, seems the logic could be cleaner if you had one function to just get the `value_type`, and another function that uses it, but returns the `value`.

pkg-r/R/client_state.R
```diff
@@ -88,7 +88,35 @@ method(client_set_ui, S7::new_S3_class(c("Chat", "R6"))) <-
 
     msgs <- contents_shinychat(client)
     lapply(msgs, function(x) {
```
cpsievert: Let's comment here about how `x` roughly translates to a `Turn`, which is a list of content objects, and each content object may have a list-like representation (like a tag).

pkg-r/tests/testthat/test-chat.R
```diff
@@ -1,30 +1,32 @@
...
         list(content = "Assistant", role = "assistant"),
         list(content = "User", role = "user")
       )
-    )
+    )))
   })
 
   expect_snapshot({
-    chat_ui(
+    cat(format(chat_ui(
```
cpsievert: What's idea behind adding this (it gives the same result)?

gadenbuie: Yeah I was getting weird errors that `expect_snapshot()` doesn't work with class `NULL`, but I think it was an odd interaction locally.

pkg-r/R/client_state.R
gadenbuie: 3426408e6b189204e6f840b2d5bb7500ac9ee755

pkg-r/R/contents_shinychat.R
gadenbuie: I mean, this is quite unlikely, but could occur if the tool result is constructed by hand.

I updated `contents_shinychat()` to throw if a request isn't attached to the result. ee514d41f1351d9d2234b62d4195e1174478d63e

js/src/chat/chat-tools.ts
gadenbuie: No, unless you can think of a way this would cause problems. My take is that we should follow `bslib::card()` as close as reasonably possible unless we have evidence that including a class is going to cause issues.

pkg-r/R/contents_shinychat.R
gadenbuie: There's some overlapping logic between `value_type` and `value` that doesn't make it worth separating into two functions. But I was able to clean this up quite a bit with a second pass in f3c28fa85b2e226c964793604e6d6b3b54809c9f

pkg-r/R/client_state.R
```diff
@@ -87,8 +87,36 @@ method(client_set_ui, S7::new_S3_class(c("Chat", "R6"))) <-
...
+            chunk = TRUE,
+          )
+        }
+
+        chat_append_message(
+          id,
+          msg = list(role = msg_turn$role, content = ""),
+          operation = "append",
+          chunk = "end",
+        )
```
cpsievert:
```suggestion
        stream <- coro::async_generator(function() {
          for (x in msg_turn$content) coro::yield(x)
        })
        chat_append(id, stream, msg_turn$role)
})
```

js/src/chat/chat-tools.ts
cpsievert: Yea, I think that's generally a good default mindset here (especially for something like `.card-body`).

I was mainly thinking about `bslib-mb-spacing` in particular. That class aims to solve a problem we don't really have here and it would be bit simpler reason about the spacing rules if we can think of spacing solely in terms of the custom elements.

pkg-r/R/contents_shinychat.R
```diff
@@ -1,14 +1,216 @@
...
+  if (is.character(display)) {
+    return(list(value = tool_string(content), value_type = "markdown"))
+  }
+
+  if (is.list(display)) {
+    has_type <- intersect(c("html", "markdown", "text"), names(display))
+    if (length(has_type) > 0) {
+      value_type <- has_type[1]
+      return(list(value = display[[value_type]], value_type = value_type))
+    }
```
cpsievert, line 320: Do we want to warn or anything if `length(has_type) == 0`?

pkg-r/R/contents_shinychat.R
cpsievert: Nice, thanks. 👍

js/src/markdown-stream/markdown-stream.ts
```diff
@@ -167,7 +180,42 @@ class MarkdownElement extends LightElement {
   }
 
   #appendStreamingDot(): void {
-    this.lastElementChild?.appendChild(SVG_DOT)
+    this.#removeStreamingDot()
+
+    const findInnermost = (element: Element): Element => {
```
cpsievert: I would love a high-level summary of what is function is trying to achieve. And if I follow correctly, it's to find the last `display: inline` element (so that we can append the dot inline next to that element)?

js/src/markdown-stream/markdown-stream.ts
```diff
@@ -167,7 +180,42 @@ class MarkdownElement extends LightElement {
...
+        "p",
+        "h1",
+        "h2",
+        "h3",
+        "h4",
+        "h5",
+        "h6",
+        "li",
+        "code",
+      ]
```
cpsievert: Is it feasible/better to use `window.getComputedStyle(x, "display")` to determine what is block v inline?

js/src/markdown-stream/markdown-stream.ts
cpsievert: Also I think the name could be improved. Something like `findLastInlineElement()`?

js/src/markdown-stream/markdown-stream.ts
```diff
@@ -167,7 +180,42 @@ class MarkdownElement extends LightElement {
   }
 
   #appendStreamingDot(): void {
-    this.lastElementChild?.appendChild(SVG_DOT)
+    this.#removeStreamingDot()
+
+    const findInnermost = (element: Element): Element => {
+      // Get the last child that's an element (not text/comment)
+      const lastChild = Array.from(element.children).pop()
```
cpsievert: Since this is recursive (on every chunk), let's make sure it's efficient

```suggestion
      const lastChild = element.lastElementChild
```

js/src/chat/chat-tools.scss
```diff
@@ -0,0 +1,152 @@
...
+  }
+
+  .card-header,
+  .card-footer {
+    font-size: inherit;
+    font-weight: 400;
+    word-break: break-word;
+  }
+
+  >.card-header {
```
cpsievert, line 41: Do you prefer no space? I prefer space. I can go either way, but there is some inconsistency with space in the selectors

js/src/chat/chat-tools.scss
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    .collapse-indicator {
+      display: grid;
+      place-items: center;
+      width: 1em;
+      height: 1em;
+      opacity: 0.66;
+
+      &, > .horizontal {
```
cpsievert:
```suggestion
      &, >.horizontal {
```

pkg-r/R/client_state.R
gadenbuie: oh that's nice! thanks!

gadenbuie: Actually I think it needs to be a sync generator, but still this is nice

js/src/markdown-stream/markdown-stream.ts
gadenbuie: Claude helped me figure this out, it gave me a different solution but with enough of the code that I figured out how to make it look more like syntax we're used to

js/src/markdown-stream/markdown-stream.ts
gadenbuie: The goal is to insert the streaming dot inside the last tag where we're currently streaming, which is sometimes contained inside other tags. It's not as direct as "find inline elements", because we don't want _any_ inline element, we want to figure out which element is being streamed into.

I need to tweak this a little bit to make sure we don't recurse into elements whose last child is a textNode (because that's where the streaming is happening).

js/src/markdown-stream/markdown-stream.ts
gadenbuie: Not really; that's too general of a solution. We can cut a lot of corners here because we know we're getting markdown. (And if it's not markdown, the behavior we get is still okay.)

pkg-r/R/contents_shinychat.R
gadenbuie: idk, that doesn't feel necessary to me. Maybe if devmode is turned on; otherwise I think it'd be too much noise shown to the wrong person. Even then, I'm not sure; it feels a little disconnected from where the problem would be and if you're developing tools with this feature in mind you'll know quickly if it works or not because you will or will not see the content in the browser during testing.

So on the whole, I don't think it's worth it right now

js/src/chat/chat-tools.ts
gadenbuie: Yeah, we can drop that class

js/src/markdown-stream/markdown-stream.ts
gadenbuie: Refactored and renamed in b43d1c1ee646cef1d70699c41d9e27ee28f80b62

js/src/markdown-stream/markdown-stream.ts
gadenbuie: Made efficient in b862a2c836c52756174b65a9517d49c4ed10d5a3

pkg-r/R/contents_shinychat.R
```diff
@@ -1,14 +1,216 @@
...
+tool_result_display <- function(content) {
+  display <- content@extra$display
+
+  if (tool_errored(content) || is.null(display)) {
+    return(list(value = tool_string(content), value_type = "code"))
+  }
+
+  if (inherits(display, c("html", "shiny.tag", "shiny.tag.list"))) {
+    return(list(value = display, value_type = "html"))
+  }
```
cpsievert: It recently occurred to me that a reasonable request might be to see the request, value (sent to the LLM),  _and_ also display something custom below it? Not saying we have to support that now, but at least make sure we're OK with API/tradeoffs we're taking with that in mind?

gadenbuie: I'm okay with the trade offs we're introducing. The goal is to provide _alternate_ displays in different settings, so you can show the LLM a jsonified table but the user can see a rich HTML table.

If you want your tool to show both the value and a rich display, you can still do that manually.

If anything, we might want to put the use of the display property behind an option. Even then, though, that alternative display feature is something that's only available to people who are writing tools in R (and not, e.g., to tools from MCPs).

cpsievert: > show both the value and a rich display, you can still do that manually.

In order to do that faithfully (i.e., always reflect what the LLM sees), you'd have to replicate `ellmer:::tool_string()`, right?

FWIW, this does feel like another situation where we'll regret not defaulting to a behavior where the end user can verify what information the LLM is receiving.

pkg-r/R/contents_shinychat.R
```diff
@@ -1,14 +1,231 @@
...
+
+tool_result_display <- function(content) {
+  display <- content@extra$display
+
+  if (tool_errored(content) || is.null(display)) {
+    return(list(value = tool_string(content), value_type = "code"))
+  }
+
+  if (inherits(display, "htmlwidget") || inherits(display$html, "htmlwidget")) {
+    widget <- tagList(display$html %||% display)
```
cpsievert: `display$html` isn't what you want when `display` is an `htmlwidget`.

More generally, it doesn't feel right to be in the business of detecting html content. It would make more sense to me if we handled the `is.character(display)` and `display$html/markdown/text` cases first, then have the "base" case handle `display` in the usual "tag child" sense. Related to that point, with this current approach, if `display` has dependencies, seems as though we'd be dropping them?

js/src/markdown-stream/markdown-stream.ts
```diff
@@ -167,7 +180,68 @@ class MarkdownElement extends LightElement {
...
+    ])
+
+    /**
+     * Find the innermost element where streaming is happening, i.e. where the
+     * streaming is appending new content.
+     */
+    const findInnermostStreamingElement = (element: Element): Element => {
+      let current = element
+
+      while (true) {
```
cpsievert: Would be great to have some limit to how far we traverse. Something like:

```js
const findInnermostStreamingElement = (element: Element): Element => {
  let current = element
  let depth = 0
  const MAX_DEPTH = 20 // Reasonable limit for markdown nesting

  while (depth < MAX_DEPTH) {
    // ... existing logic ...
    depth++
  }
  
  return current // Fallback if max depth reached
}
```

pkg-r/R/contents_shinychat.R
gadenbuie: Good point, it does make a lot of sense to avoid trying to maintain a list of classes that are "html". I had fixed the missing dependencies issue earlier this morning; just pushed.

pkg-r/R/contents_shinychat.R
gadenbuie: > FWIW, this does feel like another situation where we'll regret not defaulting to a behavior where the end user can verify what information the LLM is receiving.

The default behavior **is** to show what the LLM saw. That's what, idk roughly, 80-99% of tool calls will look like. What we're enabling is much richer tool result displays, which allows app authors to write richer chat interfaces. 

I think we need both: 

1. a default display that shows users what's going on
2. a customizable interface for rich tool integrations with the app

and this PR gives us both. The problem is that it's hard for us to ensure _both_ objectives are satisfied at the same time; but I think at some point we need to trust app authors. 

I also think that by the time you have a situation where the custom tool result interface creates risk for the user, the risk to an actual end user is minimal; most of the risk is born by the author/host.

That said, I can get behind an R option/envvar that can be used to control tool result displays; I'll add that shortly.

gadenbuie: I added `shinychat.tool_display = c("none", "basic", "rich")` with an envvar `SHINYCHAT_TOOL_DISPLAY`. You can use `"none"` to turn off tool display entirely; `"basic"` to turn use basic tool display without any additional formatting; or `"rich"` (default) for the full display

pkg-r/R/contents_shinychat.R
gadenbuie: Actually, I'm going a step further and removing the option of including HTML objects or character strings as `@extra$display`. Instead, you'll need to use a list with one of `html`, `markdown` or `text`.

js/src/markdown-stream/markdown-stream.ts
gadenbuie: Sure. I think even shallower than 20 levels is fine; I used 5 in 36b7891cc20ba9898b0aa0cea184518671df4649

pkg-r/R/contents_shinychat.R
```diff
@@ -1,14 +1,296 @@
...
+
+  tool <- content@tool
+  if (!is.null(tool)) {
+    props$title <- tool@annotations$title
+  }
+
+  tagList(
+    htmltools::tag("shiny-tool-request", props),
+    chat_deps()
+  )
```
cpsievert:
```suggestion
  tool <- content@tool
  
  htmltools::tag(
    "shiny-tool-request", 
    list(
      "request-id" = content@id,
      name = content@name,
      arguments = jsonlite::toJSON(content@arguments, auto_unbox = TRUE),
      intent = content@arguments$.intent,
      title = if (!is.null(tool)) tool@annotations$title,
      chat_deps()
    )
  )
```

pkg-r/R/contents_shinychat.R
```diff
@@ -1,14 +1,296 @@
...
+
+  if (is.null(display) || opt_shinychat_tool_display() == "basic") {
+    return(list())
+  }
+
+  invalid_display_fmt <- "Invalid {.code @extra$display} format for {.code ContentToolResult} from {.fn {content@request@name}} (call id: {content@request@id})."
+
+  if (
+    inherits(display, c("html", "shiny.tag", "shiny.tag.list", "htmlwidgets"))
+  ) {
```
cpsievert, line 280: FWIW, if `display` is a character string or a bare list, it currently gets silently dropped

pkg-r/R/contents_shinychat.R
```diff
@@ -1,14 +1,296 @@
...
+  names(props) <- gsub("_", "-", names(props))
+
+  # Get dependencies for the value because we store its HTML as an attribute
+  # where `processDeps()` won't find it.
+  deps <- htmltools::findDependencies(props$value)
+
+  htmltools::tagList(
+    htmltools::tag("shiny-tool-result", props),
+    deps,
+    icon_deps,
```
cpsievert: This approach would lead to a reduced number of objects in scope

```suggestion
    htmltools::findDependencies(props$value),
    htmltools::findDependencies(props$icon),
```

pkg-r/R/contents_shinychat.R
```diff
@@ -1,14 +1,296 @@
...
+    return(NULL)
+  }
+
+  if (is.null(content@request)) {
+    cli::cli_abort(
+      "`ContentToolResult` objects must have an associated `@request` property."
+    )
+  }
+
+  content@extra$display <- ensure_content_display(content)
```
cpsievert: I'd find this method generally easier to follow with less `@`, `$`, and mutation.

```suggestion
  display <- content@extra$display %||% list()
  if (opt_shinychat_tool_display() == "basic") {
    display <- list()
  }
  # Doesn't return anything, just throws warnings
  check_content_display(display)
```

This would mean `display` would have to be passed to `tool_result_display()` separately, which is a bit awkward but feels worth it?

pkg-r/R/contents_shinychat.R
```diff
@@ -1,14 +1,296 @@
...
+    expanded = if (isTRUE(content@extra$display$open)) NA
+  )
+
+  icon_deps <- NULL
+  tool <- content@request@tool
+
+  # Let tool results override global tool properties
+  if (!is.null(content@extra$display$title)) {
+    props$title <- content@extra$display$title
+  }
```
cpsievert: Seems this `if` isn't necessary and `title = display$title` could just get inlined where `props` gets initially constructed?

pkg-r/R/contents_shinychat.R
```diff
@@ -1,14 +1,296 @@
...
+  if (!is.null(content@extra$display$icon)) {
+    props$icon <- content@extra$display$icon
+    icon_deps <- htmltools::findDependencies(props$icon)
+  }
```
cpsievert: Similar note to https://github.com/posit-dev/shinychat/pull/52/files#r2252686730

pkg-r/R/contents_shinychat.R
gadenbuie: I went with something similar in intent (c2e51389cbb1144a2dffa32656a899a599c9073f): use `get_tool_result_display()` to pull out `display` (which guarantees a list so we can avoid extra type checks later), and then pass `display` to `tool_result_display(content, display)`.

It saves a bunch of messiness to have the function return a `list()` when we detect incorrect structure, and at the same time the checks aren't complete enough to merit the name `check_content_display()`.

pkg-r/R/contents_shinychat.R
gadenbuie: No, the check in the lines just after this catch the case where `display` is a string. And lists aren't dropped.

``` r
library(ellmer)
pkgload::load_all()
#> ℹ Loading shinychat

the_tool <- tool(
  \() "output",
  name = "tool_call",
  description = "A tool that returns output"
)

request <- ContentToolRequest(
  id = "abc123",
  name = "tool_call",
  tool = the_tool
)

ContentToolResult(
  value = "output",
  extra = list(display = "bad!"),
  request = request
) |>
  contents_shinychat() |>
  format()
#> Warning: Invalid `@extra$display` format for `ContentToolResult` from `tool_call()`
#> (call id: abc123).
#> ✖ Expected a list with fields `html`, `markdown`, `text`, `show_request`,
#>   `open`, `title`, or `icon`, not a string.
#> [1] "<shiny-tool-result request-id=\"abc123\" request-call=\"tool_call()\" name=\"tool_call\" status=\"success\" show-request value=\"output\" value-type=\"code\"></shiny-tool-result>"

ContentToolResult(
  value = "output",
  extra = list(display = list(markdown = "this works fine")),
  request = request
) |>
  contents_shinychat() |>
  format()
#> [1] "<shiny-tool-result request-id=\"abc123\" request-call=\"tool_call()\" name=\"tool_call\" status=\"success\" show-request value=\"this works fine\" value-type=\"markdown\"></shiny-tool-result>"
```

pkg-r/R/contents_shinychat.R
```diff
@@ -1,14 +1,278 @@
...
+    chat_deps()
+  )
+  names(props) <- gsub("_", "-", names(props))
+
+  htmltools::tag("shiny-tool-result", props)
+}
+
+get_tool_result_display <- function(content) {
+  display <- content@extra$display
+  request <- content@request
```
cpsievert, line 270:
```suggestion
```

pkg-r/R/contents_shinychat.R
cpsievert: Ahh right, my bad.

pkg-r/tests/testthat/apps/tool-map/app.R
```diff
@@ -0,0 +1,72 @@
...
+)
+
+ui <- function(req) {
+  page_fillable(
+    chat_mod_ui("chat")
+  )
+}
+
+server <- function(input, output, session) {
+  client <<- ellmer::chat(
```
cpsievert:
```suggestion
  client <- ellmer::chat(
```

pkg-r/R/contents_shinychat.R
cpsievert: nice, thanks 👍

pkg-r/R/contents_shinychat.R
gadenbuie: Why? We use `request` in the warning messages

cpsievert: 🤦

pkg-r/vignettes/articles/tool-ui.Rmd
```diff
@@ -0,0 +1,513 @@
...
+
+The weather in Boston today is partly sunny with temperatures around 75-76°F
+during the afternoon. There is some humidity at about 64-66%. There is a
+possibility of rain starting around 4 PM, with increasing chances in the
+evening. Winds are coming from the northeast at about 8-10 mph.
+```
+
+If you're interested in learning more about how to use and create tools with `{ellmer}`, we recommend reading the [Tool/function calling article](https://ellmer.tidyverse.org/articles/tool-calling.html) on the `{ellmer}` website.
+
+If you're working interactively in the R console---and if you've set the `echo = "output"` option---ellmer shows you when tool calls are made and gives a preview of the tool result.
```
cpsievert, line 96:
```suggestion
If you're working interactively in the R console---or set the `echo = "output"` option---ellmer shows you when tool calls are made and gives a preview of the tool result.
```

pkg-r/vignettes/articles/tool-ui.Rmd
```diff
@@ -0,0 +1,513 @@
...
+chat$chat("What's the weather in Boston like today?")
+```
+```
+◯ [tool call] get_weather_forecast(lat = 42.3601, lon = -71.0589)
+● #> [{"time":"2025-08-05 11:00:00 EDT","temp":76,"dewpoint":17.7778,"hum…
+
+The weather in Boston today is partly sunny with temperatures around 75-76°F
+during the afternoon. There is some humidity at about 64-66%. There is a
+possibility of rain starting around 4 PM, with increasing chances in the
+evening. Winds are coming from the northeast at about 8-10 mph.
```
cpsievert, line 91: It would be nice to follow this up with a `$stream_async(stream = "content")` + shinychat example.

(The way this reads, one might think `$stream_async()` "just works" for displaying tool calls, but you need to know to opt-in, right?)

pkg-r/vignettes/articles/tool-ui.Rmd
gadenbuie: It's an _and_, `echo` only applies when the chat output is being printed to the console.

pkg-r/vignettes/articles/tool-ui.Rmd
gadenbuie: Agreed, I do think it's worth mentioning that this works "out of the box" in `chat_app()` and `chat_mod_server()`, which would also be a natural place to mention `$stream_async(stream = "content")` when using this with `chat_ui()`

pkg-r/tests/testthat/apps/tool-map/app.R
```diff
@@ -0,0 +1,72 @@
+library(shiny)
+library(bslib)
+library(ellmer)
+library(leaflet)
+pkgload::load_all()
```
cpsievert:
```suggestion
```

PR #53, opened 2025-05-19
Accept `initial_stream` in `chat_mod_server()`

t-kalinowski:
This is a minor patch that allows a chat server to start with an initial stream already running.
The goal is to let users do something like:

```r
quartohelp::ask_quarto_docs("How can I make a two column layout?")
```

and have the shiny gadget launch with the initial response already streaming.


t-kalinowski:
Hmm, this might be better as a custom-built app; I’m not sure this is useful outside of a shiny gadget.

PR #56, opened 2025-05-22
Reorganize `shinychat` into a monorepo

gadenbuie:
Collecting the web assets, R & Python packages, and docs into a single repo


cpsievert (approved):

.github/workflows/py-release.yaml
```diff
@@ -0,0 +1,54 @@
...
+  UV_VERSION: "0.7.x"
+  PYTHON_VERSION: 3.13
+
+jobs:
+  py-release:
+    name: Build and release Python package
+    runs-on: ubuntu-latest
+
+    # Python release tags start with "py/v*"
+    if: startsWith(github.ref, 'refs/tags/py/v')
```
schloerke, line 17: Nice

.vscode/settings.json
```diff
@@ -0,0 +1,40 @@
...
+    "editor.codeActionsOnSave": {
+        "source.fixAll.eslint": "always"
+    },
+    "eslint.validate": [
+        "javascript",
+        "typescript",
+        "jsx",
+        "tsx"
+    ],
+    "quarto.path": "~/.local/share/qvm/versions/v1.7.31/bin/quarto",
```
schloerke, line 31: This feels off

.vscode/extensions.json
```diff
@@ -0,0 +1,6 @@
+{
+    "recommendations": [
+        "Posit.air-vscode",
+        "dbaeumer.vscode-eslint"
```
schloerke, line 4:
```suggestion
        "dbaeumer.vscode-eslint",
        "charliermarsh.ruff",
        "esbenp.prettier-vscode"
```

pyproject.toml
```diff
@@ -0,0 +1,159 @@
+[project]
```
schloerke, line 1: Notes for me...

By having this in root, it avoids users having to require a subdirectory for install: https://pip.pypa.io/en/stable/topics/vcs-support/#url-fragments

schloerke (commented):
Sorry for the dealyed comments...

.vscode/extensions.json
gadenbuie: Thanks! 8ddd121cbad90fef45f0dac760d28a9f26e32f80

pyproject.toml
schloerke: (Resolving)

.vscode/settings.json
gadenbuie: Yeah this is tricky because `qvm` is the best thing around but also isn't being developed or maintained.

The idea here is that the Quarto version used in the project is controlled with the `QUARTO_VERSION` string in the `Makefile`. To update the Quarto version, you change that string and call `make install-quarto`, which uses `qvm` to install the desired version of Quarto _and_ updates the GitHub Actions files AND the VS Code settings so that everyone uses the same version of Quarto.

Unfortunately, `qvm` doesn't give you a way to control where the `quarto` binary is installed, or we'd use a local folder. I also don't really want to go write code to download and install quarto since that's basically covered by `qvm`, so we're kind of stuck here.

It also means that in CI, we install `quarto` but can't use `make docs` and have to rely on `quarto render ...`, which is annoying but not _the worst thing ever_.

PR #71, opened 2025-06-28
Make chat_app refresh its prepopulated UI messages on reload

jcheng5:
Before this commit:

1. Call `chat_app(client)` (or `ellmer::live_browser(client)`, which passes through to `chat_app`)
2. Chat with the bot a little bit
3. Hit Reload in the browser window

The discussion you have disappears, but the state is still present in the client--the UI and server state have gotten out of sync.

If you Ctrl+C out of the app, and then call `chat_app(client)` again, the existing messages are now displayed. But if you talk more, and then hit Reload, you'll be reverted back to the UI state that was correct at the time `chat_app(client)` was most recently called.

This is fixed by calling `chat_ui` at page load time, not at `chat_app()` invocation time.

(Note that it's unsafe to use a `chat_app()` from multiple browser windows simultaneously, since every chat will mutate the same client object. Might be worth putting in a tiny bit of code that closes existing sessions when a new session arrives.)

### DIFF
--- a/pkg-r/NEWS.md
+++ b/pkg-r/NEWS.md
@@ -4,6 +4,8 @@
 
 * Added `chat_enable_bookmarking()` which adds Shiny bookmarking hooks to save and restore the `{ellmer}` chat client. (#28)
 
+* `chat_app()` now correctly restores the chat client state when refreshing the app, e.g. by reloading the page. (#71)
+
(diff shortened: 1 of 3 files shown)


gadenbuie (approved):
Thanks!

PR #79, opened 2025-07-18
Update `chat_enable_bookmarking()` to work with latest `{ellmer}`

cpsievert:
If you run this app and submit input, you currently get an error -- I believe because the function signature of these ellmer functions changed.

cc @schloerke 

```r
library(shiny)
library(bslib)
library(shinychat)

ui <- function(request) {
  page_fillable(chat_ui("chat"))
}

  chat_client <- ellmer::chat_anthropic(
    system_prompt = "Important: Always respond in a limerick",
    echo = TRUE
  )

  chat_enable_bookmarking("chat", chat_client)

    chat_append("chat", stream)
  })
}

shinyApp(ui, server, enableBookmarking = "server")
```

```
Error in FUN: unused argument (chat = <environment>)
```

### DIFF
--- a/pkg-r/R/client_state.R
+++ b/pkg-r/R/client_state.R
@@ -23,8 +23,7 @@ method(client_get_state, S7::new_S3_class(c("Chat", "R6"))) <-
     # Instead, save only the `turns` information
     recorded_turns <- lapply(
       client$get_turns(),
-      ellmer::contents_record,
-      chat = client
+      ellmer::contents_record
     )
 
     if (is_url_bookmarkstore()) {
@@ -74,8 +73,7 @@ method(client_set_state, S7::new_S3_class(c("Chat", "R6"))) <-
 
     replayed_turns <- lapply(
       recorded_turns,
-      ellmer::contents_replay,
-      chat = client
+      ellmer::contents_replay
     )
 
     client$set_turns(replayed_turns)


pkg-r/R/client_state.R
```diff
@@ -74,8 +73,7 @@ method(client_set_state, S7::new_S3_class(c("Chat", "R6"))) <-
 
     replayed_turns <- lapply(
       recorded_turns,
-      ellmer::contents_replay,
-      chat = client
```
schloerke, line 78: We need to pass in tools = client$get_tools() 

(Or something similar.)

gadenbuie: @cpsievert @schloerke Did y'all end up resolving this? Or is this currently broken?

gadenbuie: Seems like maybe some of the changes in [`19a206c` (#80)](https://github.com/posit-dev/shinychat/pull/80/commits/19a206ceddffe148962e4d00e665672ae74a069f) should make it into `main`?

cpsievert: This PR #79 gets bookmarking working (without tool calling). Those changes in #80 get bookmarking working with tool calling.

gadenbuie: Do you remember if that changed worked for you as-is? I'm trying to re-apply the patch and the call to `chat_enable_bookmarking()` in `chat_mod_server()` leads to

```
Error in chat_enable_bookmarking("chat", client, session = session) : 
  Error: Shiny bookmarking is not enabled. Please enable bookmarking in your
  Shiny app either by calling `shiny::enableBookmarking("server")` or by
  setting the parameter in `shiny::shinyApp(enableBookmarking = "server")`
```

And if I add `shiny::enableBookmarking("server")` to the server function of `chat_app()` or `chat_mod_server()`, I'm getting

```
Error in session$onBookmarked(fun) : 
  onBookmarked() can't be used in a module.
```

cpsievert: Ohh hmm, yea. I don't think I had tested yet with the module approach. I think the way go about it properly would be to first check if bookmarking is enabled?

```r
if (!is.null(shiny::getShinyOption("bookmarkStore"))) {
   chat_enable_bookmarking("chat", client, session = session)
}
```

PR #82, opened 2025-07-30
feat: `chat_restore()` replaces `chat_enable_bookmarking()`

gadenbuie:
Replaces `chat_enable_bookmarking()` with `chat_restore()`, which now also restores the message history either from the bookmark or from the messages on `client` when passed in.

@schloerke and I paired through getting `chat_enable_bookmarking()` to work inside modules. Along the way, we realized that it'd be best for this function to also handle setting UI state from the message history attached to the chat `client`, in particular because we only want to perform that restore if we're not restoring from a bookmark.

This change meant that `chat_enable_bookmarking()` was not quite accurately named. Also, because `chat_enable_bookmarking()` still requires `enableBookmarking` to be turned on in the Shiny app, `chat_restore()` feels a bit more generally appropriate.

We also now turn on bookmarking in `chat_app()` and we also avoid the issues described in #71. 

Finally, `chat_mod_server()` gains the `bookmark_on_*` arguments of `chat_restore()` and does the restoration automatically. And we deprecate `client` in `chat_mod_ui()` because this is now handled entirely on the server.

Implementation note: this PR includes a `contents_shinychat()` placeholder that will be replaced with `contents_shinychat()` in #52.


pkg-r/R/chat_app.R
```diff
@@ -96,7 +101,7 @@ chat_app <- function(client, ...) {
     })
   }
 
-  shiny::shinyApp(ui, server, ...)
+  shiny::shinyApp(ui, server, ..., enableBookmarking = "url")
```
cpsievert:
```suggestion
  shiny::shinyApp(ui, server, ..., enableBookmarking = bookmark_store)
```

pkg-r/R/chat_app.R
```diff
@@ -73,20 +80,18 @@
 #'   app is suitable for interactive use by a single user; do not use
 #'   `chat_app()` in a multi-user Shiny app context.
 #' @export
-chat_app <- function(client, ...) {
+chat_app <- function(client, ..., bookmark_store = "url") {
   check_ellmer_chat(client)
 
-  ui <- function(req) {
```
cpsievert, line 79: I thought bookmarking requires the UI to be a function?

pkg-r/R/chat_restore.R
```diff
@@ -123,11 +117,14 @@ chat_enable_bookmarking <- function(
...
       }
 
+      cancel_set_ui$destroy()
       client_set_state(client, client_state)
 
       # Set the UI
-      chat_clear(id)
-      client_set_ui(client, id = id)
+      shiny::withReactiveDomain(session, {
+        chat_clear(id)
```
cpsievert, line 130: Will this also clear starting messages (i.e., `chat_ui(messages = list(...))`)?

pkg-r/R/chat_app.R
gadenbuie: Oops, thanks for catching that!

pkg-r/R/chat_restore.R
gadenbuie: Good point and yeah, that's true. I've updated the docs of `chat_mod_ui()` to clarify that `messages` only applies when the client doesn't have any history. And left a similar note in the `chat_restore()` docs.

PR #96, opened 2025-08-07
feat(pkg-py): Add new `message_content()` and `message_content_chunk()`

cpsievert:
Adds new `message_content()` and `message_content_chunk()` generic functions (via `singledispatch`). They aren't intended to be called directly by users, but instead, provide an opportunity to teach `Chat.append_message()`/`Chat.append_message_stream()` how to extract a `ChatMessage()` from different bespoke objects.

```python
from shiny import reactive
import shiny.express

from shinychat import message_content
from shinychat.types import ChatMessage

class MyObject:
    foo = "bar"

@message_content.register
def _(message: MyObject):
    return ChatMessage(content=message.foo, role="assistant")

chat = Chat("chat")
chat.ui()

@reactive.effect
async def _():
    await chat.append_message(MyObject())

```

These new functions are the `singledispatch` equivalent of the previous (internal) `normalize_message()` and `normalize_message_chunk()` helpers. Although not quite as flexible as the previous strategies pattern, this approach much simpler (implementation wise and also for those registering new methods) and aligns much better with `shinychat::contents_shinychat()` (#52).


pkg-py/CHANGELOG.md
```diff
@@ -7,10 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [UNRELEASED]
 
-* The chat input no longer submits incomplete text when the user has activated IME completions (e.g. while typing in Japanese or Chinese). (#85)
+### New features
+
+* Added new `message_content()` and `message_chunk_content()` generic (`singledispatch`) functions. These functions aren't intended to be called directly by users, but instead, provide an opportunity to teach `Chat.append_message()`/`Chat.append_message_stream()` to extract message contents from different types of objects. (#96)
```
gadenbuie: I feel bad suggesting it, but I like `message_content_chunk()` a little better. Both make sense, but I think there's more precedence for the `_chunk` variants to follow regular message methods, e.g.

* `_append_message()` -> `message_content()`
* `_append_message_chunk()` -> `message_content_chunk()`

PR #101, opened 2025-08-22
refactor: Consolidate chat messages into a single class with `data-role` attribute

gadenbuie:
This PR refactors `<shiny-chat-message>` and `<shiny-user-message>` into a single `<shiny-chat-message>` component. We keep `ShinyUserMessage` for backwards compatibility (because it's easy to support by extending `ShinyChatMessage`.

The goal is to consolidate the logic into a single class so that future updates to the chat message class can be handled in one place. It will also make it easier to add icon support for user avatars (at a future time, not in this PR).


js/src/chat/chat.scss
```diff
@@ -1,4 +1,4 @@
-@import "chat-tools.scss";
+@use "chat-tools.scss";
```
gadenbuie, line 1: We're using Dart Sass to compile this file; this avoids a warning because `@import` is going away in a future version.

js/src/chat/chat.ts
```diff
@@ -72,25 +76,41 @@ class ChatMessage extends LightElement {
...
 
     return html`
-      <div class="message-icon">${unsafeHTML(icon)}</div>
+      ${icon}
       <shiny-markdown-stream
         content=${this.content}
         content-type=${this.contentType}
         ?streaming=${this.streaming}
-        auto-scroll
+        ?auto-scroll=${this.role === "assistant"}
```
cpsievert, line 90: Does `role` need to be defined as a property/attribute?

cpsievert: Oh nvm

gadenbuie:
> We keep `ShinyUserMessage` for backwards compatibility (because it's easy to support by extending `ShinyChatMessage`).

@cpsievert Do you think we should do this? I think the only external way that people would interact with `shiny-user-message` is in their CSS and, well, they'll have to update that if they're using the tag name because we're not anymore.

We _could_ still use `<shiny-user-message>` as the tag, which was my original plan going in, but then I found the simplification nice and think internally it's better as a single tag.

So now that I'm talking this through, I think maybe we'd want to just go all in on `shiny-chat-message`.

cpsievert:
It would have implications for exporting to HTML in chatlas, so it'd be nice not to have to worry about updating that, but I don't have strong opinions either way.

PR #107, opened 2025-09-02
feat(pkg-py): Python tool displays v2

cpsievert:
This PR ports the new, more customizable, tool calling displays (already added to R in #52) over to Python.

In order to make use of it, you'll need the latest chatlas (if you don't have the latest, you'll get the legacy displays):

```shell
pip install -U chatlas
```

Also note that the manual testing apps added in #52 (under `pkg-r/tests/testthat/apps`) have been ported over (under `pkg-py/tests/playwright/tools`) and used for testing. It'd be great to automate some tests here, but @gadenbuie has requested that we wait post-conf (and focus instead on shipping other needed features)

### TODO

- [ ] ~Review, merge, and release a fix for bookmarking custom displays https://github.com/posit-dev/chatlas/pull/167~

### DIFF
--- a/pkg-py/CHANGELOG.md
+++ b/pkg-py/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### New features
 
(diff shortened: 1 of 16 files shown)


gadenbuie (approved):
This looks great! I didn't read it in great detail, but I read through the changes and everything is very nicely organized and clean. I really like the way that `ToolResultComponent` and `ToolResultDisplay` are set up, it composes nicely.

I'm okay with merging this; I'll try to update Python querychat soon to use these features, which will give me a chance to try them out. We can follow up then with any feedback.

PR #119, opened 2025-09-05
feat(pkg-py): Use "raw html" code blocks for HTML content

### LINKED ISSUE #118: [python] Tool UI: Wrap `shiny-tool-{request,result}` HTML in "raw html" blocks
In R, we wrap content that we identify as being pure HTML in "raw html" blocks

https://github.com/posit-dev/shinychat/blob/2706000c49410d73caa22c0d705347d59b69b329/pkg-r/R/chat.R#L403-L411

Without this, two things happen:

1. The HTML for tool requests and results are blended into text, e.g.

   ```
   Let me check:<shiny-tool-request></shiny-tool-request><shiny-tool-result></shiny-tool-result>I can see that...
   ```

2. `marked` puts the tool request and result elements (and any other raw html emitted) inside paragraphs tags, which we don't want.

With the raw html block we instead emit

``````````
Let me check:

``````{=html}
<shiny-tool-request></shiny-tool-request>
``````

``````{=html}
<shiny-tool-result></shiny-tool-result>
``````

I can see that...
``````````

### PR DESCRIPTION
**Pull request by gadenbuie**
Fixes #118

### DIFF
--- a/pkg-py/src/shinychat/_chat_types.py
+++ b/pkg-py/src/shinychat/_chat_types.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 from dataclasses import dataclass
 from typing import Literal, TypedDict
 
-from htmltools import HTML, TagChild, Tagifiable, TagList
+from htmltools import HTML, Tag, TagChild, Tagifiable, TagList
 from shiny.session import require_active_session
 
 from ._typing_extensions import NotRequired

### CODE REVIEW

File: pkg-py/src/shinychat/_chat_types.py [hunk 1]
```diff
@@ -26,6 +26,12 @@ def __init__(
     ):
         self.role: Role = role
 
+        is_html = (
+            isinstance(content, HTML)
+            or isinstance(content, TagList)
+            or isinstance(content, Tagifiable)
+        )
```
cpsievert: Most importantly, this adds `Tag`

```suggestion
        is_html = isinstance(content, (Tag, TagList, HTML, Tagifiable))
```

gadenbuie: oh right, thanks!

cpsievert:
Yea, the transform logic definitely complicates things. Let's punt on it for now -- I filed an issue in #124

PR #120, opened 2025-09-05
fix: Open external links in a new tab

### LINKED ISSUE #11: Small suggestion - render links so that they open in a new tab
It appears that when I click on a link within the chat component, the link opens in the current browser tab. Upon returning to the application, the chat component is empty. If this behavior is consistent for all users, it would be beneficial for links to open in a new tab by default.

gadenbuie:
Fixes #11

```r
pkgload::load_all("pkg-r")

output_markdown_stream(
  "foo",
  content = "Here's a link to [wikipedia](https://wikipedia.org)."
)

output_markdown_stream(
  "foo",
  content = "Here's a link to [wikipedia](https://wikipedia.org).",
  content_type = "semi-markdown"
)
```


cpsievert:
Have you tested that this works? I can confirm the `rel="noopener noreferrer"` is getting added, but it appears the `target="_blank"` gets stripped out, probably by DOMpurify https://github.com/cure53/DOMPurify/issues/317

gadenbuie:
> Have you tested that this works? I can confirm the `rel="noopener noreferrer"` is getting added, but it appears the `target="_blank"` gets stripped out, probably by DOMpurify https://github.com/cure53/DOMPurify/issues/317

Oh shoot, yeah I saw the `rel` attribute was added and thought it was both. I'll fix the sanitization too

gadenbuie:
@cpsievert okay, now it works! 😃

js/src/utils/_utils.ts
```diff
@@ -103,6 +103,14 @@ sanitizer.addHook("uponSanitizeElement", (node, data) => {
   }
 })
 
+sanitizer.addHook("afterSanitizeAttributes", function (node) {
+  // set all elements owning target to target=_blank
+  if (node.hasAttribute("data-external-link")) {
```
cpsievert: If I understand correctly, this means only markdown links will have `target="_blank"`? 

To support this behavior more broadly, what if we:

1. In the renderer, add `rel="noopener noreferrer"`
2. When `uponSanitizeAttribute` happens,  allow/add `target="_blank"` when `noopener noreferrer` is present
3. Allow one to opt-out of the addition of `target="_blank"`. While I agree it's a better default behavior, I could also see an argument where one prefers allowing the user to opt-into a new tab (especially when bookmarking is enabled).

gadenbuie: yeah we can move all of the logic into the sanitizer. I'll do that. I think it's so much better of a default behavior for shinychat in Shiny apps that I don't think it's worth offering to opting out.

What I do think would be a good feature is to install our own `onclick` handlers and to allow users to register an onclick callback, which would give people a way to do easily add logic like opening a link in an embedded iframe instead of a new tab.

For now, though, I think we should make this the default and come back later with the onclick handler

cpsievert: Yea, that would neat. It would also be nice to have a confirmation modal (like other chat interfaces do)? It does feel like one of those things where the extra click is mostly annoying, but also aligns with our overall "do AI responsibly" stance

gadenbuie: And now it's a real thing 

image

js/src/chat/chat.ts
```diff
@@ -596,6 +602,33 @@ class ChatContainer extends LightElement {
...
+    // Show confirmation dialog and open the link if confirmed
+    showExternalLinkConfirmation(linkEl.href)
+      .then((confirmed) => {
+        if (confirmed) {
+          window.open(linkEl.href, "_blank", "noopener,noreferrer")
+        }
+      })
+      .catch(() => {
+        // If dialog fails for any reason, fall back to opening the link directly
+        window.open(linkEl.href, "_blank", "noopener,noreferrer")
```
cpsievert, line 629: If we're opening links this way, seems we could drop the `afterSanitizeAttributes` hook?

cpsievert: Or do we want to keep it for non-chat situations?

js/src/chat/chat-external-link.ts
```diff
@@ -0,0 +1,177 @@
...
+      try {
+        this.dialog?.showModal()
+      } catch (err) {
+        // If showModal fails, resolve with true to fallback to default behavior
+        resolve(true)
+      }
+    })
+  }
+
+  private createDialog(): void {
```
cpsievert, line 56: I'd prefer this broken down into 2 methods: `createDialog(): HTMLElement` and `attachDialogListeners(x: HTMLElement): void`. Then do all state mutation at the top-level in `connectedCallback()`

js/src/chat/chat-external-link.ts
```diff
@@ -0,0 +1,177 @@
...
+}
+
+/**
+ * Shows a confirmation dialog for external links
+ * @param url The URL to confirm
+ * @returns A promise that resolves to true if confirmed, false otherwise
+ */
+export function showExternalLinkConfirmation(url: string): Promise<boolean> {
+  // Check if the browser supports HTMLDialogElement
+  if (typeof window.HTMLDialogElement !== "undefined") {
```
cpsievert, line 143: Hmm, I'm not sure if this is needed? It'd be a surprise to me if Bootstrap depended on `<dialog>`, and also the browser support is pretty good https://caniuse.com/?search=HTMLDialog

js/src/chat/chat-external-link.ts
```diff
@@ -0,0 +1,177 @@
...
+
+import { LightElement } from "../utils/_utils"
+
+declare global {
+  interface Window {
+    shinychat_always_open_external_links?: boolean
+  }
+}
+
+export class ChatExternalLinkDialog extends LightElement {
```
cpsievert: This could be used in other non-chat scenarios, right? I don't want to make this a requirement, but perhaps worth noting that we could abstract this and re-use if/when we decide to do tool approval?  

```suggestion
export class ExternalLinkDialog extends LightElement {
```

js/src/chat/chat.ts
```diff
@@ -380,6 +382,7 @@ class ChatContainer extends LightElement {
     )
 
     this.inputSentinelObserver.observe(sentinel)
+    this._boundOnExternalLinkClick = this.#onExternalLinkClick.bind(this)
```
cpsievert, line 385: I'm a bit surprised the `.bind()` is necessary?

js/src/chat/chat.ts
gadenbuie: I remember deleting them, but maybe they crept back in?

js/src/chat/chat-external-link.ts
gadenbuie: Yeah that's a good idea! But we own all the code here so we can just rename this when we get to that

js/src/chat/chat.ts
gadenbuie: Yeah, we need it because we add the event listener to the window and not the chat container element

js/src/chat/chat-external-link.ts
gadenbuie: I renamed the class. This also didn't need to be a lit element, it's simpler as just a class, so I dropped the `extends LitElement` part.

js/src/chat/chat-external-link.ts
gadenbuie: 94% is good but also it's easy enough to check for support that I think it's worth keeping this.

js/src/chat/chat.ts
gadenbuie: Oh I see what you mean with this comment. Yeah, we still want links to have the target/rel attributes so they open in a new tab even if you don't go through the external link confirmation -- e.g. you click "always open external links".

js/src/chat/chat-external-link.ts
gadenbuie: I considered this but the event listeners aren't really _state_. Creating the `<dialog>` element and adding its event listeners are part of the same process and you wouldn't want to do them separately.

PR #127, opened 2025-09-08
feat(pkg-py): pick up tool annotation icon and never escape icon or html display

### LINKED ISSUE #116: [python] Use icon when attached as a tool annotation
In R, shinychat will use `title` and `icon` from `tool@annotations`, but in Python the tool UI only uses `extra.display.icon`.

### LINKED ISSUE #117: [python] Tool UI should accept strings for `display.html` and `display.icon`
Currently, if you provide raw HTML in `extra.display.html` or `extra.display.icon`, that text is escaped and rendered literally in the client.

In both cases, we can reasonably assume that strings _are_ html and shouldn't require tool authors to wrap those values in `htmltools.HTML()`.

cpsievert:
Follow up to #107.

Closes #116.
Closes #117.

### DIFF
--- a/pkg-py/src/shinychat/_chat_normalize_chatlas.py
+++ b/pkg-py/src/shinychat/_chat_normalize_chatlas.py
@@ -288,8 +288,18 @@ def tool_result_contents(x: "ContentToolResult") -> Tagifiable:
 
     tool = x.request.tool
     tool_title = None
+    icon = None
     if tool and tool.annotations:
         tool_title = tool.annotations.get("title")
+        icon = tool.annotations.get("extras", {}).get("icon")
+        icon = icon or tool.annotations.get("icon")
+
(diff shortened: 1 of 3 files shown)


pkg-py/src/shinychat/_chat_normalize_chatlas.py
```diff
@@ -288,8 +288,18 @@ def tool_result_contents(x: "ContentToolResult") -> Tagifiable:
 
     tool = x.request.tool
     tool_title = None
+    icon = None
     if tool and tool.annotations:
         tool_title = tool.annotations.get("title")
+        icon = tool.annotations.get("extras", {}).get("icon")
```
gadenbuie, line 294: shoot, it's `extra`, not `extras`: https://github.com/posit-dev/chatlas/pull/173

PR #138, opened 2025-09-26
fix: change send button selector to a specific class instead of generic <button>

### LINKED ISSUE #137: Chat send button behavior affects other (injected) buttons due to non-specific selector
When I add a button near the chat input, the send button stops working properly. The chat's enable/disable behavior (blue when typing, gray when empty) gets applied to the wrong button instead of the actual send button.

This happens because the code uses querySelector("button") which just grabs the first button it finds, not necessarily the send button.

The fix is to give the send button a specific CSS class and update the selectors to target that class instead of just "button".

## Reasoning

I found it when I was trying to [inject a reset button into the text area](https://github.com/DeepanshKhurana/shinychat-reset-demo). I found a workaround in using a styled `a` tag but it just makes more sense to have a specific selector on the button.

I already have somewhat of a fix and am ready to create a pull request.

DeepanshKhurana:
Fixes #137 

This enables usage of JS injection for additional buttons without breaking functionality for those buttons or the send button. This also opens up a potential plugin buttons approach where a user could choose if they want to include more buttons (this is more a suggestion). But mostly, this makes things more customisable for advanced users.

### DIFF
--- a/.github/workflows/py-check.yaml
+++ b/.github/workflows/py-check.yaml
@@ -17,7 +17,7 @@ permissions:
   contents: read
 
 env:
-  UV_VERSION: "0.7.x"
+  UV_VERSION: "0.9.x"
 
 jobs:
   py-test:
@@ -26,11 +26,11 @@ jobs:
       fail-fast: false
       matrix:
         python-version:
-          - "3.9"
           - "3.10"
           - "3.11"
           - "3.12"
           - "3.13"
+          - "3.14"
 
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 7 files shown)


pkg-py/src/shinychat/playwright/_chat.py
```diff
@@ -95,7 +95,7 @@ def __init__(self, page: Page, id: str) -> None:
         self.loc_latest_message = self.loc_messages.locator("> :last-child")
         self.loc_input_container = self.loc.locator("> shiny-chat-input")
         self.loc_input = self.loc_input_container.locator("textarea")
-        self.loc_input_button = self.loc_input_container.locator("button")
+        self.loc_input_button = self.loc_input_container.locator("button.chat-send-button")
```
DeepanshKhurana: I couldn't find an equivalent test for R but I figured it makes sense to repair the Python test.

PR #139, opened 2025-09-30
chore: Migrate front-end logic to React

cpsievert:


js/src/components/chat/useChatState.ts
```diff
@@ -0,0 +1,177 @@
...
+  messages: Message[]
+  inputValue: string
+  inputDisabled: boolean
+
+  // Message operations (for Shiny integration)
+  appendMessage: (message: Message) => void
+  appendMessageChunk: (message: Message) => void
+  clearMessages: () => void
+  removeLoadingMessage: () => void
+  updateUserInput: (update: UpdateUserInput) => void
```
cpsievert, line 15: Seems like we could take some inspiration from #105 and re-work/simplify the event system?

js/src/components/ShinyMarkdownStream.tsx
```diff
@@ -0,0 +1,244 @@
...
+      ShinyMarkdownStreamOutput.#doBind(this)
+    }
+
+    // Callback for when content changes - can be used for custom logic
+    this.dispatchEvent(
+      new CustomEvent("contentchange", {
+        detail: { content: this.content },
+      }),
+    )
+  }
```
cpsievert, line 151: Feels like the binding logic could be moved into the `MarkdownStream` component implementation?

js/src/components/MarkdownStream.tsx
```diff
@@ -0,0 +1,669 @@
...
+      scrollableElementRef.current?.removeEventListener("scroll", onScroll)
+    }
+  }, [onScroll])
+
+  // Render content based on type
+  const renderContent = () => {
+    if (contentType === "text") {
+      return <div dangerouslySetInnerHTML={{ __html: processedContent }} />
+    } else if (contentType === "html") {
+      return <div dangerouslySetInnerHTML={{ __html: processedContent }} />
```
cpsievert, line 644: This could do something closer to what `'react-markdown'` does (parses string into hast tree, then creates a JSX runtime from that) https://github.com/remarkjs/react-markdown/blob/fda7fa560bec901a6103e195f9b1979dab543b17/lib/index.js#L348

PR #149, opened 2025-10-10
chore(pkg-r): Label all observers and reactives for otel

### LINKED ISSUE #148: Label all observers and reactives for otel
See https://github.com/rstudio/bslib/issues/1236

gadenbuie:
Fixes #148

### DIFF
--- a/pkg-r/R/chat_app.R
+++ b/pkg-r/R/chat_app.R
@@ -116,7 +116,7 @@ chat_app <- function(client, ..., bookmark_store = "url") {
   server <- function(input, output, session) {
     chat_mod_server("chat", client)
 
-    shiny::observeEvent(input$close_btn, {
+    shiny::observeEvent(input$close_btn, label = "on_close_btn", {
       shiny::stopApp()
     })
   }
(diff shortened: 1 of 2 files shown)


pkg-r/R/chat_app.R
```diff
@@ -275,8 +275,8 @@ chat_mod_server <- function(
     }
 
     list(
-      last_turn = shiny::reactive(last_turn()),
-      last_input = shiny::reactive(last_input()),
+      last_turn = shiny::reactive(last_turn(), label = "mod_last_turn"),
+      last_input = shiny::reactive(last_input(), label = "mod_last_input"),
```
schloerke, line 279: `read_only`?

```suggestion
      last_turn = shiny::reactive(last_turn(), label = "last_turn_read_only"),
      last_input = shiny::reactive(last_input(), label = "last_input_read_only"),
```

gadenbuie: Hmm I was thinking of it like "this is the return value of the module" vs the internal reactive value used in the module. I almost called these `return_last_value`. But read only works too. I'll ponder

gadenbuie: I'm going to stick with `mod_`. My take is that if I encountered these labels in an otel trace, having never seen them before, I'd immediately wonder why there's a `_read_only` version of the two other reactives. Whereas `mod_` hopefully signals that it's the value returned by the module.

Issue #76, opened 2025-07-16
Can shinychat display the thinking process of inference LLM？


bianchenhao:
display the thinking process of inference models, such as deepseek-r1, instead of just streaming out responses.

now shinychat, no thingking process:
<img width="786" height="204" alt="Image" src="https://github.com/user-attachments/assets/7251f8b5-5b1f-4e7d-b1bb-c298ac308b29" />

<img width="754" height="376" alt="Image" src="https://github.com/user-attachments/assets/ec41d837-ac84-4f60-a74e-5aad67677373" />

deepseek website with deepseek-r1, showing thingking process:

<img width="921" height="878" alt="Image" src="https://github.com/user-attachments/assets/d57cbf3e-b00b-4b5d-b399-8a2f810774cf" />

cpsievert:
Fixed by #208

PR #167, opened 2025-12-19
feat(pkg-r): Support streaming `ContentThinking`

simonpcouch:
Addresses #76, but only on the R side. Adds support for displaying `ContentThinking` during streaming and on reload.

Without corresponding changes in ellmer, this PR doesn't do anything; with `main` ellmer, shinychat behaves as it currently does, just streaming the thinking content as if it were normal assistant text content. With an incoming ellmer PR (that adds a Remotes on this PR), this is how the content streams + reloads:

https://github.com/user-attachments/assets/38df93a6-9d4b-48dc-96a2-94e49476a02d

Here's the script I was using the test in that video:

```r
# ...load dev versions of both packages

# Anthropic with thinking ------------------------------------------
chat_anthropic <- chat_claude(
  model = "claude-sonnet-4-5-20250929",
  params = params(reasoning_tokens = 1024)
)

clipr::write_clip(
  "Write a limerick about the tidyverse, returning only the limerick."
)
chat_app(chat_anthropic)

# OpenAI with reasoning ------------------------------------------
chat_openai <- chat_openai(
  model = "gpt-5.1",
  params = params(reasoning_effort = "high")
)

chat_app(chat_openai)

# Gemini with thinking ------------------------------------------
chat_gemini <- chat_google_gemini(
  model = "gemini-3-flash-preview",
  params = params(reasoning_tokens = 1024)
)

chat_app(chat_gemini)
```

### DIFF
--- a/pkg-r/R/contents_shinychat.R
+++ b/pkg-r/R/contents_shinychat.R
@@ -376,33 +376,33 @@ tool_string <- function(x) {
   }
 }
 
+flush_thinking_buffer <- function(thinking_buffer) {
+  if (length(thinking_buffer) == 0) {
+    return(list())
+  }
+  combined <- ellmer::ContentThinking(paste(thinking_buffer, collapse = "\n\n"))
+  list(combined)
+}
 
-S7::method(contents_shinychat, ellmer::Turn) <- function(content) {
-  contents <- content@contents
-
-  # Consolidate adjacent ContentThinking into single blocks
+consolidate_thinking <- function(contents) {
   consolidated <- list()
   thinking_buffer <- character()
-
-
-  flush_thinking <- function() {
-    if (length(thinking_buffer) > 0) {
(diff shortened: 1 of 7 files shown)


pkg-r/R/contents_shinychat.R
```diff
@@ -358,8 +378,32 @@ tool_string <- function(x) {
 
 
 S7::method(contents_shinychat, ellmer::Turn) <- function(content) {
-  # Process all contents in the turn, filtering out empty results
-  compact(map(content@contents, contents_shinychat))
+  contents <- content@contents
+
+  # Consolidate adjacent ContentThinking into single blocks
```
simonpcouch: Gemini likes to provide multiple sequential thinking blocks, otherwise resulting in UI like this without this change:

adjacent-thinking-blocks

cpsievert:
Thanks for nudging this along @simonpcouch. https://github.com/posit-dev/shinychat/pull/208 should cover what you're needing, but please reach out if not

PR #177, opened 2026-03-02
Paste images into chat

christophsax:
This addresses #176 and seems to work quite well.


cpsievert:
Hi @christophsax, thanks for the well thought out PR, and apologies for the radio silence.

This feature is definitely high up on our priority list. That said, we recently migrated from Lit to React, so this will need a pretty different approach. I think there's also a fair amount of UI polish that we're looking to add to make the experience a good one. 

If you are *really* motivated to own this feature, we may be willing to work with you. However, to be upfront and honest about it, we will probably be *very* particular about what we want, which might be a frustrating experience for both parties.

cpsievert:
Thanks again @christophsax, we'll be taking a much broader view of this feature over in #250

PR #191, opened 2026-04-06
fix(pkg-py): re-send HTMLDependency objects on bookmark restore

### LINKED ISSUE #190: HTMLDependency objects not restored on bookmark restore
## Problem

When a chat message includes HTML content with an `HTMLDependency` (e.g., custom CSS for tool result styling), the dependency is loaded on the client via `Shiny.renderDependenciesAsync()` during the initial session. However, when restoring from a bookmark, the message HTML is restored but the `HTMLDependency` is **not** re-sent to the client. This means any CSS/JS loaded via `HTMLDependency` is missing in the restored session.

## Reproducible example

```python
from pathlib import Path
from typing import Any

from htmltools import HTMLDependency, TagList, tags
from shiny.express import app_opts
from shiny.types import Jsonifiable

app_opts(bookmark_store="server")

CSS_DIR = Path(__file__).parent / "_assets"
CSS_DIR.mkdir(exist_ok=True)
(CSS_DIR / "custom.css").write_text(
    ".custom-card { border: 3px solid red; padding: 12px; }"
)

custom_dep = HTMLDependency(
    name="custom-card",
    version="1.0.0",
    source={"subdir": str(CSS_DIR)},
    stylesheet=[{"href": "custom.css"}],
)


class MockClient:
    def __init__(self) -> None:
        self.turns: list[Any] = []

    async def get_state(self) -> Jsonifiable:
        return {"version": 1, "turns": self.turns}

    async def set_state(self, state: Jsonifiable) -> None:
        assert isinstance(state, dict)
        turns = state["turns"]
        assert isinstance(turns, list)
        self.turns = turns


client = MockClient()

chat = Chat(id="chat")
chat.ui()
chat.enable_bookmarking(client, bookmark_on="response")


@chat.on_user_submit
async def handle_user_input(user_input: str):
    client.turns.append({"role": "user", "content": user_input})
    styled = TagList(
        custom_dep,
        tags.div({"class": "custom-card"}, f"Response to: {user_input}"),
    )
    client.turns.append({"role": "assistant", "content": f"Response to: {user_input}"})
    await chat.append_message(styled)
```

Steps:
1. Run the app and send a message
2. Observe the response has a red border (CSS from `HTMLDependency` is loaded)
3. Open the bookmark URL in a new tab
4. The message HTML is restored, but the red border is gone — the CSS dependency was not re-sent

## Where the issue lies

In `_chat.py`, `_on_bookmark_ui` saves messages via `self.messages(format=MISSING)` which returns `ChatMessageDict` objects (`{content: str, role: str}`). The `HTMLDependency` objects attached to each `TransformedMessage` are not included in the serialized bookmark state.

On restore, `_on_restore_ui` calls `self.append_message(message_dict)` for each saved dict. Since `content` is a plain string at that point, no `HTMLDependency` is extracted or sent to the client.

A possible fix: save the serialized HTML dependency dicts (via `dep.as_dict()`) in the bookmark state alongside the messages, and re-send them to the client during restore (the transport layer already supports an `html_deps` field on message envelopes).

cpsievert:
Closes #190

## Summary
- Save serialized HTML dependency dicts in bookmark state alongside messages (`dep.as_dict()`)
- On restore, re-send the deps to the client before appending messages so CSS/JS is loaded
- Add a dedicated `render_deps` action type (no-op in the reducer — deps are rendered by the transport layer) instead of piggybacking on `remove_loading`
- Dedup saved dependencies by `(name, version)` instead of deep dict equality
- Add Playwright test that verifies CSS from an `HTMLDependency` survives a bookmark round-trip

**Why Python-only:** The R side doesn't have this issue because it reconstructs the display from the ellmer client's restored turns — `contents_shinychat()` rebuilds the full tag tree (including dependencies) on every restore, and `chat_append()` re-registers them with the session. Python's restore path instead saves and re-appends serialized message strings (`{content: str, role: str}`), which don't carry `HTMLDependency` objects.

## Test plan
- [x] New test: `test_bookmark_html_deps` verifies red border CSS is present after bookmark restore
- [x] Existing bookmark test still passes
- [x] Ruff + pyright clean


pkg-py/src/shinychat/_chat.py
```diff
@@ -1590,6 +1602,18 @@ async def _on_restore_ui(state: RestoreState):
                     f"Bookmark value with id (`{resolved_bookmark_id_msgs_str}`) must be a list of messages."
                 )
 
+            # Re-send any HTML dependencies that were saved alongside the
+            # messages. The way we do this is admittedly hacky...ideally
+            # the bookmarked messages would already contain the deps, but
+            # that would require digging into transfomed message logic,
```
```suggestion
            # that would require digging into transformed message logic,
```

pkg-py/src/shinychat/_chat.py
```diff
@@ -976,7 +976,9 @@ async def _send_append_message(
             return
 
         content = message.content_client
-        content_type: ContentType = "html" if isinstance(content, HTML) else "markdown"
+        content_type: ContentType = (
+            "html" if isinstance(content, HTML) else "markdown"
+        )
```
```suggestion
        content_type: ContentType = "html" if isinstance(content, HTML) else "markdown"
```

pkg-py/src/shinychat/_chat.py
```diff
@@ -1590,6 +1607,14 @@ async def _on_restore_ui(state: RestoreState):
                     f"Bookmark value with id (`{resolved_bookmark_id_msgs_str}`) must be a list of messages."
                 )
 
+            # Re-send any HTML dependencies that were saved alongside the
+            # messages.
```
```suggestion
            # Re-render any HTML dependencies needed by the messages.
```

PR #195, opened 2026-04-13
feat(js): replace custom auto-scroll with use-stick-to-bottom

gadenbuie:
https://github.com/user-attachments/assets/656449bf-b7f8-4c02-8398-e1809195bae7

| Streaming | Static |
|--------|--------|
| image | image |

## Summary

- Replace the custom `useAutoScroll` hook with the [`use-stick-to-bottom`](https://github.com/nickmccurdy/use-stick-to-bottom) library by StackBlitz Labs, which uses ResizeObserver and velocity-based spring animation for smoother, more reliable auto-scrolling during streaming.
- Add a floating "scroll to bottom" button that appears when the user scrolls up, with a spinning gradient border animation during active streaming (customizable via `--shiny-chat-streaming-color`).
- Simplify `ChatContainer` by removing the custom scroll detection, IntersectionObserver sentinel, and direction-based re-engage logic in favor of the library's `isAtBottom` / `scrollToBottom` API.

## Verification

1. Open any chat app and send a message — the chat should auto-scroll to new content.
2. Scroll up during a streaming response — a circular "scroll to bottom" button with a spinning gradient border should appear centered above the input.
3. Click the button — it should smoothly scroll to the bottom and disappear.
4. After the response finishes, the button border animation should stop (if still visible).
5. The `--shiny-chat-streaming-color` CSS variable can be overridden to customize the gradient colors.


cpsievert (approved):
Thanks!

PR #202, opened 2026-04-29
chore(pkg-py): rename stored-message test helper

cpsievert:
## Summary
- follow up to #193
- rename the pytest helper from `transformed_message` to `stored_message` for clarity

## Why
After revisiting the broader `.messages()` / bookmark serialization follow-up, we decided not to carry it forward. The real bug from `#193` is already fixed in `main`, and the remaining inconsistency is a pretty edge-case situation that did not justify the additional API and implementation complexity.

This PR keeps only the low-risk test cleanup so the helper name matches the current stored-message terminology.

## Testing
- `uv run pytest -q pkg-py/tests/pytest/test_chat.py`
- `uv run ruff check pkg-py/tests/pytest/test_chat.py`
- `uv run pyright pkg-py/tests/pytest/test_chat.py`

### DIFF
--- a/pkg-py/CHANGELOG.md
+++ b/pkg-py/CHANGELOG.md
@@ -23,7 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * Removed the deprecated `transform` parameter from `.user_input()`. `.user_input()` now always returns the currently stored user message content, which matches the simplified transform handling used by `.messages()`. (#193)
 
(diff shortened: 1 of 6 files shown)


cpsievert:
After revisiting this in the context of #193, we decided not to carry the broader `.messages()`/bookmark serialization follow-up forward. The original fix in `#193` addresses the real bug: preserving HTML dependencies during bookmark restore. The remaining inconsistency is a pretty edge-case situation, and the follow-up added more API/implementation complexity than it justified.

I reduced this PR to the low-risk cleanup that still seemed worth keeping: renaming the outdated `transformed_message` test helper to `stored_message` in `pkg-py/tests/pytest/test_chat.py` so the test terminology matches the current implementation.

PR #211, opened 2026-05-07
Use ContentThinkingDelta for streaming thinking

cpsievert:
## Summary

Thinking content (`ContentThinkingDelta` and `ContentThinking`) now flows through the same normalization and message-sending pipeline as all other content types, instead of being special-cased in the stream loop. This makes thinking a property of the message rather than the code path, so non-streaming thinking (e.g., bookmark restore) works correctly without separate handling.

## Dependencies

- [posit-dev/chatlas#299](https://github.com/posit-dev/chatlas/pull/299) — adds `ContentThinkingDelta` to chatlas
- [tidyverse/ellmer#975](https://github.com/tidyverse/ellmer/pull/975) — adds `ContentThinkingDelta` to ellmer

## What this replaces

Alternative to [#210](https://github.com/posit-dev/shinychat/pull/210), which removed `content_type: "thinking"` from the wire protocol entirely. This PR keeps the typed wire protocol while achieving the server-side simplification.

## TODO

- [ ] Merge https://github.com/tidyverse/ellmer/pull/975
- [ ] After the next Chatlas release, bump its version requirement.

### DIFF
--- a/pkg-py/src/shinychat/_chat.py
+++ b/pkg-py/src/shinychat/_chat.py
@@ -132,8 +132,7 @@ PendingMessage = Tuple[
 ]
 
 
-def _is_content_thinking_delta(msg: Any) -> bool:
-    """Check if a message is a ContentThinkingDelta object from chatlas."""
+def is_content_thinking_delta(msg: object) -> "TypeGuard[chatlas.ContentThinkingDelta]":
     try:
         from chatlas.types import (
             ContentThinkingDelta,  # pyright: ignore[reportAttributeAccessIssue]
@@ -969,7 +968,7 @@ class Chat:
 
         try:
             async for msg in message:
-                if _is_content_thinking_delta(msg):
+                if is_content_thinking_delta(msg):
                     thinking_text: str = msg.thinking
                     thinking_msg = ChatMessage(content=thinking_text, role="assistant")
                     await self._send_append_message(
--- a/pkg-py/src/shinychat/_chat.py
+++ b/pkg-py/src/shinychat/_chat.py
@@ -132,14 +132,14 @@ PendingMessage = Tuple[
 ]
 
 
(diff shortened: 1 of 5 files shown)


gadenbuie (commented):
I think it'd be helpful if you could take a step back and discuss the problem you're wanting to solve with these additional PRs. It's hard to read between the lines of Claude's PR description to understand what part is bugging you and what problem you're wanting to solve. And I think it'd be best to do that in my original PR as a review so we can have that discussion there.

pkg-py/src/shinychat/_chat.py
```diff
@@ -971,27 +968,22 @@ async def _append_message_stream(
...
-                        thinking_msg,
+                        ChatMessage(content=msg.thinking, role="assistant"),
                         chunk=True,
                         content_type_override="thinking",
                     )
+                    if msg.phase == "start":
+                        self._current_stream_message += "<thinking>\n"
+                    self._current_stream_message += msg.thinking
+                    if msg.phase == "end":
+                        self._current_stream_message += "\n</thinking>\n\n"
```
cpsievert: Looking at #208 more closely again, I think this could be simplified. 

```suggestion
                    self._current_stream_message += msg.thinking
```

pkg-py/src/shinychat/_chat.py
```diff
@@ -1651,7 +1612,12 @@ async def _on_restore_ui(state: RestoreState):
                     html_deps=message_dict.get("html_deps"),
                 )
                 self._store_message(stored)
-                await self._send_append_message(stored)
+                await self._send_append_message(
+                    stored,
+                    content_type=resolve_content_type(
+                        message_dict, stored.content
+                    ),
```
cpsievert, line 1619: Currently this won't restore the correct content type. We already have this problem on main though, and I think it's worth us actually fixing this.

PR #213, opened 2026-05-08
feat(pkg-py): segment-based message storage for mixed content type streaming

cpsievert:
## Summary

- Replaces flat string accumulation (`_current_stream_message` + `_current_stream_deps`) with a segment list that preserves per-chunk content type boundaries during streaming
- Adds `segments` to the wire protocol (`MessagePayload`) so the client can faithfully restore mixed-content messages from bookmarks
- Extracts segment logic into `_chat_segments.py` with dedicated helpers for append, copy, serialize, and content/dep extraction

## Motivation

When an assistant response streams a mix of content types, the Python server previously concatenated everything into a single flat string tagged with one `content_type`. This worked fine for live streaming (the client tracked segments internally), but broke on bookmark restore: the server had lost the content type boundaries, so it sent the entire message back tagged as the first content type.

The bug specifically surfaces when a message starts with HTML content and then switches to markdown. On restore, the whole message gets tagged as HTML (the first content type), causing the trailing markdown to be rendered as raw HTML instead of being markdown-processed. (The reverse — markdown first followed by HTML — is less visibly broken because HTML passed through a markdown renderer is often passable.)

The deeper issue is that the server's internal state didn't reflect what the client already knew: a message is an ordered list of typed segments, not a flat string with a single type. This PR aligns the two, making bookmark save/restore correct by construction.

This also lays groundwork for an upcoming PR that introduces an additional content type, which will make mixed-type messages more common and the restore bug more noticeable.

## Test plan

- [x] New unit tests for segment accumulation, serialization, bookmark round-trip, nested stream checkpoints, and per-chunk content type on the wire
- [ ] Manual test: stream a mixed markdown+HTML response, bookmark it, restore it, verify rendering matches the original


cpsievert:
Superseded by #243

PR #219, opened 2026-05-13
feat: auto-promote suggestion lists to clickable card grid

### LINKED ISSUE #218: Suggestion lists
When `<suggestion>` elements are organized in a list (i.e. a markdown list containing only `<suggestion>` items), we could promote the list into a card view. We could also support a `title` attribute on `<suggestion>` that would be used in the card in this case.

gadenbuie:
Closes #218

https://github.com/user-attachments/assets/8acaf1ef-8052-4ccf-bd48-4c994c9626a2

## Summary

A markdown list (`<ul>` or `<ol>`) in which every item is a single `<span class="suggestion">` element is now rendered as a grid of clickable cards instead of inline chips. The existing inline-chip mechanism is untouched.

**Why a rehype plugin, not a React component.** Promoting at the HAST stage keeps the suggestion element as the click target, so `ChatContainer`'s existing `target.closest(".suggestion, [data-suggestion]")` handler keeps working unchanged. The plugin annotates the existing tree with layout classes, wraps the suggestion body, optionally extracts `title` into a heading, and **locks `data-suggestion`** to the body text — required because injecting a title node would otherwise contaminate the existing `el.textContent` submission fallback.

**Streaming.** A trailing list is marked `data-pending` and renders as "Generating suggestions…" with the streaming dot inline. The qualification rule for the in-progress `<li>` is **evidence-based**: it must be empty, contain a complete suggestion element, or hold a text node matching a prefix of a `<span ...>` opening tag (`PARTIAL_SPAN_OPENING_RE`). Plain bulleted lists like `* Merge conflicts that affect files` therefore never enter the pending state. `finalizePendingSuggestionLists` runs at stream end, path-copying the pending spine and re-validating against the strict rule — the cached HAST is never mutated.

**Roving tabindex.** Suggestion grids get one Tab stop each; arrow keys / Home / End rove between cards with wrap. Setup is lazy on first focusin (marked with `data-roved`) and reset on focus-out so a re-rendered grid is re-initialised on next entry. `data-last-clicked` survives for the lifetime of the message DOM.

**Public CSS vars.** `--shiny-chat-suggestion-card-min-width` (default `175px`) and `--shiny-chat-suggestion-card-reveal-stagger` (default `100ms`) let callers tune layout and motion without touching shinychat selectors.

**Tests.** 342 passing (30 files). Unit tests for the plugin (qualifying / non-qualifying shapes, partial-span evidence rule, streaming round-trip on cached HAST, aria-label, roles), plus integration tests covering the click → submit invariant (body text only, never title-prefixed), `data-last-clicked` lifecycle, and roving navigation.

## Verification

Render a qualifying list in any chat message:

\`\`\`markdown
Here are a few directions we could take this:

- <span class="suggestion" title="Worked example">Show me a worked example with code.</span>
- <span class="suggestion" title="Common pitfalls">What are the common mistakes people make here?</span>
- <span class="suggestion" title="Go deeper">Explain the underlying mechanism in more detail.</span>
\`\`\`

Confirm:

- Tabs into the grid once; arrows rove; Enter/Space submits; activated card stays outlined.
- A plain bulleted list streamed at end-of-message does **not** flash "Generating suggestions…".
- A streaming suggestion list shows the placeholder while incoming, then transitions cleanly to cards.
- \`prefers-reduced-motion: reduce\` suppresses hover lift, chevron slide, card reveal, and pending pulse.

<details>
<summary><strong>Live demo — Python (chatlas + Anthropic)</strong></summary>

\`_dev/agents/suggestions/app.py\` drives a real model with a system prompt that teaches it the suggestion-list syntax. Requires an \`ANTHROPIC_API_KEY\`:

\`\`\`bash
uv run shiny run _dev/agents/suggestions/app.py
\`\`\`

</details>

<details>
<summary><strong>Live demo — R (ellmer + Anthropic)</strong></summary>

\`_dev/agents/suggestions/app.R\` is the R equivalent. Requires \`ANTHROPIC_API_KEY\`:

\`\`\`bash
Rscript -e "shiny::runApp('_dev/agents/suggestions/app.R')"
\`\`\`

</details>

<details>
<summary><strong>Streaming debug harness (no API key needed)</strong></summary>

\`_dev/agents/suggestions/debug.R\` injects hand-crafted chunked streams so the pending → cards transition is reproducible at arbitrary speed. Sidebar exposes a stream preset (ul titled / ul body-only / ol titled / ol body-only / plain bulleted negative control), a 1–6 suggestion-count slider, and a chunk-delay slider.

\`\`\`bash
Rscript -e "shiny::runApp('_dev/agents/suggestions/debug.R')"
\`\`\`

Useful for verifying the partial-\`<span>\` evidence rule and reduced-motion behaviour.

</details>


js/src/markdown/plugins/rehypeSuggestionCards.ts
```diff
@@ -0,0 +1,418 @@
...
+    const existingDs = suggestionEl.properties.dataSuggestion
+    if (!(typeof existingDs === "string" && existingDs !== "")) {
+      suggestionEl.properties.dataSuggestion = bodyText
+    }
+
+    // aria-label: always set (overwriting any label from rehypeAccessibleSuggestions)
+    // so the announcement reflects the full visible card content.
+    if (ordered) {
+      const n = itemIndex + 1
+      suggestionEl.properties.ariaLabel = titleStr
```
cpsievert, line 142: I think we may want to preserve an existing `aria-label` here.

In `promoteListToCards()`, we currently always replace `ariaLabel` with one derived from the card title/body. That seems to undo the behavior in `rehypeAccessibleSuggestions`, which only synthesizes an accessible name when one isn’t already present.

Would it make sense to keep an existing `ariaLabel` if present, and only generate the card-specific label otherwise?

gadenbuie: yeah good point (although it's extreeeeeeemely unlikely that these elements will come with an `aria-label`)

cpsievert (approved):
Nice, thanks!

cpsievert:
Oh, almost forgot. Let's make sure to mention this in the news / changelog.

cpsievert:
Now that I'm giving this a test run on some real examples, I noticed sublists don't seem to be supported at the moment. Does it feel like we should also be supporting something like this? 

```markdown
Here are a few ideas to get you started:

* **Analyze the data**
  * <span class="suggestion">What is the average body mass of the penguins?</span>
  * <span class="suggestion">How many species are in this dataset?</span>
* **Create visualizations**
  * <span class="suggestion">Show me a bar chart of penguin counts by species.</span>
  * <span class="suggestion">Plot a scatter plot of bill length vs. bill depth.</span>
* **Filter and sort records**
  * <span class="suggestion">Show me all penguins from the Biscoe island.</span>
  * <span class="suggestion">Sort the penguins by flipper length, largest to smallest.</span>
```

gadenbuie:
> Now that I'm giving this a test run on some real examples, I noticed sublists don't seem to be supported at the moment. Does it feel like we should also be supporting something like this?

It's certainly an interesting example, but my take at the moment is that it's complicated enough that it we might not want to. It's also relatively easy to prompt into a different layout, e.g. 

```
**Analyze the data**

* <span class="suggestion">What is the average body mass of the penguins?</span>
* <span class="suggestion">How many species are in this dataset?</span>
```

or

```
### Create visualizations

* <span class="suggestion">Show me a bar chart of penguin counts by species.</span>
* <span class="suggestion">Plot a scatter plot of bill length vs. bill depth.</span>
```

I'll open an issue to come back to this

PR #221, opened 2026-05-14
feat: add stream cancellation support

gadenbuie:
## Summary

Adds the ability to cancel in-progress AI response streams. When enabled, a stop button appears during streaming and users can also press Escape to cancel.

**R package:** `chat_ui()` gains an `enable_cancel` parameter. For manual usage, observe `input$<id>_cancel` and call `ctrl$cancel()` on an `ellmer::stream_controller()` to cooperatively stop the stream. The chat module (`chat_mod_server()`) handles all of this automatically — it creates a fresh `stream_controller()` per request and wires up the cancel observer internally, so `chat_mod_ui()` enables cancellation with no extra server code required.

**Python package:** `chat_ui()` and `ChatExpress.ui()` gain an `enable_cancel` parameter. Like the manual path in R, cancellation requires server-side wiring: observe `input.<id>_cancel` and call `ctrl.cancel()` on a chatlas `StreamController`. The docstrings and getting-started docs include complete Shiny Express and Core examples.

**Client-side (JS):** The stop button replaces the send button while streaming. The cancel flow is cooperative: clicking stop dispatches a `CANCEL_REQUESTED` action and fires a Shiny input event (`input.<id>_cancel`). When the stream ends, the reducer finalizes the message with `cancelled: true`, rendering a "Response cancelled" indicator. The Escape key listener is scoped to the `shiny-chat-container` element and uses a ref pattern to avoid re-registering on every streaming state change.

**Tests:** 10 new reducer tests covering cancel state transitions and 6 new ChatInput component tests for stop button visibility and interaction.

### R (automatic via module)

```r
library(shiny)
library(bslib)
library(shinychat)

ui <- page_fillable(
  chat_mod_ui("chat")
)

  client <- ellmer::chat_anthropic()
  chat_mod_server("chat", client)
}

shinyApp(ui, server)
```

The stop button appears during streaming. Click it or press Escape to cancel.

### Python (manual wiring)

```python
from shiny import reactive, App, ui
from chatlas import ChatAnthropic, StreamController

app_ui = ui.page_fillable(
    chat_ui("chat", enable_cancel=True),
)

    client = ChatAnthropic()
    chat = Chat("chat")
    ctrl = StreamController()

    @chat.on_user_submit
    async def handle_user_input(user_input):
        response = await client.stream_async(user_input, controller=ctrl)

    @reactive.effect
    @reactive.event(input.chat_cancel)
    def handle_cancel():
        ctrl.cancel()

app = App(app_ui, server)
```


cpsievert:
Sweet! Can you help me understand why we need the `nonlocal` pattern?

gadenbuie:
> Sweet! Can you help me understand why we need the `nonlocal` pattern?

Good catch, I had meant to go back to revisit that. I've cleaned things up, we only need one stream controller instance, so no need for the reactive value or `nonlocal` stuff that was in the examples.

cpsievert (approved):
🚀

PR #223, opened 2026-05-15
feat: greeting messages — first-class welcome UI for chat

### LINKED ISSUE #217: Greeting messages
We should support greeting messages natively as a first-class feature, separate from an initial chat message.

A collection of design-related thoughts:

* Greetings could be static or generated messages.
  * Static messages should appear on start up and should skip scroll-to-bottom behavior
  * For a generated greeting, we'd submit a prompt on startup that would be streamed into the UI
* Greetings may appear as stand-alone content or as assistant messages
  * My preference is for greetings to be stand-alone by default, i.e. not within an assistant chat message.
  * But some users will likely want to have the message appear to come from the assistant.
* Greetings may or may not be shown to the model, i.e. included in the message history
  * In some cases, the greeting is a message to the user and the conversation starts with the first user message
  * In other cases, authors may want to include the greeting in the conversation history.
  * This likely interacts with the choice to have the message appear to come from the assistant (but should be independently controlled).
* Authors may want to keep or dismiss the greeting once the first message is sent.
* Greetings may include prompt suggestions.

gadenbuie:
Closes #217

## Summary

Adds native greeting messages as a first-class feature in shinychat. Greetings are a distinct UI element — separate from the conversation message stream — that welcome users when the chat loads and can include suggestion cards.

https://github.com/user-attachments/assets/839c7c85-b7e7-43f4-a5ab-57714aee1601

### Content paths

- **Static**: `chat_ui(greeting = "## Welcome!")` — renders immediately on mount
- **Dynamic**: `chat_set_greeting()` / `Chat.set_greeting()` with computed or streamed content
- **Generated**: server listens for `<id>_greeting_requested` and streams an LLM-generated greeting

### Key design decisions

- **Stand-alone visual**: greetings are centered, styled separately from assistant messages, with their own entrance/dismiss animations
- **Dismissible by default**: greeting fades out on first user message; reappears if chat is cleared
- **`greeting_requested` input**: a state-driven Shiny input that fires when the chat is visible, empty, and has no greeting — handles init, clear, bookmark restore, and hidden-tab cases uniformly
- **`chat_clear(greeting=TRUE)`**: clears both messages and greeting, naturally triggering `greeting_requested` for regeneration
- **Module support (R)**: `chat_mod_server(greeting=)` accepts a static value or a function; functions are called automatically on init and after `clear(greeting=TRUE)`, with optional client injection for one-arg functions

### Protocol

Five new action types for the stand-alone greeting path: `greeting`, `greeting_start`, `greeting_chunk`, `greeting_end`, `greeting_clear`. Content flows through the same HAST markdown pipeline as regular messages.

### What was dropped during implementation

- `as_assistant_message` — removed; authors who want assistant-styled welcomes use `chat_append()` directly
- `include_in_history` — removed; shinychat doesn't own model history

## Verification

Set `ANTHROPIC_API_KEY` and run any of the example apps below. Things to verify:
- Greeting streams in on load with suggestion cards
- Clicking a suggestion fills the input
- Sending a message dismisses the greeting with a fade animation
- "Clear chat" re-shows the original greeting
- "Clear chat + greeting" regenerates a fresh greeting from the LLM

<details>
<summary><code>haiku_app.R</code> — R with manual chat plumbing</summary>

```r
library(shiny)
library(bslib)
library(shinychat)
library(ellmer)

MODEL <- "claude-haiku-4-5"

GREETING_PROMPT <- r"(
You generate short, friendly welcome greetings for a programming explainer app.
Respond with markdown only — no preamble, no closing remarks.

Use exactly this structure:

## <a brief, varied welcome heading>

<one inviting sentence about explaining programming and data science concepts>

**Try one of these:**

- <span class="suggestion">programming or data science topic 1</span>
- <span class="suggestion">programming or data science topic 2</span>
- <span class="suggestion">programming or data science topic 3</span>

Suggestions should be specific concepts like "What is a closure?",
"Explain tidy evaluation", "How does gradient descent work?", etc.
Vary the wording each time. Keep the entire greeting under 60 words.
)"

CHAT_PROMPT <- "You explain programming and data science concepts concisely in 1-2 sentences. Use markdown when appropriate."

ui <- page_fillable(
  card(
    card_header(
      "Haiku greeting demo (manual)",
      toolbar(
        align = "right",
        toolbar_input_button("clear_chat", "Clear chat", icon = icon("eraser")),
        toolbar_input_button(
          "clear_chat_and_greeting",
          "Clear chat + greeting",
          icon = icon("trash")
        )
      )
    ),
    chat_ui("chat", placeholder = "Ask me anything...", fill = TRUE)
  )
)

  client <- ellmer::chat_anthropic(
    model = MODEL,
    system_prompt = CHAT_PROMPT,
    echo = "none"
  )

    stream <- client$stream_async(input$chat_user_input)
    chat_append("chat", stream)
  })

  # Fires when the chat is visible, empty, and has no greeting.
  observeEvent(input$chat_greeting_requested, {
    greeter <- ellmer::chat_anthropic(
      model = MODEL,
      system_prompt = GREETING_PROMPT,
      echo = "none"
    )
    stream <- greeter$stream_async("Generate a welcome greeting for the user.")
    chat_set_greeting("chat", chat_greeting(stream))
  })

  observeEvent(input$clear_chat, {
    chat_clear("chat")
  })

  observeEvent(input$clear_chat_and_greeting, {
    chat_clear("chat", greeting = TRUE)
  })
}

shinyApp(ui, server)
```

</details>

<details>
<summary><code>haiku_mod_app.R</code> — R with <code>chat_mod_server()</code></summary>

```r
library(shiny)
library(bslib)
library(shinychat)
library(ellmer)

MODEL <- "claude-haiku-4-5"

GREETING_PROMPT <- r"(
You generate short, friendly welcome greetings for a programming explainer app.
Respond with markdown only — no preamble, no closing remarks.

Use exactly this structure:

## <a brief, varied welcome heading>

<one inviting sentence about explaining programming and data science concepts>

**Try one of these:**

- <span class="suggestion">programming or data science topic 1</span>
- <span class="suggestion">programming or data science topic 2</span>
- <span class="suggestion">programming or data science topic 3</span>

Suggestions should be specific concepts like "What is a closure?",
"Explain tidy evaluation", "How does gradient descent work?", etc.
Vary the wording each time. Keep the entire greeting under 60 words.
)"

CHAT_PROMPT <- "You explain programming and data science concepts concisely in 1-2 sentences. Use markdown when appropriate."

ui <- page_fillable(
  card(
    card_header(
      "Haiku greeting demo (module)",
      toolbar(
        align = "right",
        toolbar_input_button("clear_chat", "Clear chat", icon = icon("eraser")),
        toolbar_input_button(
          "clear_chat_and_greeting",
          "Clear chat + greeting",
          icon = icon("trash")
        )
      )
    ),
    chat_mod_ui("chat")
  )
)

  client <- ellmer::chat_anthropic(
    model = MODEL,
    system_prompt = CHAT_PROMPT,
    echo = "none"
  )

  chat <- chat_mod_server("chat", client, greeting = function() {
    greeter <- ellmer::chat_anthropic(
      model = MODEL,
      system_prompt = GREETING_PROMPT,
      echo = "none"
    )
    greeter$stream_async("Generate a welcome greeting for the user.")
  })

  observeEvent(input$clear_chat, {
    chat$clear()
  })

  observeEvent(input$clear_chat_and_greeting, {
    chat$clear(greeting = TRUE)
  })
}

shinyApp(ui, server)
```

</details>

<details>
<summary><code>haiku_app.py</code> — Python</summary>

```python
"""
Live Haiku-powered greeting demo (Python).

Generates a fresh greeting via the state-driven `<id>_greeting_requested`
Shiny input. Set ANTHROPIC_API_KEY before running.

    uv run shiny run _dev/agents/greeting/haiku_app.py
"""

from shiny import App, reactive, ui

from shinychat import Chat, chat_greeting, chat_ui

MODEL = "claude-haiku-4-5"

GREETING_PROMPT = """\
You generate short, friendly welcome greetings for a programming explainer app.
Respond with markdown only — no preamble, no closing remarks.

Use exactly this structure:

## <a brief, varied welcome heading>

<one inviting sentence about explaining programming and data science concepts>

**Try one of these:**

- <span class="suggestion">programming or data science topic 1</span>
- <span class="suggestion">programming or data science topic 2</span>
- <span class="suggestion">programming or data science topic 3</span>

Suggestions should be specific concepts like "What is a closure?",
"Explain tidy evaluation", "How does gradient descent work?", etc.
Vary the wording each time. Keep the entire greeting under 60 words.
"""

CHAT_PROMPT = (
    "You explain programming and data science concepts concisely "
    "in 1-2 sentences. Use markdown when appropriate."
)


app_ui = ui.page_fillable(
    ui.card(
        ui.card_header(
            "Haiku greeting demo (Python)",
            ui.toolbar(
                ui.toolbar_input_button("clear_chat", "Clear", tooltip="Clear chat"),
                ui.toolbar_input_button(
                    "clear_chat_and_greeting",
                    "Reset",
                    tooltip="Clear chat + greeting",
                ),
                align="right",
            ),
        ),
        chat_ui("chat", placeholder="Ask me anything...", fill=True),
    ),
)


    chat = Chat(id="chat")
    client = ChatAnthropic(model=MODEL, system_prompt=CHAT_PROMPT)

    @chat.on_user_submit
    async def _(user_input: str):
        stream = await client.stream_async(user_input)
        await chat.append_message_stream(stream)

    # Fires when the chat is visible, empty, and has no greeting.
    @reactive.effect
    @reactive.event(input.chat_greeting_requested)
    async def _generate_greeting():
        greeter = ChatAnthropic(model=MODEL, system_prompt=GREETING_PROMPT)
        stream = await greeter.stream_async("Generate a welcome greeting for the user.")
        await chat.set_greeting(chat_greeting(stream))

    @reactive.effect
    @reactive.event(input.clear_chat)
    async def _clear_chat():
        await chat.clear_messages()

    @reactive.effect
    @reactive.event(input.clear_chat_and_greeting)
    async def _clear_chat_and_greeting():
        await chat.clear_messages(greeting=True)


app = App(app_ui, server)
```

</details>


pkg-py/src/shinychat/_chat_types.py
```diff
@@ -129,6 +168,103 @@ def __init__(
...
+            return
+
+        deps: list[HTMLDependency] = []
+        content_type: ContentType = "markdown"
+        if isinstance(content, str):
+            pass
+        elif isinstance(content, HTML):
+            content_type = "html"
+            content = str(content)
+        else:
```
cpsievert: What's the reasoning for this logic being different from `ChatMessage`? In that case, I'm pretty sure `HTML` will get the ` split_html_islands()` treatment, which I think is important.

gadenbuie: Good catch, thank you! Fixed in bd83aa0

pkg-py/src/shinychat/__init__.py
```diff
@@ -1,9 +1,11 @@
-from ._chat import Chat, chat_ui
+from ._chat import Chat, ChatGreeting, chat_greeting, chat_ui
 from ._chat_normalize import message_content, message_content_chunk
 from ._markdown_stream import MarkdownStream, output_markdown_stream
 
 __all__ = [
     "Chat",
+    "ChatGreeting",
```
cpsievert: I think I'd prefer to move this export to the `types` module.

gadenbuie: Good idea; fixed in abad234

cpsievert:
I noticed that `set_greeting()` is a silent no-op after the greeting has been dismissed — it updates the internal state but nothing visible changes until `clear(greeting=TRUE)` resets things (via `computeGreetingVisibility` in `state.ts:142-150`).

I think that makes sense as a design choice, but I could see someone calling `set_greeting("Try asking about X\!")` mid-conversation and being confused when nothing happens. Might be worth calling this out in the `set_greeting()` docstrings so people don't have to discover it through debugging.

js/src/chat/state.ts
```diff
@@ -41,6 +42,17 @@ export interface ChatMessageData {
...
+export interface GreetingData {
+  content: string
+  contentType: ContentType
+  streaming: boolean
+  visible: boolean
+  dismissed: boolean
+  dismissing: boolean
+  options: GreetingOptions
+  blocks: ContentBlock[]
+}
```
cpsievert, line 52: The dismiss lifecycle uses three correlated booleans (`visible`, `dismissed`, `dismissing`) that have 8 possible combinations, most of which are invalid — the reducer has to be careful not to produce them. Would a single `status: "visible" | "dismissing" | "dismissed"` enum work here? It would make impossible states unrepresentable while keeping the same animation coordination. The layout-stability trick with `[data-greeting-dismissing]` and the centering CSS is clever — not suggesting changing any of that.

pkg-r/R/chat.R
```diff
@@ -626,9 +811,269 @@ rlang::on_load(
...
+        break
+      }
+
+      if (is.character(msg)) {
+        ui <- list(html = msg, deps = NULL)
+        chunk_content_type <- "markdown"
+      } else {
+        ui <- process_ui(pre_process_ui(msg), session)
+        chunk_content_type <- "html"
+      }
```
cpsievert: I think this means that `HTML()` will get treated like markdown.

pkg-py/src/shinychat/_chat_types.py
```diff
@@ -129,6 +168,98 @@ def __init__(
...
+class ChatGreeting:
+    def __init__(
+        self,
+        content: Union[str, HTML, Tag, TagList, "AsyncIterator[str]"],
+        *,
+        dismissible: bool = True,
+    ):
+        self.dismissible = dismissible
+
+        if isinstance(content, AsyncIterator):
```
cpsievert: I think this check is too narrow. In Python, a value can be a valid async stream for `async for` without being an `AsyncIterator` specifically. If that happens here, we will fall through into the non-stream path instead.

I think the fix here should be:
- widen the types from `AsyncIterator` to `AsyncIterable`
- update the runtime check accordingly, not just the annotation
- add a test with a custom async-iterable object, not just a native async generator

I would keep this scoped to async iterables only for this PR. Supporting sync iterables would be a separate feature expansion.

js/src/chat/state.ts
gadenbuie: Good call — collapsed into `status: "visible" | "dismissing" | "dismissed"` in c11a87e.

pkg-r/R/chat.R
gadenbuie: Yep, the streaming path was missing the `!inherits(msg, "html")` guard that the static path had. Fixed in 2bbe686.

pkg-py/src/shinychat/_chat_types.py
gadenbuie: Widened to `AsyncIterable` in both the type annotations and runtime checks, plus added a test with a custom async-iterable class. f3e8696.

gadenbuie:
> Might be worth calling this out in the `set_greeting()` docstrings so people don't have to discover it through debugging.


Agreed — added a note in the `set_greeting()` descriptions for R and Python explaining that post-dismiss calls update hidden content and pointing to `clear(greeting = TRUE)` to make a new greeting visible. e441679.

pkg-r/R/chat.R
```diff
@@ -626,9 +811,274 @@ rlang::on_load(
...
+    id,
+    stream,
+    options,
+    session = shiny::getDefaultReactiveDomain()
+  ) {
+    send_greeting_action <- function(action) {
+      send_chat_action(id, action = action, session = session)
+    }
+
+    send_greeting_action(list(
```
cpsievert, line 1049: If the greeting stream throws after `greeting_start`, we never send the terminal `greeting_end`, so the client can be left in a permanent `streaming: true` state with a half-rendered greeting. Suggestion: mirror the normal streamed-message cleanup path here by moving `greeting_end` into a `finally`/guaranteed teardown block. That way, once `greeting_start` is sent, the UI always receives a terminal action even on error. The same issue pattern appears in the Python greeting stream path as well.

gadenbuie: Good catch! Fixed in 0ee71c4 and 41fc71f.

R now mirrors the `chat_append_stream` pattern — `promises::catch` around the `chat_set_greeting_stream` call sends `greeting_end` and logs a warning on error. Python uses `try/finally`, matching its existing message stream cleanup.

js/src/chat/state.ts
```diff
@@ -703,6 +750,117 @@ export function chatReducer(state: ChatState, action: AnyAction): ChatState {
...
+      return {
+        ...state,
+        greeting: { ...greeting, content, contentType: chunkType, blocks },
+      }
+    }
+
+    case "greeting_end": {
+      const greeting = state.greeting
+      if (!greeting?.streaming) return state
+      return { ...state, greeting: { ...greeting, streaming: false } }
```
cpsievert, line 861: Because both greeting stream senders start with an empty `greeting_start` and always close with `greeting_end` (`pkg-r/R/chat.R` and `pkg-py/src/shinychat/_chat.py`), a stream that fails before its first chunk reaches this branch with `blocks = []` and `content === ""`. Finalizing it here leaves `state.greeting` non-null, which seems to suppress future `greeting_requested` events and can wedge auto-generated greetings after a transient failure.

Would it make sense to treat an empty finished greeting as "no greeting" here and return `greeting: null` instead? A reducer test for `greeting_start("") -> greeting_end` with no chunks in between would cover the case.

cpsievert (approved):
Once the issue outlined in https://github.com/posit-dev/shinychat/pull/223#discussion_r3277809829 is addressed, then this LGTM, thanks!

js/src/chat/state.ts
gadenbuie: Fixed in 2a554c3 — `greeting_end` now clears `state.greeting` to `null` when the stream produced no content, so `greeting_requested` can fire again after a transient failure.

PR #225, opened 2026-05-15
feat: support images and PDFs in tool results

### LINKED ISSUE #161: Update tool result handling to support S7 content types (images, PDFs)
## Background
The ellmer package recently added support for tool results that return non-JSON content types like images and PDFs ([ellmer PR #735](https://github.com/tidyverse/ellmer/pull/735)).

## Problem
Currently, shinychat tries to convert tool results to JSON strings using `tool_string()`, which fails when the tool result contains S7 content types like `content_image()` or `content_pdf()`.

## Proposed Solution
Modify the tool result handling to:
- Detect S7 content types
- Render these content types appropriately in the chat interface
- Fallback to existing JSON conversion for other content types

## Example Scenario
```r
library(ellmer)

# Tool that returns an image
cat_image_tool <- tool(
  function() {
    # Simulate generating a cat image
    tmpf <- tempfile(fileext = ".jpg")
    download.file("[https://placekitten.com/200/300](vscode-file://vscode-app/Applications/Positron.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html#)", tmpf, quiet = 
TRUE)
    content_image_file(tmpf)
  },
  name = "get_cat_image",
  description = "Generate a random cat image"
)

# Create a chat
chat <- chat_openai()
chat$register_tool(cat_image_tool)

shinychat::chat_app(chat)
```

## Related Work
- ellmer PR: https://github.com/tidyverse/ellmer/pull/735
- Specific issue: Handling of `tool_string()` for S7 content types

/cc @gadenbuie

gadenbuie:
Closes #106 (rewritten)
Closes #161

## Summary

When ellmer/chatlas tools return `ContentImage` or `ContentPDF` objects, shinychat now renders them visually inside tool result cards: images as `` elements, PDFs as a filename badge with a PDF icon.

Tools can also return **mixed content lists** — e.g. `list(ContentText("summary"), content_image_file("plot.png"), ContentText("stats"))` — and each item renders in order with appropriate formatting. Text items in mixed lists default to markdown rendering.

### How it works

The server sends a new `valueType = "content_extra"` with a JSON array of typed items:

```json
[
  {"type": "text", "value": "Generated 50 samples", "value_type": "markdown"},
  {"type": "image", "src": "data:image/png;base64,..."},
  {"type": "text", "value": "Mean: 0.123, SD: 0.987", "value_type": "markdown"}
]
```

### Changes by layer

- **JS**: New `content_extra` value type in `ToolResult` with renderers for `image`, `pdf`, and `text` item types. `ContentExtraText` dispatches on `value_type` (markdown, html, code, plain text). New `PdfBadge` component.
- **R**: `tool_default_display()` detects `Content` objects in tool result values and serializes them via `as_content_extra_item_or_text()`. `ContentText` extracts `@text`; images/PDFs serialize to structured data.
- **Python**: Handles both streaming path (raw `ContentToolResult` with image/PDF values) and turn-replay path (standalone content items after chatlas expansion). Suppresses `<tool-content>` XML wrapper tags from chatlas expansion.

### Design notes

- Text items in mixed content default to `value_type: "markdown"`. There's currently no per-item API to override this — a future enhancement could add that via the display options.
- Mixed content lists expect all items to be `Content` subclasses (e.g. `ContentText()` for text, not bare strings).
- Custom `display` options (html/markdown/text) still take full precedence over automatic content-extra rendering.

<details>
<summary>R example app</summary>

```r
library(shiny)
library(bslib)
library(ellmer)
library(shinychat)

tool_result_open <- function(x, ...) {
  ContentToolResult(
    value = x,
    extra = list(
      display = list(open = TRUE, show_request = FALSE, ...)
    )
  )
}

get_sample_image <- tool(
  function(title = "Random Normal Distribution") {
    path <- tempfile(fileext = ".png")
    png(path, width = 600, height = 400)
    hist(rnorm(100), main = title, col = "steelblue", border = "white")
    dev.off()
    tool_result_open(content_image_file(path))
  },
  name = "get_sample_image",
  description = "Generate a histogram plot of random normal data and return it as an image.",
  title = type_string("Title for the histogram plot.")
)

get_image_from_url <- tool(
  function(label = "placeholder") {
    w <- sample(2:6, 1) * 100
    h <- sample(2:6, 1) * 100
    tool_result_open(
      content_image_url(sprintf("https://loremflickr.com/%d/%d?lock=42", w, h))
    )
  },
  name = "get_image_from_url",
  description = "Return a remote image from a URL to demonstrate URL-based image rendering.",
  label = type_string(
    "A label for the image (not displayed, just for identification)."
  )
)

get_sample_pdf <- tool(
  function(title = "Sample PDF Report") {
    path <- tempfile(fileext = ".pdf")
    pdf(path, width = 7, height = 5)
    plot(1:10, main = title, type = "b", col = "darkred")
    dev.off()
    tool_result_open(content_pdf_file(path))
  },
  name = "get_sample_pdf",
  description = "Generate a simple PDF plot and return it to demonstrate PDF rendering in tool results.",
  title = type_string("Title for the PDF plot.")
)

get_mixed_content <- tool(
  function(n = 50L) {
    path <- tempfile(fileext = ".png")
    data <- rnorm(n)
    png(path, width = 600, height = 400)
    hist(
      data,
      main = sprintf("Distribution (n=%d)", n),
      col = "steelblue",
      border = "white"
    )
    dev.off()
    tool_result_open(
      list(
        ContentText(sprintf("Generated %d samples", n)),
        content_image_file(path),
        ContentText(
          sprintf("```\nMean: %.3f, SD: %.3f\n```", mean(data), sd(data))
        )
      )
    )
  },
  name = "get_mixed_content",
  description = "Generate random data and return a summary with an embedded histogram. Returns text, image, and text interleaved.",
  n = type_integer("Number of random samples to generate.")
)

ui <- page_fillable(
  chat_mod_ui("chat")
)

  chat <- chat_anthropic(
    model = "claude-haiku-4-5",
    system_prompt = paste(
      "You are a helpful assistant demonstrating shinychat's tool result rendering.",
      "When asked, use the available tools to generate and return images or PDFs.",
      "You have three tools:",
      "- get_sample_image: generates a histogram plot as an inline image",
      "- get_image_from_url: returns a remote image URL as an image",
      "- get_sample_pdf: generates a PDF and returns it as a PDF badge",
      "- get_mixed_content: generates data with text summary + histogram + stats interleaved",
      "Use them when the user asks to see images or PDFs."
    )
  )
  chat$register_tool(get_sample_image)
  chat$register_tool(get_image_from_url)
  chat$register_tool(get_sample_pdf)
  chat$register_tool(get_mixed_content)
  chat_mod_server("chat", chat)
}

shinyApp(ui, server)
```

</details>

<details>
<summary>Python example app</summary>

```python
import random
import tempfile

import matplotlib.pyplot as plt
import numpy as np
from chatlas import (
    ChatAnthropic,
    ContentToolResult,
    content_image_file,
    content_image_url,
    content_pdf_file,
)
from chatlas.types import ContentText
from shinychat.types import ToolResultDisplay


def tool_result_open(x):
    return ContentToolResult(
        value=x,
        extra={"display": ToolResultDisplay(open=True, show_request=False)},
    )


def get_sample_image(title: str = "Random Normal Distribution"):
    """
    Generate a histogram plot of random normal data and return it as an image.

    Parameters
    ----------
    title
        Title for the histogram plot.
    """
    path = tempfile.mktemp(suffix=".png")
    fig, ax = plt.subplots(figsize=(6, 4))
    ax.hist(np.random.normal(size=100), color="steelblue", edgecolor="white")
    ax.set_title(title)
    fig.savefig(path)
    plt.close(fig)
    return tool_result_open(content_image_file(path))


def get_image_from_url(label: str = "placeholder"):
    """
    Return a remote image from a URL to demonstrate URL-based image rendering.

    Parameters
    ----------
    label
        A label for the image (not displayed, just for identification).
    """
    w = random.choice(range(2, 7)) * 100
    h = random.choice(range(2, 7)) * 100
    return tool_result_open(
        content_image_url(f"https://loremflickr.com/{w}/{h}?lock=42")
    )


def get_sample_pdf(title: str = "Sample PDF Report"):
    """
    Generate a simple PDF plot and return it to demonstrate PDF rendering in tool results.

    Parameters
    ----------
    title
        Title for the PDF plot.
    """
    path = tempfile.mktemp(suffix=".pdf")
    fig, ax = plt.subplots(figsize=(7, 5))
    ax.plot(range(1, 11), marker="o", color="darkred")
    ax.set_title(title)
    fig.savefig(path, format="pdf")
    plt.close(fig)
    return tool_result_open(content_pdf_file(path))


def get_mixed_content(n: int = 50):
    """
    Generate random data and return a summary with an embedded histogram.
    Returns text, image, and text interleaved.

    Parameters
    ----------
    n
        Number of random samples to generate.
    """
    data = np.random.normal(size=n)
    path = tempfile.mktemp(suffix=".png")
    fig, ax = plt.subplots(figsize=(6, 4))
    ax.hist(data, color="steelblue", edgecolor="white")
    ax.set_title(f"Distribution (n={n})")
    fig.savefig(path)
    plt.close(fig)
    return tool_result_open([
        ContentText(text=f"Generated {n} samples"),
        content_image_file(path),
        ContentText(text=f"Mean: {data.mean():.3f}, SD: {data.std():.3f}"),
    ])


chat_client = ChatAnthropic(
    model="claude-haiku-4-5",
    system_prompt=(
        "You are a helpful assistant demonstrating shinychat's tool result rendering. "
        "When asked, use the available tools to generate and return images or PDFs. "
        "You have three tools: "
        "- get_sample_image: generates a histogram plot as an inline image "
        "- get_image_from_url: returns a remote image URL as an image "
        "- get_sample_pdf: generates a PDF and returns it as a PDF badge "
        "- get_mixed_content: generates data with text summary + histogram + stats interleaved "
        "Use them when the user asks to see images or PDFs."
    ),
)
chat_client.register_tool(get_sample_image)
chat_client.register_tool(get_image_from_url)
chat_client.register_tool(get_sample_pdf)
chat_client.register_tool(get_mixed_content)

app_ui = ui.page_fillable(
    chat_ui("chat"),
    fillable_mobile=True,
)


def server(input: Inputs, output: Outputs, session: Session):
    chat = Chat(id="chat")

    @chat.on_user_submit
    async def _():
        user = chat.user_input()
        response = await chat_client.stream_async(user, content="all")


app = App(app_ui, server)
```

</details>

## Test plan

- [ ] R: Tool returning `content_image_file()` renders inline image in tool result card
- [ ] R: Tool returning `content_image_url()` renders remote image
- [ ] R: Tool returning `content_pdf_file()` renders PDF badge
- [ ] R: Mixed content list (`ContentText` + image + `ContentText`) renders interleaved
- [ ] Python: Same four tool types render correctly during streaming
- [ ] Python: Images/PDFs render correctly when replaying from history
- [ ] Python: No `<tool-content>` XML wrapper text visible in UI
- [ ] Error tool results still display error text in code block
- [ ] Custom `display` (html/markdown/text) still takes precedence over content-extra


pkg-py/src/shinychat/_chat.py
cpsievert, line 1: I think the changes in this file and `_chat_provider_types.py` could be reverted?

pkg-py/src/shinychat/_chat_normalize_chatlas.py
```diff
@@ -100,7 +100,55 @@ def tagify(self):
         )
 
 
-ValueType = Literal["html", "markdown", "text", "code"]
+ValueType = Literal["html", "markdown", "text", "code", "content_extra"]
+
+
+def _is_content(value: object) -> bool:
+    from chatlas._content import Content
```
cpsievert: Two things that apply to the helpers in this file:

1. I generally prefer to avoid the underscore prefix (in a module that already has one). 
2. These types should probably come `from chatlas.types`

pkg-py/src/shinychat/_chat_normalize.py
```diff
@@ -144,12 +144,51 @@ def _(chunk: Content):
 
     @message_content.register
     def _(message: ContentText):
-        return ChatMessage(content=message.text)
+        text = message.text
+        if text.startswith("<tool-content") or text.startswith(
+            "</tool-content"
+        ):
```
cpsievert, line 153: When does this become relevant and why does it seem specific to Python?

pkg-py/src/shinychat/_chat_normalize.py
```diff
@@ -144,12 +144,51 @@ def _(chunk: Content):
...
+        def _(chunk: ContentImageRemote):
+            return message_content(chunk)
+
+        @message_content.register
+        def _(message: ContentPDF):
+            return ChatMessage(content=message.filename or "document.pdf")
+
+        @message_content_chunk.register
+        def _(chunk: ContentPDF):
+            return message_content(chunk)
```
cpsievert: I'm a bit surprised that we need this given that we already have:

```python
@message_content.register
def _(message: Content):
    return ChatMessage(content=str(message))

@message_content_chunk.register
def _(chunk: Content):
    return message_content(chunk)
```

Is there a reason why these need to be different?

pkg-py/src/shinychat/_chat_normalize.py
gadenbuie: Yeah, this is chatlas-specific. Claude may have overlooked something fundamental, but it appeared that in https://github.com/posit-dev/chatlas/pull/231 chatlas eagerly rewrites the message history to move image and PDF content out of the tool results. In R, I modified provider logic individually so that the transformation happens only when preparing the message history for the provider API. In chatlas, I think you went with an earlier optimization to do this transformation eagerly so that you don't need separate updates/logic for each provider. That makes sense, but we have to undo that logic when replaying chat history.

Here's the relevant section from my research with Claude:

> ### Key Difference: Two Processing Paths
> 
> Because chatlas expands tool results at `UserTurn` construction time, shinychat Python encounters image/PDF content in two distinct situations:
> 
> **Path 1 — Streaming (individual content items)**
> During a live chat stream, shinychat receives raw `ContentToolResult` objects before they are assembled into a `UserTurn`. At this point, `value` may still be a `ContentImageInline` or `ContentPDF` object. `message_content.register` for `ContentToolResult` calls `tool_result_contents()` → `tool_result_display()`.
> 
> **Path 2 — Turn replay (history)**
> When a full `Turn` is processed (e.g., replaying history), the `UserTurn` has already been expanded. The `ContentToolResult.value` is the placeholder string, and the image/PDF appears as a standalone `ContentImageInline` or `ContentPDF` item in `turn.contents`. The `Turn` handler in `_chat_normalize.py` iterates `message.contents` and calls `message_content(x)` on each — so standalone images/PDFs need their own `message_content.register` handler.

cpsievert: Ah, okay, thanks for tracking that down for me. Maybe it'd be worth a comment highlighting that this is specific to chatlas' tool result content expansion logic?

pkg-py/src/shinychat/_chat_normalize.py
gadenbuie: The generic `Content` handler uses `str(message)`, which for image types produces markdown image syntax (``) rather than explicit `` HTML. 

There was a moment during development when I was treating plain text in the mixed-content tool result blocks as "code" rather than as "markdown"; in that version the markdown syntax wasn't being rendered as an image.

While that would render correctly now that we default text to markdown, the specific handlers are worth keeping for a few reasons:

1. **Performance with large payloads** — inline base64 images can be megabytes. Raw `` HTML passes through the markdown parser cheaply; markdown image syntax forces the parser to process the entire data URI.
2. **Concatenation safety** — the `Turn` handler concatenates content strings. Raw HTML is more robust than markdown syntax when content gets joined without separators (e.g., `Some text` won't render as an image).
3. **Explicit control** — we get predictable output rather than depending on chatlas's `__str__` format, which could change.

For PDFs, `str(ContentPDF)` gives `<PDF document file=report.pdf size=4 bytes>` — the specific handler gives just the filename, which is a more useful standalone representation.

I think they're worth keeping; I think we'll want the control over output format and it gives us the flexibility to add features in the future that build on targeted and specific formatting of the content objects.

pkg-py/src/shinychat/_chat_normalize_chatlas.py
gadenbuie: FYI

```
>>> import chatlas
>>> chatlas.types.ContentText
chatlas._content.ContentText
>>> chatlas.types.ContentPDF
AttributeError: module 'chatlas.types' has no attribute 'ContentPDF'
>>> chatlas.__version__
'0.18.0'
```

pkg-py/src/shinychat/_chat_normalize.py
cpsievert: Sounds good, thanks.

pkg-py/src/shinychat/_chat.py
```diff
@@ -1679,7 +1679,6 @@ def ui(
         height: "CssUnit" = "auto",
         fill: bool = True,
         icon_assistant: HTML | Tag | TagList | None = None,
-        enable_cancel: bool = False,
```
cpsievert, line 1682: Oops, I think the revert went a little bit too far by removing this.

pkg-py/src/shinychat/_chat_normalize_chatlas.py
cpsievert: https://github.com/posit-dev/chatlas/pull/312

I also could push out a release if you want to require it (up to you).

cpsievert (approved):
Thanks!

pkg-py/src/shinychat/_chat_normalize_chatlas.py
gadenbuie: No need for a release at the moment. I added a compat shim in ebef237 that tries `chatlas.types` first, and falls back to `chatlas._content`. Also, I added a test that fails when our min chatlas exceeds 0.18.0 to remind us to clean it up. And I cleaned up similar import shims that aren't needed for the current min chatlas version.

pkg-py/src/shinychat/_chat_normalize.py
gadenbuie: Added a comment in 5ae112f explaining the chatlas `expand_tool_result()` context.

PR #226, opened 2026-05-15
fix: align custom .tagify() annotations with htmltools Tagified contract

schloerke:
## Summary

The upcoming htmltools 0.7.0 release ([posit-dev/py-htmltools#120](https://github.com/posit-dev/py-htmltools/pull/120), closes posit-dev/py-htmltools#116) refactors tagified content into immutable sibling classes and tightens the `Tagifiable.tagify()` Protocol return type to `Tagified`. `Tagified` is a tighter union than the previous bare `TagList | Tag | MetadataNode | str | HTML`: it excludes the un-resolved `Tagifiable` arm and is the only tagified-shape alias exported by htmltools — downstream `.tagify()` implementations should annotate `-> Tagified` exclusively. Custom `.tagify()` implementations whose return is wider than `Tagified` no longer satisfy the protocol in strict-mode type checkers.

Changes:

- **`_chat_bookmark.py`**: change `def tagify(self) -> TagChild: return ""` to `-> Tagified`. The previous `TagChild` annotation was wider than the protocol return type (`TagChild` includes `float`/`None`/`Sequence`/etc.).
- **`_chat_normalize_chatlas.py`**: annotate both `def tagify(self):` methods as `-> Tagified` and call `.tagify()` on the returned `Tag(…)` so the value is properly tagified (matches the recursive invariant the protocol now expresses).

Pyright (basic mode, as configured) is already clean with the released htmltools; these changes make the code also correct under the upcoming protocol tightening.

## Dependency

Depends on [posit-dev/py-htmltools#120](https://github.com/posit-dev/py-htmltools/pull/120). The `[tool.uv.sources]` git pin currently points at the PR branch; revert to released `htmltools>=0.7.0` once it ships on PyPI.

## Test plan

- [x] Pyright against the new htmltools → 0 errors
- [x] Functional behavior is unchanged — `.tagify()` on a leaf-only `Tag` is a no-op at runtime

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -131,7 +131,7 @@ r-docs-preview: ## [r] Build R docs
 	cd $(PATH_PKG_R) && Rscript -e "pkgdown::preview_site()"
 
 .PHONY: py-setup
-py-setup: py-install-pr-shiny  ## [py] Setup python environment
+py-setup: py-sync py-install-pr-shiny  ## [py] Setup python environment
 
 .PHONY: py-sync
 py-sync:
@@ -145,8 +145,12 @@ py-sync:
 # (shiny → shinychat → shiny), so we install it post-sync with
 # --no-deps. Drop this target once py-shiny#2244 merges and a release
 # ships.
+#
+# This target intentionally does NOT depend on `py-sync` so callers
+# (Makefile and CI) can choose when to sync; CI syncs in a separate
+# step before invoking this target.
 .PHONY: py-install-pr-shiny
-py-install-pr-shiny: py-sync
+py-install-pr-shiny:
 	@echo ""
 	@echo "📦 Installing py-shiny from posit-dev/py-shiny#2244"
 	# Use `uv pip install` (not `uv sync`) so the install runs at the
@@ -184,13 +188,15 @@ py-check-tests:  ## [py] Run python tests
 py-check-types:  ## [py] Run python type checks
 	@echo ""
 	@echo "📝 Checking types with pyright"
-	uv run pyright
+	# --no-sync so the post-sync py-shiny#2244 override isn't reverted.
+	uv run --no-sync pyright
 
 .PHONY: py-check-format
 py-check-format:
 	@echo ""
 	@echo "📐 Checking format with ruff"
-	uv run ruff check pkg-py --config pyproject.toml
+	# --no-sync so the post-sync py-shiny#2244 override isn't reverted.
+	uv run --no-sync ruff check pkg-py --config pyproject.toml
 
 .PHONY: py-format
 py-format: ## [py] Format python code
--- a/Makefile
+++ b/Makefile
@@ -131,8 +131,35 @@ r-docs-preview: ## [r] Build R docs
 	cd $(PATH_PKG_R) && Rscript -e "pkgdown::preview_site()"
 
 .PHONY: py-setup
-py-setup:  ## [py] Setup python environment
+py-setup: py-install-pr-shiny  ## [py] Setup python environment
+
+.PHONY: py-sync
+py-sync:
 	uv sync --all-extras --all-groups --upgrade
+
+# TODO(htmltools-116): remove this target before merging. Installs
+# py-shiny from the posit-dev/py-shiny#2244 branch for the
+# Renderer.tagify() / App._render_page() fixes required by the new
+# Tagifiable contract (htmltools 0.7.0 / py-htmltools#120). Pinning
+# shiny via [tool.uv.sources] would create a circular resolution
+# (shiny → shinychat → shiny), so we install it post-sync with
+# --no-deps. Drop this target once py-shiny#2244 merges and a release
+# ships.
+.PHONY: py-install-pr-shiny
+py-install-pr-shiny: py-sync
+	@echo ""
+	@echo "📦 Installing py-shiny from posit-dev/py-shiny#2244"
+	# Use `uv pip install` (not `uv sync`) so the install runs at the
+	# package level and bypasses workspace resolution — `uv sync` would
+	# choke on the shiny → shinychat circular reference. `--no-deps` so
+	# uv doesn't replace the workspace shinychat with a PyPI build to
+	# satisfy py-shiny's own `shinychat>=0.1.0` requirement. Install
+	# `opentelemetry-api` separately because py-shiny#2244 added it as
+	# a new dep that the released shiny on PyPI didn't have, so `uv
+	# sync` won't have brought it in either.
+	uv pip install --reinstall --no-deps \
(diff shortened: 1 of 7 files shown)


pkg-py/src/shinychat/_chat_bookmark.py
```diff
@@ -10,9 +10,13 @@
...
 )
 
-from htmltools import TagChild
-
 if TYPE_CHECKING:
+    # `Tagified` is only available in htmltools >= 0.7.0; keep the
+    # import behind TYPE_CHECKING so this module still loads against
+    # the currently-released htmltools. The `from __future__ import
+    # annotations` line above defers the annotation, so the name
+    # never needs to resolve at runtime.
```
schloerke: Remove

pkg-py/src/shinychat/_chat_normalize_chatlas.py
```diff
@@ -21,6 +21,13 @@
 if TYPE_CHECKING:
     from chatlas.types import ContentToolRequest, ContentToolResult
 
+    # `Tagified` is only available in htmltools >= 0.7.0; keep the
+    # import behind TYPE_CHECKING so this module still loads against
+    # the currently-released htmltools. The `from __future__ import
+    # annotations` line above defers the annotation, so the name
+    # never needs to resolve at runtime.
```
schloerke: Remove

Makefile
```diff
@@ -131,9 +131,40 @@ r-docs-preview: ## [r] Build R docs
...
+	# package level and bypasses workspace resolution — `uv sync` would
+	# choke on the shiny → shinychat circular reference. `--no-deps` so
+	# uv doesn't replace the workspace shinychat with a PyPI build to
+	# satisfy py-shiny's own `shinychat>=0.1.0` requirement. Install
+	# `opentelemetry-api` separately because py-shiny#2244 added it as
+	# a new dep that the released shiny on PyPI didn't have, so `uv
+	# sync` won't have brought it in either.
+	uv pip install --reinstall --no-deps \
+		"shiny @ git+https://github.com/posit-dev/py-shiny.git@schloerke/htmltools-105-tagified-types"
+	uv pip install "opentelemetry-api>=1.20.0"
```
schloerke: Is this needed still?

Makefile
```diff
@@ -147,20 +178,25 @@ py-check-tox:  ## [py] Run python checks across versions with tox
 py-check-tests:  ## [py] Run python tests
 	@echo ""
 	@echo "🧪 Running tests with pytest"
-	uv run playwright install
-	uv run pytest
+	# Use --no-sync so the post-sync overrides applied in py-setup
+	# (notably the py-shiny#2244 install) aren't reverted by an
+	# implicit re-sync before each `uv run`. See py-install-pr-shiny.
```
schloerke: Are the changes in this file still needed?

Makefile
```diff
@@ -131,9 +131,39 @@ r-docs-preview: ## [r] Build R docs
...
+#
+# Intentionally does NOT depend on `py-sync` so callers (Makefile and
+# CI) can choose when to sync; CI syncs in a separate step before
+# invoking this target.
+.PHONY: py-install-pr-shiny
+py-install-pr-shiny:
+	@echo ""
+	@echo "📦 Installing latest shiny over the synced venv"
+	uv pip install --reinstall --no-deps "shiny>=1.6.2"
+	uv pip install "opentelemetry-api>=1.20.0"
```
schloerke, line 175: Is this still needed?

pyproject.toml
```diff
@@ -59,6 +59,14 @@ dev = [
 [tool.uv]
 upgrade-package = ["shinychat"]
 
+# NB: shiny>=1.6.2 depends on `shinychat>=0.1.0` (this project),
+# which uv refuses to satisfy through the workspace. As a result
+# `uv lock`/`uv sync` can't resolve a tree with the latest shiny
+# (the lockfile is pinned to an older shiny that lacks the
+# Renderer.tagify() / App._render_page() updates for htmltools
+# 0.7.0). The Makefile's `py-install-pr-shiny` target installs the
+# latest shiny on top of the synced venv as a post-sync workaround.
```
schloerke: This comment still needed?

pkg-py/src/shinychat/_chat_bookmark.py
schloerke: Removed in 19fe225. Dropped the 5-line "Tagified only in 0.7.0..." rationale — `htmltools>=0.7.0` is the hard floor, so the guard text is redundant. Kept the import under `TYPE_CHECKING` (annotation-only symbol).

pkg-py/src/shinychat/_chat_normalize_chatlas.py
schloerke: Removed in 19fe225 — same as the `_chat_bookmark.py` change.

Makefile
schloerke: Yes, `--no-sync` is still needed. Without it, each `uv run` implicitly re-syncs the lockfile-pinned shiny (1.4.0) back over the latest shiny that `py-install-pr-shiny` just installed (1.6.2), and the Tagifiable-contract fixes in shiny 1.6.2 disappear — every test that renders htmltools breaks.

I cleaned this up in 19fe225 (collapsed three duplicated `--no-sync` notes into one paragraph above the block) and 5bd0af8 (added the exit conditions to `py-install-pr-shiny` so it's clear when all of this can be deleted — when py-shiny drops its `shinychat>=0.1.0` runtime dep, or [astral-sh/uv#9610](https://github.com/astral-sh/uv/issues/9610) lands).

Makefile
schloerke: Yes. shiny 1.6.2 declares \`opentelemetry-api>=1.20.0\` as a hard runtime dep ([PyPI metadata](https://pypi.org/pypi/shiny/1.6.2/json)), so importing shiny without it raises \`ModuleNotFoundError: No module named 'opentelemetry'\` (verified locally: \`uv pip uninstall opentelemetry-api\` then \`uv run python -c 'import shiny'\` fails).

We use \`--no-deps\` on the shiny install (to stop uv from replacing the workspace shinychat with the PyPI build to satisfy shiny's own \`shinychat>=0.1.0\` — verified: dropping \`--no-deps\` installs \`shinychat==0.3.1\` from PyPI and shadows our local code). \`--no-deps\` skips opentelemetry along with everything else, so we re-install it explicitly.

pyproject.toml
schloerke: Trimmed in 19fe225 from a 7-line wall to 3 lines pointing at the Makefile target. The pointer is still worth keeping IMO — without it, anyone reading \`[tool.uv]\` won't know why \`uv sync\` alone produces a venv that fails the test suite.

PR #229, opened 2026-05-21
chore(py): drop unnecessary shiny post-sync workaround from #226

schloerke:
## Summary

- #226 added a `py-install-pr-shiny` Makefile target, matching `--no-sync` flags on the `py-check-*` targets, a CI step, and a `[tool.uv]` block in `pyproject.toml` to work around what was described as a circular `shiny ↔ shinychat` resolution failure. Empirically that resolution failure doesn't happen with uv 0.9.x — a fresh `uv lock` / `uv sync` resolves `shiny==1.6.2`, the editable workspace `shinychat`, and `opentelemetry-api==1.42.1` cleanly, and `make py-check-{format,types,tests}` (65 tests) all pass against the resulting env with no override.
- Likely root cause at PR time: a stale local `uv.lock` pinned to an older `shiny` (the lockfile is `.gitignore`d, so a fresh checkout has none). The workaround papered over that rather than refreshing the lockfile.
- Removes: the `py-install-pr-shiny` Makefile target and its rationale block, the intermediate `py-sync` target (inlines `uv sync` back into `py-setup`, matching pre-#226 shape), the `--no-sync` flags on `py-check-tests/-types/-format`, the dedicated CI workaround step, and the `[tool.uv]` block in `pyproject.toml`.

## Test plan

- [x] Delete local `uv.lock`, run `uv sync --all-extras --all-groups --upgrade` — installs `shiny==1.6.2` and `opentelemetry-api==1.42.1` with no override
- [x] `make py-check-format` — passes
- [x] `make py-check-types` — 0 errors, 0 warnings
- [x] `make py-check-tests` — 65 passed
- [ ] CI green on this PR

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -56,13 +56,6 @@ dev = [
     {include-group = "test"},
 ]
 
-[tool.uv]
-upgrade-package = ["shinychat"]
-
-# shiny>=1.6.2 depends on `shinychat` (this project), which uv can't
-# satisfy through the workspace. See `py-install-pr-shiny` in the
-# Makefile for the post-sync workaround.
-
 [build-system]
 requires = ["hatchling", "hatch-vcs"]
 build-backend = "hatchling.build"
--- a/.github/workflows/py-check.yaml
+++ b/.github/workflows/py-check.yaml
@@ -34,6 +34,16 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
+        with:
+          # Need full history + tags so hatch-vcs computes a real
+          # workspace version (e.g. 0.3.2.devN+gSHA, derived from the
+          # latest py/v* tag) instead of the fallback 0.1.dev1+gSHA.
+          # The fallback sorts before 0.1.0 (PEP 440), so it can't
+          # satisfy shiny>=1.5's transitive `shinychat>=0.1.0` and
+          # uv silently downgrades shiny to 1.4.0, which lacks the
+          # htmltools-0.7.0 Tagifiable fixes.
+          fetch-depth: 0
+          fetch-tags: true
 
       - name: 🚀 Install uv
         uses: astral-sh/setup-uv@v6.1.0
--- a/.github/workflows/py-check.yaml
+++ b/.github/workflows/py-check.yaml
@@ -46,15 +46,6 @@ jobs:
       - name: 📦 Install the project
         run: uv sync --python ${{ matrix.python-version }} --all-extras --all-groups
 
-      # Workaround for the shiny → shinychat circular dep that
-      # prevents uv from resolving a tree with shiny>=1.6.2. The
-      # lockfile is pinned to an older shiny that lacks the
-      # Renderer.tagify() / App._render_page() updates required by
-      # htmltools 0.7.0; this step installs the latest shiny on top
-      # of the synced venv. See `py-install-pr-shiny` in the Makefile.
-      - name: 📦 Install latest shiny (post-sync workaround)
-        run: make py-install-pr-shiny
-
       - name: 📜 Show uv.lock
         run: cat uv.lock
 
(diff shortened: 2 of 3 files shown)


pyproject.toml
```diff
@@ -56,13 +56,6 @@ dev = [
     {include-group = "test"},
 ]
 
-[tool.uv]
-upgrade-package = ["shinychat"]
```
schloerke, line 60: Note: `upgrade-package = ["shinychat"]` was **not** added by #226 — it pre-dates that PR (introduced in #56, the monorepo reorg). Dropping it along with the comment is intentional but technically out of scope for "revert #226's uv changes."

It's a no-op in practice: `shinychat` is the editable workspace package (`source = { editable = "." }`), so there is only one candidate version uv can ever pick for the workspace name. `upgrade-package` only matters when uv is choosing between multiple registry versions — there's nothing for it to upgrade *to* here.

Verified empirically with this directive removed: a fresh `uv lock` resolves `shiny==1.6.2` + workspace `shinychat` + `opentelemetry-api==1.42.1` cleanly, and `make py-check-{format,types,tests}` (65 tests) pass.

schloerke:
## Why CI was failing — root cause

The initial cleanup-only commit broke CI. Surprising on the surface (locally everything resolves to `shiny==1.6.2`), but the resolution chain in CI is:

1. uv tries `shiny==1.6.2` (latest matching `shiny>=1.4.0`).
2. shiny 1.6.2 needs `htmltools>=0.7.0` ✅ **and** `shinychat>=0.1.0` ❌.
3. To satisfy the `shinychat>=0.1.0` transitive constraint, uv has two candidates:
   - **Workspace `shinychat`** — but in CI, `actions/checkout@v4` does a shallow clone with no tags, so `hatch-vcs` falls back to `0.1.dev1+gSHA`. Per PEP 440, `0.1.dev1 < 0.1.0`, so the workspace version doesn't satisfy `>=0.1.0`. ❌
   - **PyPI `shinychat`** — uv refuses ("the name is shadowed by your project"). ❌
4. Both fail. uv backtracks through 1.6.1, 1.6.0, 1.5.0 — all carry the same `shinychat>=0.1.0` dep.
5. uv lands on `shiny==1.4.0`, which predates the shinychat dep. `htmltools>=0.6.0` (its requirement) is satisfied by our `htmltools>=0.7.0`. Resolution succeeds — with the wrong shiny.

That older shiny lacks the `Renderer.tagify()` / `App._render_page()` updates required by the htmltools 0.7.0 Tagifiable contract, hence the `TypeError: Expected a fully tagified value…` in every test.

Locally this didn't reproduce because my working tree has full tag history, so `hatch-vcs` yields `0.3.2.devN+gSHA` — which **is** ≥ 0.1.0 — so the workspace candidate at step 3 wins and uv stays on shiny 1.6.2.

### The fix (e71c681)

Make the CI checkout fetch tags so `hatch-vcs` produces a real workspace version instead of the `0.1.dev1` fallback:

```yaml
- uses: actions/checkout@v4
  with:
    fetch-depth: 0
    fetch-tags: true
```

This is the actual fix for the breakage that #226's post-sync `uv pip install --reinstall --no-deps shiny>=1.6.2` was working around — the comments in #226 attributed it to "uv refuses to satisfy that transitive dep through the workspace," but the real trigger was the shallow-checkout dev version sorting before `0.1.0`.

### Caveat on my earlier comment

Above in [a previous comment](https://github.com/posit-dev/shinychat/pull/229#discussion_r3284024660) I said `upgrade-package = ["shinychat"]` was a no-op. That's true *given a real workspace version*, but with the `0.1.dev1` fallback, it doesn't help either — uv still can't pick a version satisfying `>=0.1.0`. The directive is genuinely redundant; the tags are what matter.

PR #233, opened 2026-05-26
feat(py): Add client= parameter to Chat for batteries-included chatlas apps

cpsievert:
## Summary

Building a chatbot with shinychat + chatlas currently requires manually wiring up streaming, cancellation, and bookmarking — roughly 20-30 lines of boilerplate that look the same in every app. This PR adds a `client=` parameter to `Chat` that handles all of it automatically.

**Express:**
```python

chat = Chat(id="chat", client=ChatAnthropic())
chat.ui()
```

**Core:**
```python
from shiny import App, ui

app_ui = ui.page_fillable(
    chat_ui("chat", enable_cancel=True),
)

    chat = Chat("chat", client=ChatAnthropic())

app = App(app_ui, server)
```

No new API surface to learn — just one parameter on the class you already know. Client-specific methods live on a `ChatClient` wrapper at `chat.client`:

- `.value` — the raw chatlas client
- `.set(new_client, sync=True)` — swap models mid-session (deferred if streaming)
- `.clear(client_history="clear")` — reset UI and client turn history

In Express mode, `enable_cancel` defaults to `True` when a client is set.

### Other changes

- `enable_bookmarking()` no longer raises when no bookmark store is configured — it silently no-ops instead.
- Docs lead with `Chat(client=...)` as the recommended path.
- R: removed a dead `bookmark_store` local in `chat_restore()` (it was assigned but never used). No behavior change.


cpsievert:
Reflecting on this a bit more, it seems worth exploring adding a `client` param to `__init__`. Will look into it tomorrow

cpsievert:
@gadenbuie there's a bit more review and polish I'd like to do here, but before I do, would be great to have a gut check from you in terms of the overall shape of the proposal here.

pkg-py/docs/index.qmd
```diff
@@ -29,36 +29,48 @@ Or, install the development version of shinychat from [GitHub](https://github.co
...
 
 ```{.python file="app.py"}
+from chatlas import ChatAnthropic
+from shiny.express import app_opts, ui
 
-# Create a chat instance and display it
-chat = Chat(id="chat")
+client = ChatAnthropic(system_prompt="You are a helpful assistant.")
+chat = Chat(id="chat", client=client)
```
gadenbuie, line 44: This is really nice and clean! Overall, I like this approach a lot more than what I proposed using Shiny Modules

pkg-py/docs/index.qmd
```diff
@@ -29,36 +29,48 @@ Or, install the development version of shinychat from [GitHub](https://github.co
...
-```bash
-uv shiny run --reload app.py
-```
+```{.python file="app.py"}
+from shiny import App, ui
+from chatlas import ChatAnthropic
+from shinychat import Chat, chat_ui
+
+app_ui = ui.page_fillable(
+    chat_ui("chat", enable_cancel=True),
```
gadenbuie: This is a tiny bit of a bummer that the user has to know to set `enable_cancel=True`. I wonder if we could handle this with a server-side message instead so that it's not necessary?

cpsievert: Great point. Done in a2d58f5 — the server now auto-enables the stop button when a `client=` is attached, so `enable_cancel=True` is no longer needed.

PR #238, opened 2026-05-28
Improve link navigation for SPA safety and fix dialog bugs

### LINKED ISSUE #201: External link confirmation dialog does not appear in Safari (grey screen instead of modal)
When clicking a link in a shinychat message in Safari, the external link confirmation dialog introduced in #120 does not render. Instead, the page goes grey (the modal backdrop appears to render, but the dialog itself does not). Seems to happen in both the Python and R versions.

The same app behaves correctly in Chrome, Edge, and Firefox: the "External Link" modal shows up with the URL, "Open Link", "Cancel", and "Always open external links" options, as expected. 

Related to #11 and #120.

R:
```
library(shiny)
library(shinychat)

ui <- bslib::page_fluid(
  chat_ui(
    "chat",
    messages = "Hello! Click this link to [wikipedia](https://wikipedia.org) to trigger the external link dialog."
  )
)

}

shinyApp(ui, server)
```

Python:
```
from shiny.express import ui

chat = ui.Chat(id="chat")
chat.ui(
    messages=[
        "Hello! Click this link to [wikipedia](https://wikipedia.org) to trigger the external link dialog."
    ]
)
```

cpsievert:
## Open navigating links in new tabs to preserve app state

Shiny apps are single-page applications — clicking a link that navigates away from the current page loses the session state, with no way to recover it. Previously, only absolute URLs (`https://...`) were intercepted by the external link dialog; relative links like `/about` or `page.html` would navigate the browser tab directly with no warning and no protection.

Now all navigating links in assistant messages (everything except `#fragment` anchors and non-HTTP schemes like `mailto:`) are opened in a new tab, preserving the app session in the original tab. The click handler uses `window.open()` rather than relying solely on `target="_blank"`, since the attribute can be stripped by HTML sanitizers in some pipelines.

Same-origin links (both relative paths and same-origin absolute URLs) open in a new tab immediately with no dialog. Cross-origin links continue to show the confirmation dialog. The net effect:

| Link type | Before | After |
|---|---|---|
| Cross-origin absolute | Dialog → new tab | Dialog → new tab (unchanged) |
| Same-origin absolute | Dialog (misleadingly labeled "external") | New tab, no dialog |
| Relative path | Navigates away, loses app state | New tab, preserves state |
| Fragment (`#id`) | Normal | Normal (unchanged) |
| `mailto:`, `tel:`, etc. | Normal | Normal (unchanged) |

(4108c68, 1e062ce)

## Fix Escape key breaking the external link dialog

Pressing Escape to dismiss the dialog left React's `pendingUrl` state set while the browser natively closed the `<dialog>`. Subsequent link clicks would re-render the component but `showModal()` wouldn't fire again (the `useEffect` has `[]` deps), leaving the dialog invisible and broken until page reload. Fixed by listening for the native `cancel` event and clearing state. (2192714)

## Fix external link dialog not rendering in Safari

Closes #201. Bootstrap's `.modal` class sets `height: 100%`, which collapses to zero inside a content-sized `<dialog>` element in Safari. The backdrop rendered (grey overlay) but the modal content was invisible. Fixed with a CSS override: `height: auto` on the `.modal` wrapper so it sizes from content. (ecf1c75)

## Test app

<details>
<summary>Manual test app (Python)</summary>

Type anything and press Enter — the assistant replies with a message containing every link category.

```python
from shiny import App, ui

app_ui = ui.page_fillable(
    chat_ui("chat"),
)

LINKS_MESSAGE = """\
Here are some links to test:

**Cross-origin (should show dialog):**
- [Example.com](https://example.com)
- [Google](https://google.com)

**Same-origin absolute (should open in new tab, no dialog):**
- [This app](http://localhost:8000/)

**Relative paths (should open in new tab, no dialog):**
- [Relative page](about/page)
- [Root path](/some-path)

**Fragment (should scroll, no interception):**
- [Jump to anchor](#test-anchor)

**Non-http schemes (should behave normally):**
- [Email](mailto:test@example.com)

<span id="test-anchor">**Anchor target is here.**</span>
"""


    chat = Chat("chat")

    @chat.on_user_submit
    async def on_user_submit():
        await chat.append_message(LINKS_MESSAGE)


app = App(app_ui, server)
```

</details>


gadenbuie (approved):
I haven't review the code in detail but these all sound like the right fixes to me!

PR #243, opened 2026-06-04
feat: segment-based message storage (mixed content + thinking)

cpsievert:
## Why

> *(scope note: the motivating problem and the substantive fix are Python-side. R and JS get a wire-protocol update for parity, but no R user-facing behavior changes here.)*

In **Shiny for Python**, bookmarking a streamed assistant reply that mixed content types — markdown, a raw-HTML widget (with its `HTMLDependency`), and reasoning — used to lose type and dependency information on restore: the message came back flattened, styling and structure gone. This reworks how messages are represented so those replies round-trip faithfully, and folds model "thinking" into the same model instead of the previous `<thinking>`-tag string workaround.

## What changed

The core shift is from a **single-string** content model to a list of **typed segments**.

- **Before:** a message was one `content` string + one `content_type`. Thinking was smuggled in by wrapping text in `<thinking>…</thinking>` inside the markdown string, then re-parsed on the client. A reply couldn't faithfully represent "markdown, then an HTML widget, then reasoning" — it all collapsed to one type.
- **After:** a message is an ordered list of `{content, content_type}` segments, where `content_type ∈ markdown | html | text | thinking`. Adjacent same-type chunks merge; a type change opens a new segment. Thinking is a first-class segment type rather than a tag convention.

Segments back server-side storage, bookmarks, and the complete-message wire payload (the `message` action). Streaming stays delta-based: an empty segment-shaped `chunk_start` (carrying role/icon) is followed by flat `chunk` deltas, each a single content type, which the client reassembles into its render model; greetings remain flat too. The **Python** side (`StoredMessage`, segment serialize/restore) is where mixed-content replies are now preserved end to end. **R** emits the same payload shapes for parity (always a single segment today), and the **JS** client materializes segments at ingest — but neither gains new storage behavior.

## What's now possible

- A single streamed message can interleave **markdown + HTML + thinking** and survive a bookmark round-trip intact — content types preserved, HTML dependencies (CSS/JS) re-applied. Covered by a new Playwright end-to-end test.
- **Thinking is a first-class content segment.** Structured reasoning (chatlas / ellmer `ContentThinking`) and raw inline `<thinking>` tags both render through the existing thinking panel.

## Approach notes

On the JS client, `main`'s richer **`blocks`** render model — the streaming thinking-tag state machine, topic labels, duration timer, and code-fence handling — is **preserved unchanged** and derived from segments at ingest. (A pure render-time derivation couldn't reproduce the cross-chunk streaming state, so the reducer stays authoritative for live rendering, while segments own storage and the complete-message payload.) The old `<thinking>`-tag buffering hack in `_append_message_stream` is removed.

**Breaking:** bookmark restore assumes segments are present — there's no legacy fallback, so bookmarks written by an earlier version will fail to restore rather than degrade.

## Testing

- JS: 454 vitest tests pass; lint clean
- Python: 87 unit tests pass; pyright 0 errors; ruff clean
- R: 186 testthat tests pass
- New Playwright e2e: mixed markdown + thinking + HTML bookmark round-trip

---

This likely supersedes #213, which targeted the same problem before `main` gained thinking support.


js/src/transport/types.ts
```diff
@@ -6,13 +6,16 @@ export interface GreetingOptions {
...
+}
+
 export type MessagePayload = {
   id?: string
   role: "user" | "assistant"
-  content: string
-  content_type: ContentType
   icon?: string
-  html_deps?: HtmlDep[]
+  segments: MessagePayloadSegment[]
```
cpsievert, line 18: Note: `html_deps` was intentionally removed from `MessagePayload` here, not relocated.

It was a vestigial field — nothing produced or consumed it:
- The Python `MessagePayload` TypedDict never had an `html_deps` key; the server only ever built `{role, content/segments, content_type}`.
- `messagePayloadToData` in `state.ts` never read `msg.html_deps` (before or after this change).

Dependencies travel on the **envelope**, not the message payload. The server attaches them via `_send_action(action, message.html_deps)` → `ShinyChatEnvelope.html_deps`, and the client renders them in `shiny-transport.ts` before dispatching the action:

```ts
const { id, action, html_deps } = envelope
if (html_deps && Array.isArray(html_deps)) {
  await this.renderDependencies(html_deps)
}
```

So this is not a regression — `ShinyChatEnvelope.html_deps` (line 66) is the real dep-delivery path and is unchanged. (The `message.html_deps` read in `markdown-stream-entry.ts` belongs to the standalone markdown-stream component, which has its own separate type.)

pkg-py/src/shinychat/_chat.py
```diff
@@ -875,63 +881,94 @@ async def _append_message_chunk(
         if is_tool_result(message) and message.request is not None:
             await self._hide_tool_request(message.request.id)  # type: ignore
 
+        # Detect the content type from the original message before normalization.
+        # HTML is a UserString (not a str subclass), so ChatMessage processes it
+        # into a plain str with <shinychat-raw-html> wrapping. For segment storage
+        # we want the raw HTML string and the "html" content type.
+        if isinstance(message, HTML):
+            chunk_content_type: ContentType = "html"
+            chunk_segment_content: str = str(message)
```
cpsievert: Would this is better off getting handled by `message_content_chunk()`?

pkg-py/src/shinychat/_chat.py
```diff
@@ -1298,8 +1337,12 @@ async def _transform_message(
...
             return None
 
-        res.content = content
-        return res
+        content_type: ContentType = "html" if isinstance(content, HTML) else "markdown"
+        seg = StoredContentSegment(content=str(content), content_type=content_type)
+        existing_deps = res.html_deps
+        if existing_deps:
+            seg["html_deps"] = existing_deps
+        return StoredMessage(role=res.role, segments=[seg])
```
cpsievert: Could this be made any simpler or easier to reason about?

PR #248, opened 2026-06-10
fix: add clipboard fallback for code copy button in embedded contexts

### LINKED ISSUE #247: Code copy button doesn't work in RStudio/Positron Viewer pane in 0.4.0
The code copy button silently fails when the chat UI is rendered inside the RStudio or Positron Viewer pane, but works fine in a standalone browser.

This appears to be a regression when upgrading from 0.3.0 -> 0.4.0.

To reproduce:

```{r}
library(ellmer)
chat <- chat_anthropic()
#> Using model = "claude-sonnet-4-5-20250929".
shinychat::chat_app(chat)
```

(and then ask some question which produces some code)

A quick Claude analysis reckons that it needs a fallback adding in the `catch()` block here:

https://github.com/posit-dev/shinychat/blob/6258846fadd45c7f82f1e9558c19b91df3f67077/js/src/markdown/components/CopyableCodeBlock.tsx#L16-L26

Originally reported in this [ellmer issue](https://github.com/tidyverse/ellmer/issues/1014).

thisisnic:
The "code copy" button silently fails in the RStudio and Positron Viewer panes since the 0.4.0 release                                                                     

This PR adds a `document.execCommand('copy')` fallback ([same pattern as RStudio's DataViewer.js](https://github.com/rstudio/rstudio/blob/f31f0b58c71f0fbceb280ff20b9a067b9d44f419/src/cpp/session/resources/grid/DataViewer.js#L3458-L3467)) so the button works in embedded contexts.

I tested it locally in Positron and RStudio (R only in both) and now works again there.  Didn't add unit tests here as there are no existing tests for that block, but can if needed - let me know!

Fixes #247


cpsievert:
Nice, thanks @thisisnic! If you could also update the news and change log, that would be very welcome.

PR #250, opened 2026-06-10
feat: file attachment support for chat messages

cpsievert:
## Why this matters

Users expect to be able to share images, documents, and code files in a chat — it's a standard interaction in ChatGPT, Claude, Gemini, and every other modern chat UI. Until now, shinychat only accepted plain text input, so developers who wanted multimodal input had to build their own file upload plumbing: drag-and-drop handling, MIME validation, base64 encoding, size limits, thumbnail previews, and conversion to the right LLM content types. This PR makes all of that built-in, so enabling file attachments is a one-liner.

## What it looks like

Users can attach files via three input methods: the **+** button, **drag-and-drop**, or **clipboard paste**.

### Staging area (before sending)

04-image-and-text-staged

### Sent message (in conversation)

Attachments appear as compact badges above the message text in the user bubble — image thumbnails stay visual, text files show a snippet.

05-image-conversation

### Python — with `client=` (zero config)

When a chatlas client is provided, attachments are enabled automatically — no extra code needed:

```python
from shiny.express import ui

ui.page_opts(fillable=True)

chat = Chat(id="chat", client=ChatAnthropic())
chat.ui()
```

### R — with `chat_mod_server()` (zero config)

```r
library(shinychat)
library(ellmer)

ui <- bslib::page_fillable(chat_mod_ui("chat"))

  chat <- chat_anthropic(system_prompt = "You are a helpful assistant.")
  chat_mod_server("chat", client = chat)
}

shinyApp(ui, server)
```

<details>
<summary>Manual wiring (Python)</summary>

For custom LLM integrations, enable the UI and accept a second parameter in the submit handler:

```python
from shinychat import Attachment, attachment_to_content

chat = Chat(id="chat")
chat.ui(allow_attachments=True)

@chat.on_user_submit
async def handle(user_input: str, attachments: list[Attachment]):
    contents = [attachment_to_content(a) for a in attachments]
    response = await my_client.stream_async(user_input, *contents)
```

</details>

<details>
<summary>Manual wiring (R)</summary>

```r
ui <- bslib::page_fillable(
  chat_ui("chat", allow_attachments = TRUE)
)

  chat <- ellmer::chat_openai()

    # When attachments are enabled, input is a list of Content objects —
    # splice with !!! to pass them as individual arguments
    stream <- chat$stream_async(!!!input$chat_user_input)
    chat_append("chat", stream)
  })
}

shinyApp(ui, server)
```

</details>

### R

- When the user attaches files, `input$<id>_user_input` changes from a plain character string to a list of ellmer `Content` objects. Splice either form to a chat method with `!!!` (e.g., `chat$stream_async(!!!input$<id>_user_input)`)
- `chat_mod_server()`'s `last_input` reactive mirrors this shape: string when attachments are disabled, list of `Content` objects when enabled.

### Python

  - `Chat.user_input()` no longer returns a plain string. It now returns `None` before the first submission, and a `UserInput` named tuple (`text`,
  `attachments`) after submission.
  - This is a breaking change for code that treated `chat.user_input()` as a `str`, or that assumed the previous return shape without a `None` check.

### Programmatic staging with `update_user_input()`

Server-side code can now pre-populate attachments in the chat input — useful for seeding context from a database, prior conversation, or a user action elsewhere in the app:

**Python**
```python
att = Attachment.from_path("report.pdf")
chat.update_user_input(
    value="Summarize this report",
    attachments=[att],
    submit=True,  # optional: auto-send immediately
)
```

**R**
```r
att <- chat_attachment("report.pdf")
update_chat_user_input(
  "chat",
  value = "Summarize this report",
  attachments = list(att),
  submit = TRUE
)
```

`attachment_mode` controls how new attachments combine with any already-staged files: `"append"` (default) adds to the set, `"set"` replaces it. Pass `attachments=[]` with `attachment_mode="set"` to clear.

## Supported file types

| Family | Types | Behavior |
|--------|-------|----------|
| **Images** | PNG, JPEG, GIF, WebP | Thumbnail preview; large images auto-downscaled to 1568px longest edge |
| **PDF** | PDF | Document badge with filename |
| **Text/code** | `.md`, `.txt`, `.csv`, `.json`, `.py`, `.js`, `.ts`, `.r`, `.yaml`, `.toml`, `.qmd`, `.rmd`, `.sql`, `.sh`, `.html`, `.css`, `.xml`, and more | Content preview card showing first few lines of the file |

This is the initial supported set — it can grow as [chatlas](https://github.com/posit-dev/chatlas) and [ellmer](https://github.com/tidyverse/ellmer) add support for additional file types.

Max combined size defaults to **30 MB**, configurable via the `SHINYCHAT_MAX_ATTACHMENT_SIZE` environment variable.

## Notable behaviors

- **Image downscaling**: Images larger than 1568px on their longest edge are automatically downscaled client-side before sending, matching Anthropic's recommended input size.
- **Long paste → attachment**: Clipboard pastes over 1000 characters are automatically converted into a "Pasted Text" file attachment instead of flooding the textarea.
- **Animated GIF → still**: GIFs are converted to a still frame (with a notice) since most LLM APIs don't support animation.
- **Bookmarking**: Attachment state is included in bookmark/restore — user messages with attachments round-trip correctly.

## Reviewer notes

The implementation is layered across three commits, one per package:

1. **`js/`** — Attachment staging UI (React), file processing pipeline (downscaling, MIME resolution, size validation), and the `setInputValue` extension for programmatic staging. All client-side; no new dependencies.
2. **`pkg-py/`** — `Attachment` Pydantic model with `from_path()`/`from_data()`/`from_url()` constructors, `attachment_to_content()` converter to chatlas types, `InputHandler` for automatic argument injection on `@on_user_submit`, and `allow_attachments` parameter on `chat_ui()`.
3. **`pkg-r/`** — `chat_attachment()` constructor, `resolve_attachment_attrs()` for MIME validation, `allow_attachments` on `chat_ui()` / `chat_app()`, and server-side deserialization into ellmer `Content` objects.

The canonical list of supported MIME types lives in `js/src/chat/attachment-types.json` and is copied to both packages at build time — Python and R read it at load time to stay in sync with the client.


pkg-r/R/attachments.R
```diff
@@ -0,0 +1,197 @@
...
+#' @param name Filename shown in the attachment chip. Defaults to
+#'   `basename(path)`.
+#'
+#' @returns A list with elements `mime`, `name`, `size`, and `data_url`,
+#'   ready to pass as an element of the `attachments` argument of
+#'   [update_chat_user_input()].
+#'
+#' @export
+chat_attachment <- function(path, mime = NULL, name = NULL) {
+  path <- normalizePath(as.character(path), mustWork = TRUE)
```
gadenbuie, line 214: `normalizePath()` vectorizes, so passing `chat_attachment(c("a.png", "b.png"))` silently processes only the first file (`readBin()` takes a scalar). An `rlang::check_string(path)` before this line would catch it early.

js/src/chat/AttachmentLightbox.tsx
```diff
@@ -0,0 +1,127 @@
...
+import {
+  attachmentFamily,
+  dataUrlToBlob,
+  decodeTextDataUrl,
+} from "./attachments"
+
+const FOCUSABLE_SELECTOR =
+  'button, [tabindex="0"], a[href], input, select, textarea, iframe'
+
+export function AttachmentLightbox({
```
gadenbuie, line 19: Two accessibility gaps in the lightbox:

1. **Focus restoration:** When the lightbox closes, focus goes to `<body>` — keyboard users lose their place. Capture `document.activeElement` on mount and restore it in cleanup.
2. **Scroll lock:** Background page remains scrollable while the `aria-modal="true"` overlay is open. Consider setting `document.body.style.overflow = "hidden"` on mount and restoring on cleanup.

pkg-r/R/attachments.R
```diff
@@ -0,0 +1,197 @@
...
+# Resolve the max combined attachment size (bytes) from the env var or default.
+resolve_max_attachment_size <- function(default = 30L * 1024L * 1024L) {
+  env <- Sys.getenv("SHINYCHAT_MAX_ATTACHMENT_SIZE", unset = "")
+  size <- suppressWarnings(as.numeric(if (nzchar(env)) env else default))
+  if (is.na(size)) {
+    cli::cli_abort(
+      "{.envvar SHINYCHAT_MAX_ATTACHMENT_SIZE} must be a number of bytes, got {.val {env}}."
+    )
+  }
+  size
```
gadenbuie, line 49: Python's equivalent rejects negative values. R silently passes `-1` to the HTML attribute, which the JS parses as valid and rejects all attachments with a confusing UX. Consider adding a `size < 0` guard here to match Python's behavior.

pkg-py/src/shinychat/_attachments.py
```diff
@@ -0,0 +1,253 @@
...
+            f"{att.mime} attachments, but got a remote URL. "
+            f"Use Attachment.from_path() or Attachment.from_data() instead."
+        )
+
+
+def decode_data_url(data_url: str) -> bytes:
+    comma = data_url.find(",")
+    if comma == -1:
+        raise ValueError("Malformed data URL")
+    return base64.b64decode(data_url[comma + 1 :])
```
gadenbuie: `base64.b64decode()` raises `binascii.Error` with a cryptic message on malformed input. Since this is reachable from the public `attachment_to_content()` API, consider wrapping it:
```python
try:
    return base64.b64decode(data_url[comma + 1:])
except Exception as e:
    raise ValueError("Malformed base64 payload in data URL") from e
```

pkg-r/R/attachments.R
```diff
@@ -0,0 +1,197 @@
...
+
+attachment_types <- local({
+  cache <- NULL
+  function() {
+    if (is.null(cache)) {
+      path <- system.file(
+        "lib/shiny/attachment-types.json",
+        package = "shinychat"
+      )
+      types <- jsonlite::fromJSON(path, simplifyVector = TRUE)
```
gadenbuie, line 13: If `system.file()` returns `""` (source install without the JS build/copy step), this throws a cryptic `"premature EOF"` error. A guard like `if (!nzchar(path)) cli::cli_abort("attachment-types.json not found in shinychat installation.")` would save someone a confusing debugging session.

gadenbuie (commented):
Nice work on this — the architecture is solid (canonical MIME types in JS, mirrored to both packages), the `on_user_submit` arity dispatch is well-designed for backward compatibility, and the test coverage is impressively thorough across all three packages. A few questions and suggestions below.

~**Server-side size enforcement:** The max attachment size is currently enforced only in JavaScript. A crafted WebSocket message could bypass the JS check and send large base64 payloads directly. Is this an acceptable risk given Shiny's existing WebSocket/httpuv message size limits, or should the server also validate payload size against `resolve_max_attachment_size()`?~ 

See 2667137b6373df76067c7caccf09075aa4143d1f

~**Server-side MIME validation:** `Attachment.mime` accepts any string — the JS accept list is the only guard. A crafted message with an unsupported MIME type passes Pydantic validation and reaches `attachment_to_content()`, which raises `ValueError`. In the manual `on_user_submit` path, the invalid MIME propagates to user code unchecked. Worth considering whether to validate at the model level, the input handler, or leave it to callers — depends on whether custom handlers should be able to process non-standard types.~

See 2667137b6373df76067c7caccf09075aa4143d1f

~**`user_input()` return type:** This changed from `str | None` to `tuple[str, list[Attachment]]`, and the "no input yet" sentinel changed from `None` to `("", [])` (always truthy). The `on_user_submit` dispatch handles this gracefully, but direct callers of `user_input()` will break. I think it's worth considering a `NamedTuple` or small dataclass instead of a bare tuple — that would leave room for non-breaking future expansions, like the one we're handling now, while still supporting destructuring.~ 

See https://github.com/posit-dev/shinychat/commit/5bc238ad086aff5f8ed1d63e7ee76be9903a786a

~**Default max size:** The constant is `30 * 1024 * 1024` (30 MiB) but the docstrings say "defaults to 30,000,000" (30 MB). Minor — the docs could just say "approximately 30 MB" to avoid the mebibyte/megabyte mismatch.~ 

See https://github.com/posit-dev/shinychat/commit/5bc238ad086aff5f8ed1d63e7ee76be9903a786a

~**`attachmentRefs.current` is never trimmed** (`ChatInput.tsx:205`). When attachments are removed, stale DOM refs remain in the array. Minor — consider resetting `attachmentRefs.current.length = attachments.length` when attachments change.~ 

See https://github.com/posit-dev/shinychat/commit/5bc238ad086aff5f8ed1d63e7ee76be9903a786a

pkg-r/R/attachments.R
gadenbuie: Sorry, this one slipped through the cracks. I think this is totally fine. It would be nice for these lines to be covered by tests just in case we move or rename the json file.

pkg-r/R/attachments.R
cpsievert: Addressed in 5bc238ad

pkg-py/src/shinychat/_attachments.py
cpsievert: Addressed in 5bc238a

pkg-r/R/attachments.R
cpsievert: Addressed in 5bc238a

pkg-r/R/attachments.R
cpsievert: Addressed in b8f07d06

js/src/chat/AttachmentLightbox.tsx
cpsievert: Addressed in b8f07d06

pkg-r/R/chat_app.R
```diff
@@ -254,7 +269,7 @@ chat_mod_server <- function(
   append_stream_task <- shiny::ExtendedTask$new(
     function(client, ui_id, user_input, controller = NULL) {
       stream <- client$stream_async(
-        user_input,
+        !!!user_input,
         stream = "content",
         controller = controller
       )
```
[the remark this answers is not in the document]
cpsievert: https://github.com/posit-dev/shinychat/commit/1d87b292

pkg-r/R/attachments.R
```diff
@@ -0,0 +1,252 @@
...
+decode_data_url_text <- function(data_url) {
+  comma <- regexpr(",", data_url, fixed = TRUE)
+  if (comma == -1L) {
+    cli::cli_abort("Malformed data URL")
+  }
+  b64 <- substring(data_url, comma + 1L)
+  bytes <- jsonlite::base64_dec(b64)
+  bytes <- bytes[bytes != as.raw(0L)]
+  iconv(rawToChar(bytes), from = "UTF-8", to = "UTF-8", sub = "\uFFFD")
+}
```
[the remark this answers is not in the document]
cpsievert: https://github.com/posit-dev/shinychat/commit/1d87b292

pkg-r/NEWS.md
```diff
@@ -1,5 +1,18 @@
 # shinychat (development version)
 
+## Breaking changes
+
+* `input$<id>_user_input` is unchanged for text-only messages (still the typed
+  string). When the user attaches files, it is now a list of ellmer `Content`
+  objects (the typed text, if any, followed by one object per attachment), and
+  the separate `input$<id>_user_attachments` input has been removed. Forward
+  either form to a chat method by splicing with `!!!`, e.g.
+  `chat$stream_async(!!!input$<id>_user_input)`.
```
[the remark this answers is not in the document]
cpsievert: https://github.com/posit-dev/shinychat/commit/1d87b292

js/src/chat/state.ts
```diff
@@ -33,6 +34,8 @@ export interface ChatMessageData {
   /** True for the empty placeholder message shown while waiting for the assistant to respond. */
   isPlaceholder?: boolean
   icon?: string
+  /** Attachments the user sent (user messages only). */
+  attachments?: AttachmentPayload[]
```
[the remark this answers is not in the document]
cpsievert: https://github.com/posit-dev/shinychat/commit/1d87b292

cpsievert:
@gadenbuie looking to do another round of review? Or am I good to merge?

gadenbuie (approved):
LGTM! Nice work 🏄‍♂️

PR #239, opened 2026-05-28
Add slash commands: typeahead palette, command chips, and server-side handlers

cpsievert:
### What this enables

Slash commands let app authors register named shortcuts (`/clear`, `/help`, `/search`, …) that users trigger right from the chat input. Type `/` and a palette appears; pick a command, optionally type some text after it, and hit Enter. The command can expand into a prompt sent to the LLM, trigger a side-effect (clear chat, open a modal, export transcript), or be handled entirely client-side — all with a clean, discoverable affordance that separates "do a thing" from "talk to the model."

The chat input is now a [TipTap](https://tiptap.dev/) (ProseMirror) rich-text editor. The wire format is unchanged — it serializes to the same plain text the old `<textarea>` produced — but slash commands now render as styled inline chips in both the input and the message history.

---

### UX

**Input:**
- Type `/` → palette opens, filters as you type. Arrow keys, Tab, Enter to pick; Esc to dismiss.
- Selecting a command inserts an atomic chip. Backspace restores it to `/name` text and re-opens the palette.
- Pasting text that starts with a recognized `/command` auto-converts to a chip.
- `setInputValue()` from the server produces chips for recognized commands (bookmark/restore path).
- Unrecognized `/words` fall through as ordinary text — `/` is never captured.

**Message history:**
- Sent messages with a leading slash command render a `CommandChip` inline in the first paragraph, backed by the registered command list.

**`submit_key` option:**
- New parameter on `chat_ui()` in both Python and R: `"enter"` (default, Enter submits) or `"enter+modifier"` (Ctrl/Cmd+Enter submits, plain Enter inserts a line break).
- The input stays editable while a response is streaming — only submission is blocked, not typing.

**Chip styling:**
- `--shiny-chat-command-chip-color` (anchor color, default `--bs-purple`), with derived `--shiny-chat-command-chip-bg`, `-border`, `-border-color` via `color-mix()` and `light-dark()` for automatic dark mode support.

---

### API

**Python** — `slash_command` is a method on `Chat`, usable as a decorator or called directly:

```python
chat = Chat("chat")

@chat.slash_command("clear", "Clear the conversation")
async def _():
    await chat.clear_messages()

# One-argument handlers receive the text typed after the command name.
# For "/search shiny modules", user_input == "shiny modules".
@chat.slash_command("search", "Search the docs")
async def _(user_input: str):
    response = await client.stream_async(f"Search the docs for: {user_input}")
    await chat.append_message_stream(response)

[...]

PR #251, opened 2026-06-11
Replace textarea with TipTap editor and add command chip styling

gadenbuie:
Stacks on #239.

### Slash command menu

dyn-1e2d609a962c987498fb3dcf51a6d4b9

### Partial completion

dyn-238b0ae0e240c1c4cf790b0e516d2c99

### Chips in messages

file-ecbe776ff511e02c200805e2433ebafb

## Summary

Replaces the plain `<textarea>` chat input with a TipTap (ProseMirror) rich-text editor. The wire format is unchanged — TipTap serializes to the same plain text the textarea produced — but slash commands now render as styled chips in both the input and the message history.

**Input editor (TipTap):**

- Selecting a command from the palette inserts an atomic inline chip (not editable as text). Backspace restores it to `/name` text and re-opens the palette.
- Pasting text that starts with a recognized `/command` auto-converts the command portion to a chip.
- `setInputValue()` from the server also produces chips for recognized commands (bookmark/restore path).
- Input history recall (arrow up/down) works as before.

**Sent message chips:**

- `ChatMessage` detects a leading `/command` in user messages (validated against the registered command list via `SlashCommandsContext`) and renders a `CommandChip` inline inside the first paragraph.
- Unrecognized `/words` render as plain text.

**`submit_key` option:**

- New `submit_key` parameter on `chat_ui()` in both Python and R: `"enter"` (default, Enter submits) or `"enter+modifier"` (Ctrl/Cmd+Enter submits, plain Enter adds a newline).

**Chip CSS custom properties:**

- `--shiny-chat-command-chip-color` (anchor color, default `--bs-primary`), with derived `--shiny-chat-command-chip-bg`, `-border`, `-border-color` via `color-mix()` and `light-dark()` for automatic dark mode support.

**Other fixes included:**

- Input stays editable while a response streams (only submission is blocked, not typing)
- Palette positioning fixes (double-rAF for measurement, removed vestigial absolute positioning)
- Placeholder support via lightweight ProseMirror plugin (avoids `@tiptap/extension-placeholder` which crashes in JSDOM)
- Handler parameter validation in Python (`ValueError` for handlers with >1 parameter)

## Verification

Use the example apps from #239 (`example_slash_commands.py` / `example-slash-commands/app.R` in the PR description). With either app running:

1. Type `/` — palette appears anchored to the input
2. Select a command — it becomes a styled purple chip in the input
3. Press Backspace on the chip — it restores to `/name` text with the palette open
4. Submit a command — the chip appears inline in the sent message
5. Resize/scroll — palette repositions correctly
6. Arrow up — recalls previous input with chip intact


cpsievert:
Thanks!

PR #263, opened 2026-06-24
refactor: rename chat_greeting() dismissible to persistent

### LINKED ISSUE #260: Rename chat_greeting's `dismissible` to `persistent`: the current name implies a close button that doesn't exist
## Problem

The `chat_greeting()` function (Python) and `chat_greeting()` constructor (R) expose a `dismissible` parameter that controls whether the greeting stays visible after the user sends a message. The default is `dismissible=True` (greeting hides on first activity); setting `dismissible=False` keeps it pinned.

The name is misleading for one specific reason: **there is no user-initiated dismissal in the component**. There is no close button, no X affordance, no `aria-dismissible` role — nothing a user can click to close the greeting themselves. Dismissal is entirely automatic, driven by three internal events (`INPUT_SENT`, `message`, `chunk_start`).

"Dismissible" is a standard UI/UX term meaning *"the user can close this"* (dismissible banners, dismissible toasts, etc.). A developer reading `dismissible=False` would reasonably think they are suppressing a close button — not that they are changing the greeting's entire persistence behavior. The Python docstring even hedges around this: *"Whether the greeting can be dismissed when the user sends a message"* — but the user doesn't dismiss it; the system does.

## Proposed fix

Rename the parameter to `persistent`:

```python
# before
chat_greeting("Welcome!", dismissible=False)  # ambiguous — suppressing what button?

# after
chat_greeting("Welcome!", persistent=True)    # unambiguous — stays visible
```

`persistent` is the right word: it's the antonym of "ephemeral/temporary," it doesn't imply any user action, and `persistent=False` (the new default) correctly expresses "this greeting goes away once conversation starts."

## Impact

- `chat_greeting(dismissible=False)` → `chat_greeting(persistent=True)`
- Default behavior unchanged (`persistent=False` == current `dismissible=True`)
- A deprecation shim on `dismissible` would cover existing code cleanly

## Scope

The rename touches:
- `pkg-py/src/shinychat/_chat_types.py` — `ChatGreeting` class + `chat_greeting()` function signature
- `pkg-r/R/chat.R` — R-side `chat_greeting()` equivalent
- `js/src/transport/types.ts` — `GreetingOptions.dismissible`
- `js/src/chat/state.ts` — internal state logic referencing `options.dismissible`
- Tests, docs, and any examples using the parameter

gadenbuie:
Closes #260

## Summary

Renames the `dismissible` parameter on `chat_greeting()` to `persistent` across the JS client, Python package, and R package. The old name implied a user-controlled close button that never existed — dismissal is entirely automatic. `persistent` is unambiguous: `persistent = FALSE` (the default) means the greeting hides on first activity; `persistent = TRUE` keeps it visible throughout the conversation.

**Polarity is inverted from the old default:** `dismissible = TRUE` (hide) → `persistent = FALSE` (hide); `dismissible = FALSE` (stay) → `persistent = TRUE` (stay).

Both R and Python include deprecation shims so existing code using `dismissible` warns rather than breaks. In R, `lifecycle::deprecate_warn("0.4.1", ...)` is used with `lifecycle::deprecated()` as the sentinel default; in Python, the project's existing `DEPRECATED`/`DEPRECATED_TYPE` sentinel pattern is used. Both `persistent` and `dismissible` are placed after `...`/`*` so they must be passed by name — positional calls will error with a clear message.

The wire format key sent from R/Python to the JS component also changes from `"dismissible"` to `"persistent"`.

## Verification

```r
library(shinychat)

# Default: greeting hides after first message (unchanged behavior)
chat_greeting("## Welcome!")

# Persistent: stays visible throughout
chat_greeting("Please read our terms.", persistent = TRUE)

# Old code: warns and maps correctly
chat_greeting("Please read our terms.", dismissible = FALSE)
#> Warning: `chat_greeting(dismissible)` was deprecated in shinychat 0.4.1.
#> Use `chat_greeting(persistent)` instead.
```


cpsievert (approved):

gadenbuie:
FYI I incorrectly had shinychat throw if both `dismissible` and `persistent` were included; I added 0e6c1e00c264720c6358a6d1b67a918f0be721ba to main to revert that and to only warn about `dismissible` being deprecated when `persistent` is not included.

PR #264, opened 2026-06-25
feat!(pkg-r): add `chat_server()`, deprecate `chat_mod_server()` and `chat_mod_ui()`

cpsievert:
## Summary

`chat_server()` replaces `chat_mod_server()` as the primary way to wire up server-side chat logic.

The original module-based API (`chat_mod_ui()` / `chat_mod_server()`) was more structure than the use case needs. The goal of this API is simple: you hand us an ellmer client and we hook up all the server logic most people want — streaming, cancellation, bookmarking, etc. A full `moduleServer()` scope isn't required to deliver that. Furthermore, since part of the ID namespace is private/implicit, downstream logic that depends on the chat's ID (e.g., slash command JS events) need workarounds to work in the way users might expect. This change also better aligns the R package with the Python API where no modules are needed to get the "batteries included" experience. 

`chat_server()` does the same job but runs directly in the caller's session scope. If you're already inside a `moduleServer()`, you pass that session in — no extra nesting, no doubled namespaces.

## What changed

- **`chat_server(id, client, ..., session)`** — new exported function with the same logic as `chat_mod_server()`, but runs in the provided session scope rather than creating its own. Takes an explicit `session` parameter (defaults to `getDefaultReactiveDomain()`).
- **`chat_mod_server()`** — soft-deprecated (`0.5.0`). Now a thin wrapper around `chat_server()`.
- **`chat_mod_ui()`** — soft-deprecated (`0.5.0`). Replace with `chat_ui(NS(id, "chat"), ...)` in your module UI.

## Migration

The simplest case — no module needed at all:

```r
# Before (forced you into a module)
ui <- fluidPage(chat_mod_ui("chat"))
  chat_mod_server("chat", client)
}

# After
ui <- fluidPage(chat_ui("chat"))
  chat_server("chat", client)
}
```

If you're embedding chat inside an existing module, pass that module's `session` explicitly:

```r
ui <- function(id) {
  ns <- NS(id)
  chat_ui(ns("chat"))
}
server <- function(id, client) {
  shiny::moduleServer(id, function(input, output, session) {
    chat_server("chat", client, session = session)
  })
}
```

Existing code using `chat_mod_server()` continues to work with a deprecation notice.


pkg-r/R/chat_app.R
```diff
@@ -293,323 +301,352 @@ chat_mod_server <- function(
...
+  })
+  makeActiveBinding("client", function() client, ret)
+  ret$append <- chat_append_mod
+  ret$update_user_input <- chat_update_user_input
+  ret$clear <- client_clear
+  ret$set_greeting <- set_greeting_mod
+  ret$set_client <- set_client
+  ret$slash_command <- slash_command_method
+  lockEnvironment(ret)
+  ret
```
cpsievert, line 651: I'm tempted to say we should also make this a more official R6 object (or similar) so that we can better document what's here, but maybe there's a good reason why we don't do that?

pkg-r/R/chat_app.R
```diff
@@ -73,7 +72,7 @@
 #' @returns
 #'   * `chat_app()` returns a [shiny::shinyApp()] object.
 #'   * `chat_mod_ui()` returns the UI for a shinychat module.
```
gadenbuie: Are we keeping `chat_mod_ui()`?

pkg-r/R/chat_app.R
gadenbuie: This just grew organically, so there isn't a technical reason why we aren't using an R6 or similar object here.

The fields are all documented in `?chat_mod_server` or now `?chat_server` and if anything I'd recommend that we separate `?chat_app` from `?chat_server` since `chat_server` will stand on its own. Beyond that, I'm ambivalent about creating an "official" R6 object because I don't think it really improves the documentation from a user's perspective. We'd be exporting a class that users shouldn't create on their own and they'd have to go from `?chat_server` docs to the new class docs to find, essentially, the same listing of methods and their signatures/definitions.

So while I've also wanted to refactor this code every time I've looked at it, I haven't been convinced that it's worth the overhead yet.

pkg-r/R/chat_app.R
cpsievert: Nah, I was mostly just keeping the diff here minimal for a 1st pass review. [a6ca1ac](https://github.com/posit-dev/shinychat/pull/264/commits/a6ca1ac885561f580b9470c93f9e41a4b03d9f43) It goes further to remove internal usage, update documentation, etc.

gadenbuie (approved):
Looks good, thanks!

PR #266, opened 2026-06-25
feat: multi-conversation history — drawer UI, pluggable stores, and restore modes (Python + R)

cpsievert:
This PR makes so apps that use ellmer/chatlas now get multi-conversation history automatically. More precisely, if you pass a `client=` to `Chat()` in Python, or supply a client to `chat_server()` in R, the history drawer appears without any extra wiring — a toggle icon opens a panel showing all past conversations, a New Chat button, LLM-generated titles, search, rename, and delete. Conversations are persisted per-user and survive redeployments on Posit Connect.

To opt out: pass `history=False` (Python) or `history=FALSE` (R). For apps that can't use the module pattern, `chat_enable_history()` (R) or `chat.history.enable()` (Python) lets you wire history manually.

### Restore mode

The most important option to know about is `restore_mode`, which controls how the active conversation is remembered across page reloads:

- **`"browser"` (default):** the active conversation ID is kept in `localStorage`. The URL doesn't change, but reloading the page returns the user to where they left off.
- **`"url"`:** the active conversation ID is stored as a plain `?shinychat_conversation_id=<id>` query parameter. The address bar updates in place (via `history.replaceState`) on each conversation switch — no page reload, no server bookmark configuration required. Use this if you want the URL to reflect the current conversation so users can share or bookmark a specific conversation.
- **`"bookmark"`:** fully participates in Shiny server bookmarking. After every LLM response, a fresh server bookmark is minted and the address bar updates to `?_state_id_=...`. On in-session conversation switches the controller navigates to the target conversation's bookmark URL (if one exists). Requires `bookmark_store="server"` in the Shiny app. Use this when the app already uses Shiny bookmarks to capture full input state alongside the chat — the conversation and the bookmark state stay in lockstep.
- **`"none"`:** no cross-session restore; every page load starts fresh.

One important nuance: switching conversations — whether triggered by a page load or a mid-session drawer click — restores the transcript and any per-conversation values captured by `on_save`, but does **not** automatically reset raw Shiny input controls like sliders or text boxes. Use `on_restore` callbacks to sync those explicitly.

**Passive Shiny bookmark integration.** Regardless of `restore_mode`, if the app has `bookmark_store="server"` configured, the history system automatically stamps the active conversation ID into any Shiny bookmark the user triggers. Reloading from such a bookmark reopens the correct conversation, even if `restore_mode` is `"browser"` or `"none"`. This is distinct from `restore_mode="bookmark"` above, which *actively mints* bookmarks — this passive integration just ensures that developer- or user-triggered bookmarks capture the current conversation.

### Storage and identity

History is stored server-side, scoped by user (or a custom scope), stored in memory (in devmode) or the filesystem. On Posit Connect the correct persistent directory is picked up automatically. You can configure where and how conversations are stored — in memory, the filesystem, or with a completely custom `ConversationStore` backend — as well as how users are identified and how conversation titles are generated.

**Scope / user identity.** On authenticated deployments (Posit Connect with auth enabled), `session.user` is used as the scope key, so each user sees only their own conversations. For unauthenticated sessions — local dev, public deployments — there is no server-side identity, so the history system mints a random token and stores it in the browser's `localStorage`. That token becomes the scope key: the same browser gets the same history across reloads, but a different browser (or a cleared `localStorage`) starts fresh and cannot see the previous conversations. The scope can also be set to a custom string or callable via `HistoryOptions(scope=...)` / `history_options(scope = ...)` — for example to share history across a group, or to hard-code a fixed scope during development.

**Storage quota.** Each user's history is capped at 100 MB by default (`max_store_mb=100` in `HistoryOptions` / `history_options()`). After every save, if the total size for that scope exceeds the cap, the oldest conversations (by `updated_at`) are evicted one at a time until usage falls back under the limit. The active conversation is never evicted mid-session. The cap can be raised, lowered, or removed entirely (`max_store_mb=None` in Python / `max_store_mb = NULL` in R) via the options object.

### Saving app state per conversation (advanced)

For apps where other parts of the UI are tightly coupled to the chat — a selected data source, a filter panel, a chosen model — you can attach extra state to each conversation that travels with it on save and restore. This is a niche use case: most apps won't need it. But for the ones that do, it works even if you're not using full Shiny bookmarking:

**Python:**
```python
@chat.history.on_save
def _(values):
    values["tab"] = input.tab()

@chat.history.on_restore
def _(values):
    ui.update_navs("tabs", selected=values.get("tab"))
```

**R**:
```r
mod <- chat_mod_server("chat", client = llm)

mod$history$on_save(\(values) { values$tab <- input$tab; values })
mod$history$on_restore(\(values) updateTabsetPanel(session, "tabs", selected = values$tab))
```

This is independent of Shiny's bookmark system — apps that wire `onBookmark` / `on_bookmark` handlers for their own purposes are not automatically bridged into per-conversation history state.

---

## Reviewer guide: architectural decisions

The history feature is built on a small number of load-bearing decisions. Understanding why each choice was made is the fastest way to build an accurate mental model of how the pieces fit together.

### 1. In-session swap, not URL-per-conversation

The most consequential design decision is that switching conversations is **not** a page reload. When the user clicks a conversation in the drawer, the controller:

1. Fetches the target `ConversationRecord` from the store
2. Replaces the LLM client's turn history via `adapter.set_turns_json()`
3. Runs `on_restore` callbacks with the conversation's saved `values` dict
4. Sends a `history_navigate` transport message — the JS replaces the rendered message list in place

The alternative — encoding each conversation as a bookmarked URL and navigating to it on switch — was rejected because every switch would require a full page reload, bookmarks aren't enumerable per user so you can't list them in a drawer, and URL-store apps and Shinylive would be out of reach entirely. The in-session swap is what makes a responsive drawer UI possible, but it comes with trade-offs that surface repeatedly in the implementation: app state restoration is partial (no input control reset mid-session), the JS `clear` action must be overridden to carry history state forward rather than wiping it, and a phantom-save guard is needed to prevent replaying a transcript from triggering a spurious write (see below).

### 2. The conversation record: a linear chain inside a tree schema

Each conversation is stored as a `ConversationRecord` — a tree of `ConversationNode` objects where every node holds one turn and a reference to its parent. Today all conversations are strictly linear: each new turn appends to `current_leaf`, and `path_node_ids()` (root → `current_leaf`) is the canonical turn sequence.

The tree shape exists solely to support future message editing without a schema migration: a re-sent message would create a sibling node branching off the parent rather than rewriting history. That machinery is currently inert — there's no sibling-creation code path, no branch-selection UI — but the schema is already capable of representing it.

For the purposes of code review and maintenance: treat the record as a list. Any code that walks `path_node_ids()` is just iterating turns in order. The `nodes` dict and parent-pointer fields are there for the future.

### 3. Pluggable store with Connect-aware default

`ConversationStore` is a five-method interface: `list()` (metadata only, no turns), `get()` (full record), `put()` (upsert), `delete()` (remove by id), and `total_size()` (bytes used by scope). All conversation I/O flows through this boundary, which is what makes swapping backends possible. The store is intentionally kept narrow — it knows nothing about users, sessions, or Shiny; it just reads and writes records keyed by `(scope, conversation_id)`.

The Connect persistence story is worth understanding explicitly. App working directories on Posit Connect are wiped on every redeploy, so writing to a path relative to the app is not safe. The default `FileConversationStore` uses `resolve_history_dir()` to discover a safe location at runtime, checking in priority order: `CONNECT_CONTENT_DATA_DIR` (an Early Access env var) → Shiny's bookmark save directory piggybacked via a reserved ID (`shinychat-conversations`) → `.shinychat/conversations` relative to the working directory as a local fallback. The bookmark save directory route is what makes this redeploy-safe on current Connect without the Early Access flag; the piggybacking via a reserved bookmark ID is a deliberate hack with a comment pointing to it.

### 4. Per-conversation app state via `on_save` / `on_restore`

When a conversation is saved, the controller invokes all callbacks registered with `@chat.history.on_save` (Python) or `mod$history$on_save()` (R), which write into a `values` dict that is stored alongside the turns. On restore, the corresponding `on_restore` callbacks receive that dict back. "Restore" means both page-load restore (when `restore_mode` is `"browser"` or `"url"` and a prior conversation is found) and in-session conversation switches — the same callbacks fire in both cases.

This is the public API for keeping auxiliary UI state — a selected model, an active tab, a filter panel — in sync with the active conversation. Raw Shiny input values are not synced automatically; `on_restore` is where you call `ui.update_*()` / `updateXxx()` to apply them.

The only remaining private Shiny API access is in `_history_store.py` (`shiny.bookmark._global.get_bookmark_save_dir_fn`) used by `resolve_history_dir()` to locate a redeploy-safe persistence path on Posit Connect. If Shiny ever ships a public equivalent, that is the only file that needs to change.

### 5. `restore_mode` covers two separate concerns

`restore_mode` looks like a single dial but it actually governs two independent mechanisms:

**Cross-reload persistence** — which conversation to reopen when the page loads fresh. `"browser"` stores the active conversation ID in `localStorage`; `"url"` stores it as a `?shinychat_conversation_id=<id>` query parameter and updates the address bar via `history.replaceState` on each response; `"bookmark"` actively mints a server bookmark after every response, keeping `?_state_id_=...` in sync and enabling full Shiny bookmark restore on page reload (requires `bookmark_store="server"`); `"none"` skips cross-reload tracking entirely.

**Per-conversation app state** — the `values` dict captured by `on_save`/`on_restore`. This always runs when a conversation is loaded, regardless of `restore_mode`. It is not controlled by `restore_mode` at all.

None of these modes automatically reset raw Shiny input controls. Apps that want input values to follow the active conversation must do so explicitly via `on_restore` — including `"bookmark"` mode, which handles bookmark *persistence* but not input *restoration*.

### 6. The phantom-save guard

Replaying a transcript on conversation switch works by calling `chat.append_message()` for each stored UI message. That's the same path a live response takes, and the same "on response complete" reactive fires. Without intervention, every conversation switch would immediately write a redundant save back to the store — and the diff between "turns I just loaded" and "turns as persisted" is empty, so the save would be harmless but wasteful and could mask real bugs.

The fix is `_suppress_next_save` (Python) and its R equivalent: a flag set just before replay begins that causes the save-on-response effect to swallow exactly one save. It self-heals: if the flag is incorrectly set and a genuine save is suppressed, the next response will re-diff and catch up correctly. This is one of the subtler pieces of state in the controller and the most likely place to look if you see a conversation switch that appears to save stale content.


pkg-r/R/chat_app.R
```diff
@@ -333,15 +368,39 @@ chat_server <- function(
...
+        restore_ui = FALSE,
+        session = session
+      )
+    } else {
+      NULL
+    }
+
+    new_ctrl <- get_session_chat_bookmark_info(
+      session,
+      "chat.history-controller"
```
gadenbuie: **[Blocking] Hardcoded controller key disables history for any non-`"chat"` module id.**

`chat_enable_history(id, ...)` registers the controller under `paste0(id, ".history-controller")` (`chat_history.R:500`), but this read (and the ones at lines 715 and 724) hardcode `"chat.history-controller"`. For any `chat_server(id != "chat")`, `set_client()` re-registration and the `on_save`/`on_restore` wiring silently get `NULL` and no-op.

Fix: use `paste0(id, ".history-controller")` at all three sites, and add a regression test with a non-default module id — the absence of one is why this slipped through.

pkg-py/src/shinychat/_history_types.py
```diff
@@ -0,0 +1,107 @@
...
+        visited: set[str] = set()
+        cursor = self.current_leaf
+        while cursor is not None:
+            if cursor in visited:
+                raise ValueError(
+                    f"Cycle detected in conversation nodes at {cursor!r}"
+                )
+            visited.add(cursor)
+            ids.append(cursor)
+            cursor = self.nodes[cursor].parent
```
gadenbuie: **[Required] `path_node_ids()` raises a bare `KeyError` on a dangling parent pointer.**

The cycle guard protects against repeats but not against a `parent` id missing from `nodes` (truncated/partial/hand-edited JSON). One bad record crashes `switch_to`/restore with `KeyError` instead of the intended "no longer exists" path. The R equivalent (`record_path_node_ids`) already guards this.

```python
node = self.nodes.get(cursor)
if node is None:
    raise ValueError(f"Dangling parent reference at {cursor!r}")
ids.append(cursor)
cursor = node.parent
```

pkg-py/src/shinychat/_history_store.py
```diff
@@ -0,0 +1,234 @@
...
+                    )
+                except Exception as e:
+                    logger.warning("Unreadable conversation %s: %s", f.name, e)
+                    continue
+                metas.append(rec.meta)
+            metas.sort(key=lambda m: m.updated_at, reverse=True)
+        self._meta_cache[scope] = metas
+        return list(metas)
+
+    async def get(self, scope: str, conv_id: str) -> ConversationRecord | None:
```
gadenbuie: **[Required] `list()`/`get()` cache divergence surfaces as a user-facing error.**

`list()` intentionally serves stale metadata from `_meta_cache`; `get()` always hits disk. In a no-sticky-session / multi-worker deployment, a conversation deleted underneath still shows in the drawer; clicking it makes `switch_to` raise `RuntimeError("Conversation ... no longer exists.")` (`_history.py:367`), shown as an error toast.

The cache design is fine — the handling isn't. On `get()` returning `None` for an id the cache believes exists, invalidate that scope's cache entry and refresh the list rather than erroring.

pkg-py/src/shinychat/_history.py
```diff
@@ -0,0 +1,878 @@
...
+        if self._title_task is not None and not self._title_task.done():
+            self._title_task.cancel()
+
+    async def _evict_one(self, conv_id: str) -> None:
+        assert self.scope is not None
+        if self.on_evict is not None:
+            await self.on_evict(conv_id)
+        await self.store.delete(self.scope, conv_id)
+
+    async def _evict_if_needed(self) -> None:
```
gadenbuie, line 349: **[Required] Eviction re-stats the whole scope every iteration — O(k·n) filesystem calls.**

`total_size()` is a full-directory `glob` + `stat` sweep, re-run on every eviction iteration, inside the reactive flush that fired the response. Track a running total and subtract each evicted record's size (or sum meta sizes once up front). (Same shape in R at `chat_history.R:347-355`.)

pkg-py/src/shinychat/_history.py
```diff
@@ -0,0 +1,878 @@
...
+        if self.max_store_bytes is None or self.scope is None:
```
gadenbuie: **[Required] Quota loop silently leaves the store over-budget when the active conversation alone exceeds the cap.**

If the active conversation is larger than `max_store_bytes`, everything else is evicted and the loop exits still over budget — no warning, no log — then re-runs the full `list()` + N×`total_size()` sweep on every subsequent response forever. Detect the unshrinkable case and `warnings.warn` once; stop re-sweeping silently.

pkg-r/R/chat_history.R
```diff
@@ -0,0 +1,868 @@
...
+    evict_if_needed = function() {
+      max_bytes <- private$max_store_bytes
+      if (is.null(max_bytes) || is.null(self$scope)) {
+        return(invisible())
+      }
+      total <- private$store$total_size(self$scope)
+      if (total <= max_bytes) {
+        return(invisible())
+      }
+      metas <- private$store$list(self$scope)
```
gadenbuie: **[Required] Eviction loop: O(n²) re-stat + silent over-budget (parity with the Python notes).**

`total_size()` is re-computed every iteration (full `list.files` + `file.size` sweep). Track a running total instead. Also, if the active conversation alone exceeds `max_bytes`, the loop exits still over budget with no warning and re-sweeps on every response — warn once and stop.

pkg-r/R/chat_history.R
```diff
@@ -0,0 +1,868 @@
...
+      session
+    ) {
+      title <- options$title
+      private$store <- resolve_store(options$store)
+      private$chat_id <- chat_id
+      private$client <- client
+      private$title_fn <- if (is.function(title)) title else NULL
+      private$title_enabled <- !is.null(title)
+      private$session <- session
+      private$max_store_bytes <- if (!is.null(options$max_store_mb)) {
```
gadenbuie, line 61: **[Required] `as.integer` overflows the storage cap at `max_store_mb >= 2048`.**

`as.integer(options$max_store_mb * 1024 * 1024)` exceeds `.Machine$integer.max` at `max_store_mb >= 2048`, yielding `NA`; then `if (total <= max_bytes)` (line 344) becomes `if (NA)` → error, breaking history with an opaque message. The no-limit path (`max_store_mb = NULL`) works correctly. Accumulate/compare as `double`; bytes should never be `integer`. (See also `chat_history_store.R:237`.)

pkg-r/R/chat_history_store.R
```diff
@@ -0,0 +1,270 @@
...
+    total_size = function(scope) {
+      sdir <- private$scope_dir(scope)
+      if (!dir.exists(sdir)) {
+        return(0L)
+      }
+      files <- list.files(sdir, pattern = "\\.json$", full.names = TRUE)
+      if (length(files) == 0L) {
+        return(0L)
+      }
+      as.integer(sum(file.size(files)))
```
gadenbuie: **[Required] `total_size` caps at ~2 GB via `as.integer`.**

Same overflow as `chat_history.R:61` — a scope over 2 GB returns `NA` and defeats eviction. Use `as.numeric`/double accumulation.

pkg-py/src/shinychat/_history_store.py
```diff
@@ -0,0 +1,234 @@
...
+       redeploy-safe bookmarks area — piggybacking on it gives history the
+       same persistence guarantees as server bookmarks, with zero config.
+    3. `.shinychat/conversations/` (plain local dev).
+    """
+    env = os.environ.get("CONNECT_CONTENT_DATA_DIR")
+    if env:
+        return Path(env) / HISTORY_BOOKMARK_ID
+
+    # Private shiny API; coordinate upstream for a public accessor.
+    from shiny.bookmark._global import get_bookmark_save_dir_fn
```
gadenbuie: **[Required] `resolve_history_dir` imports a private Shiny API with no fallback.**

`from shiny.bookmark._global import get_bookmark_save_dir_fn` has no `try/except` and crashes hard at first store use if the symbol moves — yet `_history_bookmark.py:20` wraps the identical import. Wrap it and fall back to `.shinychat/conversations` on `ImportError`.

js/src/chat/ChatHistoryDrawer.tsx
```diff
@@ -0,0 +1,418 @@
...
+  function handleClose() {
+    setQuery("")
+    setMenuFor(null)
+    setRenaming(null)
+    setConfirmingDelete(null)
+    onClose()
+  }
+
+  function handleDrawerAnimationEnd() {
+    if (!closing) return
```
gadenbuie, line 155: **[Required] Drawer can lock open when animations are disabled.**

Close is driven only by `onAnimationEnd` setting `visible=false` and restoring focus. With `prefers-reduced-motion: reduce` or animations otherwise disabled, `animationend` never fires → the `aria-modal` drawer becomes a permanently-open focus trap with no dismiss. The project already ships `usePrefersReducedMotion.ts` and uses `@media (prefers-reduced-motion)` in `chat.scss`/`_greeting.scss`/`_suggestion-cards.scss`. Honor that convention here and add a `setTimeout` close fallback so unmount never depends solely on the animation event.

js/src/chat/ChatHistoryDrawer.tsx
```diff
@@ -0,0 +1,418 @@
...
+  }, [isOpen])
+
+  useEffect(() => {
+    if (!visible || closing) return
+    const drawer = drawerRef.current
+    if (!drawer) return
+
+    const onKey = (e: KeyboardEvent) => {
+      if (e.key === "Escape") {
+        onClose()
```
gadenbuie: **[Required] Escape and rename-blur bypass the drawer's state reset.**

Escape calls `onClose()` (no reset) instead of `handleClose()`, so a mid-rename Escape closes the whole drawer and stale `renaming`/search state reappears on reopen. Rename `onBlur={onCancelEdit}` (line ~270) discards the edit on click-away — only Enter commits — and Enter with an empty `draft` sends an empty title to the server. Route every close through `handleClose()`, scope Escape to active sub-state, commit-or-guard on blur, and reject empty titles.

js/src/chat/ChatHistoryDrawer.tsx
```diff
@@ -0,0 +1,418 @@
...
+          }}
+          onBlur={onCancelEdit}
+          aria-label="Rename conversation"
+        />
+      </div>
+    )
+  }
+
+  return (
+    <div
```
gadenbuie: **[Required] Conversation row has invalid nested-interactive ARIA and stays tabbable while disabled.**

The row is `role="button"` with interactive `<button>`s (⋯ menu, ✓/✕) nested inside it — interactive controls must not be descendants of `role="button"`. The row also keeps `tabIndex={0}` while `aria-disabled`, and the `＋ New` button (line ~158) uses a bare glyph with no `aria-label`. Make the row a real `<button>` with the action menu as a sibling, set `tabIndex={-1}` (and skip handlers) when disabled, and label the New button.

pkg-r/R/chat_history.R
```diff
@@ -0,0 +1,868 @@
+extract_state_id <- function(url) {
```
gadenbuie, line 1: **[Required] `extract_state_id` is not single-match-safe.**

`regexpr`/`regmatches` without ensuring a single match, then `!nzchar(m)`: a URL with two `_state_id_=` params makes `m` length 2 and `||` errors/warns on a vector. Take `m[[1]]` or use `regexec`.

pkg-py/src/shinychat/_history.py
```diff
@@ -0,0 +1,878 @@
...
+            ):
+                token = chat._session.input[ids.browser_token]()
+                req(token)
+            if isinstance(scope_key, str):
+                return scope_key
+            if callable(scope_key):
+                return scope_key(chat._session)
+            if user is not None:
+                return str(user)
+            token = chat._session.input[ids.browser_token]()
```
gadenbuie, line 788: **[Suggestion] `session.user` is `str()`'d without a type contract.**

The caller-supplied `scope` is `isinstance`-checked, but `session.user` is `str()`'d as-is. It's a string by Shiny convention, but unenforced — if two users' `str()` collide they'd share a scope. Document the string-identity contract (and optionally `isinstance`-guard).

pkg-py/src/shinychat/_history_bookmark.py
```diff
@@ -0,0 +1,35 @@
...
+        from shiny.bookmark._global import (
+            get_bookmark_restore_dir_fn,  # type: ignore[attr-defined]
+        )
+        from shiny.types import MISSING  # type: ignore[attr-defined]
+
+        restore_dir_fn = get_bookmark_restore_dir_fn(MISSING)
+        if restore_dir_fn is None:
+            return
+        restore_dir = await restore_dir_fn(state_id)
+    except Exception:
```
gadenbuie, line 58: **[Suggestion] Bare `except Exception` hides real failures in bookmark cleanup.**

This wraps import + private-API call + await, so a genuine failure is indistinguishable from "not configured" and orphaned bookmark state accumulates silently. Narrow to `ImportError`/`AttributeError`; log the rest.

pkg-r/R/chat_history.R
```diff
@@ -0,0 +1,868 @@
...
+          }
+        },
+        add = TRUE
+      )
+
+      chat_clear(private$chat_id, session = private$session)
+
+      turns <- record_path_turns(record)
+      if (length(turns) > 0) {
+        set_turns_recorded(private$client, turns)
```
gadenbuie, line 224: **[Suggestion] The "Lossy" comment is inaccurate.**

Restore uses `ellmer::contents_record`/`contents_replay` (`chat_history_client.R:2,8`) — ellmer's lossless round-trip API, which serializes all Content types including `ContentThinking`, `ContentToolRequest`, and `ContentToolResult`. Thinking blocks round-trip fine. The comment misrepresents the behavior and could lead a maintainer to add unnecessary fidelity machinery. Correct or remove it. (Related design question on the `node.ui` divergence is noted separately.)

pkg-py/docs/_quarto.yml
```diff
@@ -84,6 +84,20 @@ quartodoc:
         include_functions: true
       contents:
         - express.Chat
+    - title: Conversation history
+      options:
```
gadenbuie, line 88: **[Suggestion] `HistoryOptions` is exported but missing from the API reference.**

`HistoryOptions` is part of the public surface (`types/__init__.py`) and carries `restore_mode`, `scope`, `max_store_mb`, but isn't listed in this quartodoc block. Add it (which also surfaces `restore_mode`).

pkg-r/NAMESPACE
```diff
@@ -5,11 +5,15 @@ S3method(format,shinychat_tool_card)
 S3method(knitr::knit_print,shinychat_tool_card)
 S3method(print,shinychat_tool_card)
 export(ContentSlashCommand)
+export(ConversationStore)
+export(FileConversationStore)
+export(InMemoryConversationStore)
```
gadenbuie: **[Suggestion] `InMemoryConversationStore` is public in R but private in Python.**

It's exported here but not in Python's `types/__init__.py` (it exists at `_history_store.py:132`). Confirm the intended public surface matches across packages.

js/src/chat/ChatHistoryDrawer.tsx
```diff
@@ -0,0 +1,418 @@
...
+      ref={drawerRef}
+      className="shiny-chat-history"
+      data-closing={closing || undefined}
+      role="dialog"
+      aria-modal="true"
+      aria-label="Conversation history"
+    >
+      <div className="shiny-chat-history-scrim" onClick={handleClose} />
+      <div
```
gadenbuie, line 183: **[Noted]** `onAnimationEnd` bubbles — guard `if (e.target !== e.currentTarget) return` so a child animation can't trigger premature unmount during close.

js/src/chat/state.ts
```diff
@@ -1017,6 +1027,22 @@ export function chatReducer(state: ChatState, action: AnyAction): ChatState {
...
+        ...state,
+        history: {
+          enabled: action.enabled,
+          conversations: action.conversations,
+          activeId: action.active_id,
+        },
+      }
+    }
+
+    case "history_navigate": {
```
gadenbuie, line 1041: **[Noted]** The `history_navigate` reducer case is dead code (intercepted in `ChatApp` before dispatch). Document it as unreachable or drop it.

pkg-py/src/shinychat/_history_types.py
```diff
@@ -0,0 +1,107 @@
...
+        self.current_leaf = node_id
+        self.updated_at = utcnow()
+        return node_id
+
+
+def new_conversation_id() -> str:
+    # Time-prefixed for rough sortability; token for uniqueness.
+    # (Avoids a ULID dependency. Ordering in the UI always comes from
+    # updated_at, never from the id.)
+    return f"c_{int(time.time() * 1000):013x}{secrets.token_hex(5)}"
```
gadenbuie, line 128: **[Noted]** `new_conversation_id()` is last-writer-wins on a same-millisecond id collision; worth a one-line acknowledgement.

pkg-py/src/shinychat/_history.py
```diff
@@ -0,0 +1,878 @@
...
+        turns = self.adapter.get_turns_json()
+        messages = self.chat._messages_for_bookmark()
+        extend_record_linear(
+            self.record, turns, messages, ui_offset=self.ui_offset
+        )
+        await self._capture_app_state(self.record)
+        await self.store.put(self.scope, self.record)
+        self.ui_offset = len(messages)
+
+    async def _capture_app_state(self, record: ConversationRecord) -> None:
```
gadenbuie: **[Noted]** `_capture_app_state`/`_restore_app_state` are `async def` with no awaits (callbacks are sync). Drop `async` or note why it's there.

pkg-r/R/chat_history_client.R
```diff
@@ -0,0 +1,46 @@
...
+    recorded_turns,
+    ellmer::contents_replay,
+    tools = client$get_tools()
+  )
+  client$set_turns(replayed)
+}
+
+get_client_info <- function(client) {
+  tryCatch(
+    {
```
gadenbuie: **[Noted]** Reaches into ellmer's R6 private layout (`client$.__enclos_env__$private$provider`); a refactor silently degrades to `list()`. Worth an upstream accessor request. (Same theme for the shiny-private reaches in `_history_store.py:225` / `_history_bookmark.py:20`.)

pkg-py/src/shinychat/_history_bookmark.py
```diff
@@ -0,0 +1,35 @@
...
+        return
+    p = Path(restore_dir)
+    if p.is_dir():
+        await asyncio.get_event_loop().run_in_executor(
```
gadenbuie: **[Noted]** `asyncio.get_event_loop()` inside `async def` is deprecated; use `get_running_loop()`.

pkg-py/src/shinychat/_history.py
```diff
@@ -0,0 +1,878 @@
...
+                )
+                if pointed is not None:
+                    adapter.set_turns_json(pointed.path_turns())
+                    await controller.replay_ui(pointed)
+                    await controller._restore_app_state(pointed.values or {})
+                    controller.record = pointed
+            await controller.send_history_update()
+            initialized = True
+
+        @reactive.effect
```
gadenbuie, line 862: **[Design question — please confirm] Does `_save_on_response` need a reentrancy guard?**

This is a long async chain (`store.put`, `total_size`, bookmark mint) driven by a reactive effect; `ui_offset` is read at the top and written at the bottom. If two `chat.messages` flushes can overlap during a slow `put`/mint, the second reads a stale offset and double-appends. Reactive effects are generally serialized per flush — is that guarantee relied on here, or should an `asyncio.Lock` enforce it? A comment either way would help.

pkg-r/R/chat_app.R
```diff
@@ -313,13 +337,24 @@ chat_server <- function(
...
+  saved_on_restore_fns <- list()
+
+  # Set up history or legacy bookmarking
+  cancel_history <- if (!isFALSE(history)) {
+    config <- if (isTRUE(history)) history_options() else history
+    chat_enable_history(id, client, options = config, session = session)
+  } else if (bm_on_input || bm_on_response) {
+    chat_restore(
+      id,
+      client,
```
gadenbuie, line 353: **[Design question — please confirm] Does `set_client()` with history enabled double-render the restored conversation?**

A client swap re-runs this history branch and spins up a fresh controller/init effect. The pre-PR path passed `restore_ui = FALSE` on swap to avoid re-rendering; the new history path has no obvious equivalent. Does a mid-session `set_client()` replay the UI twice? No test covers it.

js/src/chat/ChatApp.tsx
```diff
@@ -105,6 +107,11 @@ export function ChatApp({
   // the imperative handle rather than the reducer.
   useEffect(() => {
     const unsubscribe = transport.onMessage(elementId, (action) => {
+      if (action.type === "history_navigate") {
```
gadenbuie, line 110: **[Design question — please confirm] Two sources of truth for the active conversation id.**

The active id lives in both the reducer (written by `history_update`) and localStorage (written imperatively here alongside `history_navigate`; see also lines 205-208). If `history_navigate` and `history_update` arrive out of order, localStorage and the highlighted row can diverge. Is ordering guaranteed, or should one be derived from the other?

pkg-py/src/shinychat/_history.py
```diff
@@ -0,0 +1,878 @@
...
+        return
+
+    new_node_ids = [record.append_linear(t) for t in new_turns]
+    user_nodes = [
+        nid
+        for nid in new_node_ids
+        if record.nodes[nid].turn.get("role") == "user"
+    ]
+
+    for message in ui_messages[ui_offset:]:
```
gadenbuie, line 174: **[Design question — for team discussion] Persisting rendered UI (`node.ui`) vs. turn-only restore.**

The two packages diverge here: Python persists the rendered UI per node and replays it verbatim (`node.ui`, type at `_history_types.py:36`), while R stores only ellmer turns and re-renders via `content_shinychat`. Since `contents_record`/`contents_replay` round-trip all Content faithfully, turn-only restore already reproduces anything that originated from the model.

The case `node.ui` captures that turns don't is **out-of-band content** — e.g. `chat_append()` pushing content into the chat UI outside the conversation the client sees. That raises a real product question with downstream cost (storage/quota, the `ui_offset` bookkeeping, part of why the phantom-save guard exists):

- Should out-of-band UI content be persisted and restored as part of history, or treated as ephemeral and explicitly *not* part of the official record?
- Should the two packages converge on one model?

There are reasonable arguments either way; the right answer affects how much restore machinery is warranted. Surfacing it here because it reads as an implicit divergence rather than a documented decision — it'd be worth calling out in the PR description.

gadenbuie (changes requested):
This is a large, well-architected feature — the replay/suppress state machine, atomic writes, path sanitization, and cycle-guarded record traversal show real care. Inline notes below cover one **Blocking** issue (a module-id bug that silently disables history for any non-`"chat"` instance), a set of **Required** correctness/robustness fixes concentrated in the eviction loop and the drawer's close/edit flow, several **Suggestions** (docs, parity, an inaccurate comment), a few **Noted** minors, and four **open design questions** we'd like surfaced for team discussion rather than treated as defects.

**Test coverage (Suggestion):** the suites are happy-path-heavy and don't exercise the paths where these findings live — a non-`"chat"` module id (would have caught the Blocking issue), a dangling-parent record, cached-`list` + deleted-file `get`, an active conversation exceeding the quota, and a malformed/multi-param `_state_id_` URL. Please add regression cases alongside the fixes.

Requesting changes on the Blocking + Required items; the design questions are non-blocking but worth answering.

pkg-r/R/chat_app.R
cpsievert: Fixed in a8a38326 — all three sites now use `paste0(id, ".history-controller")`. Added a regression test with module id `"mychat"` (confirmed red against the old code).

pkg-py/src/shinychat/_history_types.py
cpsievert: Fixed in 9bc20653, exactly as suggested. Test: `test_path_node_ids_raises_on_dangling_parent`.

pkg-py/src/shinychat/_history_store.py
cpsievert: Fixed in 9bc20653 — kept the cache, fixed the handling: `get()` now pops the stale `_meta_cache` entry on a miss so the next `list()` re-reads disk. Test: `test_get_missing_after_cached_list_invalidates_cache`.

pkg-py/src/shinychat/_history.py
cpsievert: Fixed in 4aa5b2ae (see the reply on #6 for the full story — same redesign in both languages). `ConversationMeta` now requires `size_bytes`; `total_size()` is concrete on the base class from `list()`; eviction calls `list()` once with a running total. Test asserts `list()` called once, `total_size()` zero times.

pkg-py/src/shinychat/_history.py
cpsievert: Fixed in 9bc20653 — warns once via an instance-level flag when the active conversation alone exceeds the cap. Test: `test_evict_if_needed_warns_once_when_active_alone_exceeds_budget`.

pkg-r/R/chat_history.R
cpsievert: Fixed in 4aa5b2ae. We'd initially deferred this twice as a breaking interface change — since nothing depends on `ConversationStore` yet, we redesigned instead: `record_meta()`/`new_conversation_meta()` now require `size_bytes`, `total_size()` is concrete on the base class, eviction calls `list()` once with a local running total. Regression test uses an R6 subclass spy to assert call counts (R6 locks public bindings, so a plain mock wasn't an option).

pkg-r/R/chat_history.R
cpsievert: Fixed in 45a73b26 — `as.double()` instead of `as.integer()`. Regression test with `max_store_mb = 2048`.

pkg-r/R/chat_history_store.R
cpsievert: Fixed in 45a73b26 — `total_size()` stays double throughout in both store implementations. Regression test mocks `file.size()` to simulate a >2GB scope.

pkg-py/src/shinychat/_history_store.py
cpsievert: Fixed in 9bc20653 — falls back to `.shinychat/conversations` on `ImportError`. Note: catches `ImportError` only, not `(ImportError, AttributeError)` like `_history_bookmark.py` — happy to align them exactly if you'd like. Test: `test_falls_back_when_private_api_missing`.

js/src/chat/ChatHistoryDrawer.tsx
cpsievert: Fixed in 56cbd635 — immediate close under `usePrefersReducedMotion()`, otherwise a 300ms `setTimeout` fallback with proper cleanup so a stale timer can't fire after unmount.

js/src/chat/ChatHistoryDrawer.tsx
cpsievert: Mostly fixed in 56cbd635: Escape now routes through `handleClose()` (scoped to menu/delete-confirm first), rename's Escape `stopPropagation()`s instead of bubbling, and an empty/whitespace draft on Enter is a no-op. Left blur-discard as-is — an existing passing test (`"cancels rename on blur"`) confirms that's the intended UX, not a bug. Flag if you'd rather have commit-on-blur.

js/src/chat/ChatHistoryDrawer.tsx
cpsievert: Fixed, in two passes (30332ade, then b821f70d). Your suggested fix as literally stated (wrap the row in a `<button>`, keep ⋯/✓/✕ nested inside) turns out to be invalid HTML — browsers auto-reparent nested `<button>`s. Instead: the row is a plain layout `<div>`, the title/timestamp is a real `<button class="shiny-chat-history-item-select">`, and the ⋯ menu/delete-confirm are siblings, never descendants. A native `disabled` button gets correct tab-order exclusion and Enter/Space activation for free, so the manual `role`/`tabIndex`/`aria-disabled`/`onKeyDown` machinery is gone. Verified live in a browser against the `history_actions` Playwright app.

pkg-r/R/chat_history.R
cpsievert: Investigated, no fix — doesn't reproduce. `regexpr()` (unlike `gregexpr()`) already returns only the first match, so two `_state_id_=` params still yield a length-1 match with no warning — verified empirically. Happy to add `regexec` hardening anyway if you'd like future-proofing.

pkg-py/src/shinychat/_history.py
cpsievert: Addressed via a comment in ff04a6b8 documenting the string-identity contract. Skipped the `isinstance` guard since you flagged it as optional.

pkg-py/src/shinychat/_history_bookmark.py
cpsievert: Fixed in 91ab6fc7 — narrowed to `except (ImportError, AttributeError)`; everything else is logged via `logger.warning(..., exc_info=True)`. New tests cover both paths.

pkg-r/R/chat_history.R
cpsievert: Fixed in 3c9b8820 — comment now correctly describes lossless round-tripping via `contents_record`/`contents_replay`.

pkg-py/docs/_quarto.yml
cpsievert: Fixed in ff04a6b8 — added to the quartodoc section. Side note: `quartodoc build` was already broken pre-existing (a `WorkaroundKeyError` on the sibling `ConversationStore` entry) — out of scope here, flagging in case it's news.

pkg-r/NAMESPACE
cpsievert: Resolved — went with un-exporting in R rather than exporting in Python (0d5c13e0), treating it as an internal default rather than a public extension point. Shout if you intended the opposite direction.

js/src/chat/ChatHistoryDrawer.tsx
cpsievert: Fixed in 56cbd635, exactly as suggested.

js/src/chat/state.ts
cpsievert: No change needed — already resolved on `main` independently, with a comment and a dedicated test confirming the intentional no-op.

pkg-py/src/shinychat/_history_types.py
cpsievert: Acknowledged with a one-line comment in ff04a6b8.

pkg-py/src/shinychat/_history.py
cpsievert: Fixed in 4ef9664e — converted both to plain `def`; the now-unneeded `await`s at all 5 call sites were removed too.

pkg-r/R/chat_history_client.R
cpsievert: No code change — agreed this is worth an upstream accessor request rather than a workaround here.

pkg-py/src/shinychat/_history_bookmark.py
cpsievert: Fixed in 91ab6fc7 — replaced with `asyncio.get_running_loop()`.

pkg-py/src/shinychat/_history.py
cpsievert: Confirmed safe by construction, no lock added — Shiny serializes reactive flushes per-process behind a single `asyncio.Lock` from `reactive.lock()`, held across the entire per-message handling loop including awaits, so `on_response()` invocations can't overlap. Documented the guarantee in d4c5b15a.

pkg-r/R/chat_app.R
cpsievert: Good catch — this was a real, reproducible bug. Every `set_client()` call while history is enabled re-ran the init effect and replayed the UI plus double-fired `on_restore`, on every swap. Fixed in d02d18b9 with a `restore_ui` parameter (mirrors `chat_restore()`'s existing convention); `swap_client()` now passes `restore_ui = FALSE`. Reproduced with a regression test before fixing.

js/src/chat/ChatApp.tsx
cpsievert: Resolved — not currently exploitable but fragile, agreed. Every call site that fires both messages for the same transition always carries the same id, and sends are serialized by the same lock from #25, so there's no reachable divergence today. Added a comment in `ChatApp.tsx` (7d5a5d9e) documenting the invariant rather than a structural fix.

pkg-py/src/shinychat/_history.py
cpsievert: Still open — I'll think about this a bit more. 

I'm tempted to go in a direction where the UI message state derives from React state, and we pass that client->server via an input value. I've actually investigated this a bit when addressing some bookmarking issues, but ended up stashing it, mainly since I wasn't sure whether we'd want to sign up for this on the R side. There would be a couple notable upsides here:

1. Less state management complexity in Python. 
2. More consistency between R/Python in terms of the save/restore approach
    * This would move us away from turns being the source of truth for the UI in R, but I don't think the complexity/correctness tradeoff with this approach is as bad as it has been historically.

jcheng5: I think we encountered this question in Databot and ended up having to persist both (right, @wch?). It's super weird to navigate away from your conversation, and then back, and suddenly there are all these differences that don't seem to have much rhyme or reason.

pkg-r/R/chat_history_client.R
gadenbuie: Oh wait, if this is just reaching into the Chat R6 class for provider, that's available via the `$get_provider()` method

cpsievert: Fixed in 4fe6ea0a: R now uses ellmer's public `get_provider()`/`get_model()` (which had been available all along — the private reach was actually silently broken under S7 providers). On the Python side there's no public getter to swap to yet, so I consolidated the two reaches into one spot in `_history_bookmark.py`.

https://github.com/posit-dev/shinychat/commit/4fe6ea0a

pkg-py/src/shinychat/_history.py
cpsievert: I think I have a reasonable approach to persisting the UI message state via React state. 

That said, I think it'd probably make more sense to do this as a follow-up than to roll it into this branch (this PR is already stupidly large).

cpsievert:
@gadenbuie thanks for the extensive review, ready for another look if you are

gadenbuie:
@cpsievert looking now. gpt-5.5 found a small issue that `file.rename()` doesn't work as well on Windows so I just pushed a commit to fall back to `file.copy()` if needed.

gadenbuie:
## `store="auto"` / in-memory persistence: a few things worth discussing

While tracing how `store="auto"` interacts with Shiny devmode, I found a few things about the `ConversationStore` design that I think are worth flagging now since they bear on how developers will actually experience the default dev-mode path.

As a high-level summary, I'm a little nervous about pinning in-memory storage to `devmode()`, especially in the R case.

### 1. In-memory storage doesn't survive a plain page reload, not just a process restart

`resolve_store()` is called fresh, unmemoized, on every session start in both languages (`_history.py:637`, `chat_history.R:55`). For `FileConversationStore` that's harmless — the object is just a handle to a shared external resource (the directory). For `InMemoryConversationStore`, the object *is* the storage (`self._data` lives on the instance), so a new instance per session means history is gone on **any** new session, not just a `--reload` process restart.

Confirmed live: reloading the browser tab on a `store="auto"` (dev mode) app wipes the transcript, and the server actively tells the client to forget the active conversation id (`shinychat-current:<id>` gets cleared from localStorage, since the lookup against the fresh empty store comes back empty). Ran the same test against `FileConversationStore` as a control and it survives reload fine, so this isn't a test artifact.

This means the log message ("History is lost on restart") undersells what's actually happening — it reads as "survives while the server is running," but it doesn't even survive a browser refresh.

### 2. Proposal: dev-mode in-memory storage should be global by `chat_id`, not scoped by browser

As a developer, losing my test conversation every time I refresh the page (let alone switch browsers to check something) is pure friction with no upside — during local dev, "different browser" essentially always means "same person, testing," not two different end users. I'd propose `store="auto"`'s in-memory path stop scoping by browser token and instead key purely by `chat_id`, kept alive for the life of the process. That would also fix (1) above as a side effect — reload as many times/browsers as you want, dev-mode history survives until the process actually restarts (or autoreload respawns it), which matches what I'd guess was the original intent.

Whether this should be `InMemoryConversationStore`'s default behavior unconditionally, or specifically gated to dev mode (so an explicitly-chosen `store="memory"` outside dev mode keeps today's browser-token scoping) is an open question — my instinct is the latter, since "ignore browser identity" is really a dev-mode-specific assumption, not a property of in-memory storage in general. Curious what you think.

### 3. This surfaces a real gap: `ConversationStore` has no `chat_id`/app-identity partition key

Digging into what "global by `chat_id`" would require, I noticed `ConversationStore.list/get/put/delete` partition **only by `scope`** — there's no `chat_id` anywhere in `ConversationRecord`/`ConversationMeta`. And the default `scope` (browser token) is explicitly shared across every chat element on a page. So today, the *only* thing keeping two `Chat()`/`chat_server()` widgets with the same scope from mixing histories together is that `resolve_store()` happens to hand each `HistoryController` its own fresh, unshared store object — isolation by accident (object identity), not by an actual partition key.

That's fine as long as every store construction stays 1:1 with a `HistoryController`, but it means:
- We can't fix (1)/(2) with a naive shared singleton without first adding `chat_id` (or similar) to the partition key — otherwise two chat widgets sharing scope would suddenly see each other's conversations.
- Multi-chat-widget apps deserve first-class support regardless of the dev-mode question — I'd expect separated history *by default*, with a documented, supported way to opt into sharing (e.g. passing the same explicit `scope` to both). Right now that's not really something the interface can express cleanly.

cpsievert:
>  losing my test conversation every time I refresh the page 

~Interesting, I'm not able to reproduce that behavior. Could you share a reprex?~

nvm, I've got it now

cpsievert:
Fixed in 46e11c98:

**(1) Reload wiping history.** Confirmed. `resolve_store()` built a fresh `InMemoryConversationStore()` per session, so the store itself never outlived one session — reload or new tab, new empty store. Fix: `store="auto"` in dev mode now hands out one process-lifetime singleton (`auto_dev_memory_store()`), so history survives reloads and new tabs/browsers and is only lost when the process actually restarts, matching what the log message already claimed.

**(2) Missing `chat_id` partition key.** Also confirmed, and fixed as a prerequisite to (1): added `ConversationPartition(chat_id, scope)` and switched `ConversationStore.list/get/put/delete` to take it instead of a bare `scope` string. Two `Chat()`/`chat_server()` widgets sharing a scope are now isolated by an actual key rather than each happening to get its own store instance — which is what made introducing the singleton in (1) safe. Multi-widget apps get separated history by default now, with sharing available by passing the same explicit `scope` to both.

**(3) Scoping by browser token in dev mode.** Went a different direction than proposed: the partition still includes the browser-token scope, so the dev-mode singleton is one store per process, but still internally partitioned by browser token — it doesn't collapse to global-by-`chat_id`. Reasoning: if `SHINY_DEV_MODE` ends up set in a shared/multi-user deployment by mistake, dropping browser-token scoping means every viewer sees every other viewer's history under that `chat_id` — that's a data-exposure failure mode I don't want as a default, even on a dev-only path. The reload-losing-history complaint is fully addressed by the singleton alone, since browser tokens persist in localStorage across reloads in the same browser — it was never the token causing the loss, it was the store being rebuilt from scratch every time. Different-browser/incognito testing during dev still gets separate (but now persistent-within-that-browser) history, which seems like the safer default. Open to revisiting if you've got a reload-losing-history scenario this doesn't cover.

Tests: partition isolation (Python + R), the process-singleton behavior, and an end-to-end Playwright test (`history_auto_memory_reload`) that reloads a live dev-mode app and asserts the conversation survives.

gadenbuie:
> **(3) Scoping by browser token in dev mode.** Went a different direction than proposed:

Thanks, that makes sense. I do feel like we're approaching the line where devmode is overloaded as a signal, but I think the approach you went with is a very reasonable compromise.

gadenbuie (approved):
this is awesome, thank you! I'm ready to merge this so we can follow up with smaller PRs 😄

PR #269, opened 2026-07-01
feat: editable messages and sibling navigation

cpsievert:
### What this PR does

You can now edit a message you already sent, instead of only being able to send a new one. Hover a user message and click the pencil icon to open an inline editor (pre-filled with the original text); press **Cmd/Ctrl+Enter** to save and resend, or **Escape** to cancel. Editing a message forks the conversation from that point — the old branch isn't lost, it's kept as a sibling. `‹ 1 / 2 ›`-style controls appear on any message with more than one version, letting you step back and forth between them at any time, including after reloading the page or returning from the history drawer.

This builds on top of the [multi-conversation history feature](https://github.com/posit-dev/shinychat/pull/266) in the parent branch — editing and branch navigation both go through the same persistence path as regular turns, so edits and the sibling you branched from are saved to the conversation store and are still there (and still navigable) the next time you open that conversation, not just for the rest of the current session.

---

### Reviewer guide: architectural decisions

This branch is stacked on `feat/chat-history` (#266) — only the diff introduced by the tip commit is new; everything else is already reviewed/merged there.

**1. The tree schema stops being inert**

`feat/chat-history` deliberately built `ConversationRecord` as a tree of `ConversationNode`s "solely to support future message editing without a schema migration," while noting the branching machinery was inert — no sibling-creation path, no branch-selection UI. This PR is that future: `UpdateSiblingsAction`, `HistoryController.handle_edit` / `handle_navigate`, and the new tree-query helpers (`siblings_of`, `subtree_leaf`, `path_sibling_metadata`, `node_id_for_message_index`) are what actually walk the tree instead of treating it as a list.

**2. How editing forks a branch**

`handle_edit` doesn't create the sibling node itself — it rewinds `current_leaf` to the edited message's parent and replays the UI up to that point. The sibling is created implicitly when the resubmitted content flows back through the normal `on_response` save path (append-to-`current_leaf`), the same path a brand-new turn takes. The comment in `_history.py` around `handle_edit` calls this out explicitly, since it's easy to assume `handle_edit` does the branching directly.

**3. Navigation persists immediately, edits persist on response**

`handle_navigate` calls `store.put()` itself, since switching the active sibling doesn't otherwise trigger a save. `handle_edit` relies on the existing save-on-response effect, since the fork isn't "real" until the resubmitted turn actually completes.

**4. Sibling metadata is a side channel, not part of the message payload**

`_send_sibling_metadata` walks `path_node_ids()` and maps each node to a UI message index, sending an `update_siblings` transport action keyed by index rather than embedding position/count in the message data itself. Worth checking that this stays in sync with `ui_offset` and replay logic if either changes later.


cpsievert:
@gadenbuie this is still in need of an R port -- other than that, I think it's in a good spot if you wanted to poke around with it. I'll probably try to stay focused on my tutorial next week, so feel free to polish this up if you'd like.

pkg-r/R/chat_history.R
```diff
@@ -938,13 +1057,91 @@ chat_enable_history <- function(
...
+  # The save trigger: fires once the browser echoes an updated `_messages`
+  # snapshot back to the server. This must be event-driven (not chained
+  # directly onto the streaming response's promise) because the echo is a
+  # separate round trip -- the browser only reports a finished assistant
+  # reply *after* it has rendered it and sent it back over the websocket.
+  # Triggering on stream completion instead would run on_response() one
+  # message ahead of what the client has actually reported, and
+  # extend_record_linear() would misattach the still-unreported assistant
+  # reply to the wrong (next round's) node.
+  message_response_effect <- shiny::observeEvent(
```
gadenbuie, line 1086: This `message_response_effect` block looks like it duplicates the existing observer at ~line 1035 — same `_messages` input, same `label = "history_on_response"`, same body. If the intent here was to replace that earlier one (the comment is updated), it looks like the original wasn't removed, so two things seem to follow:

- Both observers register, so `on_response()` fires twice per client echo. The count guard around `:99–101` may already absorb the duplicate, so this could be harmless today.
- `message_response_effect` now points at this second observer, so `cancel()` calls `$destroy()` on it twice (`:1141` and the added `:1144`) and the earlier one is never destroyed — leaving it firing after teardown.

Was the earlier block meant to be removed as part of this change?

js/src/chat/ChatMessage.tsx
```diff
@@ -33,20 +41,165 @@ function parseLeadingCommand(
...
+  const focusEditor = useCallback(() => editRef.current?.focus(), [])
+  const staging = useAttachmentStaging({
+    uploadAccept,
+    maxUploadSize,
+    enableUpload,
+    focusEditor,
+  })
+  const { applyPayloads, getPayloads, onPaste, onDrop } = staging
+
+  useEffect(() => {
```
gadenbuie, line 109: Question about the seeding effect: since its deps include `message.content` / `message.attachments`, would it re-run if one of those changes while `isEditing` is still true — and if so, could it overwrite text the user is mid-way through typing (and re-steal focus)? I wasn't sure how often a new `ChatMessageData` object for the same `id` appears during an open edit under the replay model. If it can, would seeding once on the open transition (e.g. guarding on a ref) be worth considering?

js/src/chat/ChatMessages.tsx
```diff
@@ -1,20 +1,63 @@
...
+  disabled,
+  inputId,
+  submitKey,
+  uploadAccept,
+  maxUploadSize,
+  enableUpload,
+}: ChatMessagesProps) {
+  // Only one message can be edited at a time: opening edit on any message
+  // overwrites this index, which implicitly closes whichever one was open.
+  const [editingIndex, setEditingIndex] = useState<number | null>(null)
```
gadenbuie: Small related thought to the comment on `ChatMessage.tsx` — `editingIndex` tracks a position while the list is keyed by `msg.id`. If the array ever reorders or shrinks while an edit is open, `i === editingIndex` would point at a different message than the one opened. It does seem like keying the edit state on `msg.id` would be more robust — is there a reason to prefer the index?

pkg-py/src/shinychat/_history_types.py
```diff
@@ -103,6 +103,45 @@ def path_turns(self) -> list[dict[str, Any]]:
...
+                result[nid] = (siblings.index(nid), len(siblings))
+        return result
+
+    def node_id_for_message_index(self, index: int) -> tuple[str, int]:
+        if index < 0:
+            raise IndexError(f"Message index {index} out of range")
+        path = self.path_node_ids()
+        cumulative = 0
+        for i, nid in enumerate(path):
+            n_ui = len(self.nodes[nid].ui or [])
```
gadenbuie: A question about the index math here vs. `replay_ui`: this counts `len(node.ui or [])`, but `replay_ui` (`_history.py:449–462`) renders one fabricated fallback message when a node's `ui` is `None`. If that's right, would the client's message positions and this mapping disagree whenever an on-path node has `ui = None` — e.g. client index 0 mapping to the wrong node, or the last index raising `IndexError`?

`ui = None` on the active path may be uncommon (normal turns populate `ui`), but the schema seems to allow it and there's a test for the state. If it is reachable, would a shared `ui_message_count(node)` (returning `len(node.ui)` or `1`) used by `replay_ui`, this method, and `_send_sibling_metadata` keep them from drifting?

pkg-py/src/shinychat/_history.py
```diff
@@ -489,6 +501,104 @@ async def delete(self, conv_id: str) -> None:
...
+        else:
+            if current_pos == len(siblings) - 1:
+                return
+            target = siblings[current_pos + 1]
+
+        leaf = self.record.subtree_leaf(target)
+        self.record.current_leaf = leaf
+        self.adapter.set_turns_json(self.record.path_turns())
+        await self.replay_ui(self.record)
+        self._suppress_next_save = False
```
gadenbuie: `_suppress_next_save` is assigned here (and at ~575) but I couldn't find anywhere it's read or initialized. It looks like a leftover from an earlier approach that could be removed — am I missing a read somewhere?

pkg-r/R/chat_history.R
```diff
@@ -381,6 +383,119 @@ HistoryController <- R6::R6Class(
...
+      if (!is.null(attachments)) {
+        # sendMessageEdit() always sends `attachments` (defaulting to `[]`),
+        # which jsonlite deserializes to a non-NULL, length-0 list() -- so an
+        # empty-but-present list must still force attachment_mode = "set"
+        # below, rather than falling to the else branch and leaving whatever
+        # happens to be staged in the main compose box untouched.
+        #
+        # Same normalize-then-validate pattern as the regular (non-edit) send
+        # path in attachments.R -- never trust client-side validation alone.
+        validate_attachments(attachments)
```
gadenbuie, line 484: Parity question against the Python `handle_edit`: Python re-validates and then reconstructs a normalized `{mime, data_url, name, size}` list (dropping any extra client fields), whereas here the raw client `attachments` are validated and passed straight through. Downstream only seems to read mime/data_url/name so this may be fine — is matching Python's normalize-then-reconstruct step worth it, or is the passthrough intentional?

js/src/chat/ChatMessages.tsx
```diff
@@ -1,20 +1,63 @@
...
+            onEdit={onEdit}
+            onNavigate={onNavigate}
+            disabled={disabled}
+            inputId={inputId}
+            submitKey={submitKey}
+            uploadAccept={uploadAccept}
+            maxUploadSize={maxUploadSize}
+            enableUpload={enableUpload}
+            isEditing={i === editingIndex}
+            onStartEdit={() => setEditingIndex(i)}
```
gadenbuie: `onStartEdit` / `onCancelEdit` are defined inline, so `ChatMessage`'s `memo()` can't skip re-renders — the prop reference changes every render, so every message subtree reconciles on each parent render. On long threads during streaming (frequent re-renders) this class of memoization issue tends to cause real, noticeable slowdowns, so I'd suggest stabilizing these with `useCallback` (passing `index` as an argument). The same shape appears in `AttachmentTray.tsx:51–54`.

pkg-py/src/shinychat/_history_types.py
```diff
@@ -103,6 +103,45 @@ def path_turns(self) -> list[dict[str, Any]]:
...
+            ]
+            children.sort(key=lambda nid: int(nid.split("_")[1]))
+            return children
+        return list(self.nodes[node_id].children)
+
+    def siblings_of(self, node_id: str) -> list[str]:
+        parent = self.nodes[node_id].parent
+        return self.children_of(parent)
+
+    def subtree_leaf(self, node_id: str) -> str:
```
gadenbuie, line 144: `subtree_leaf` follows the newest child down to a leaf. `current_leaf` is already persisted and `switch_to` restores it, so returning to a conversation lands on the last-viewed version at the top level. One level down, though, there's a gap: when navigating between siblings, `handle_navigate` recomputes `subtree_leaf(target)` (the newest leaf) rather than the descendant the user last had selected inside that subtree.

That seems worth addressing. If two sibling branches each have their own sub-branches, a user comparing them has to re-descend into each subtree every time they switch sides instead of picking up where they left off, and it gets more costly the deeper the branches go — which cuts against comparing alternative versions, a good part of the point of branching.

A fairly contained way to do it: add an optional `selected_child` to `ConversationNode`, set it along the active path wherever `current_leaf` moves (only three sites today), and have `subtree_leaf` follow it, falling back to the newest child when unset. It's one field, backward-compatible (unset → today's behavior), and the main tradeoff is introducing a downward pointer to keep consistent with the upward `current_leaf` walk. Does that seem like a reasonable direction?

js/src/chat/ChatMessage.tsx
```diff
@@ -130,90 +283,212 @@ export const ChatMessage = memo(function ChatMessage({
...
+                <button
+                  type="button"
+                  disabled={message.siblings.index === 0 || disabled}
+                  onClick={() => onNavigate?.(index, "prev")}
+                  aria-label="Previous version"
+                >
+                  ‹
+                </button>
+                <span>
+                  {message.siblings.index + 1} / {message.siblings.total}
```
gadenbuie, line 466: Accessibility: the `‹` / `›` buttons have `aria-label`s, but the `1 / 2` counter is a plain `<span>`, so screen readers won't announce the change when the user navigates versions. Suggest wrapping it in `aria-live="polite"` so the new position is announced.

pkg-py/src/shinychat/_history.py
```diff
@@ -489,6 +501,104 @@ async def delete(self, conv_id: str) -> None:
                 await self.on_active_id_change(None)
         await self.send_history_update()
 
+    # -- branch navigation --------------------------------------------------
+
+    async def _send_sibling_metadata(self) -> None:
+        if self.record is None:
+            return
+        sibling_meta = self.record.path_sibling_metadata()
```
gadenbuie, line 509: Edge-case question: `_send_sibling_metadata` returns early when there are no branch points, and the client seems to only clear a stale badge when it receives an `update_siblings` action omitting that index (`state.ts:1076`). I couldn't construct a case where a path goes from having siblings to none within one conversation, so this is likely a non-issue — but would sending an explicit empty payload (or a comment on why it's safe) be worthwhile, to be sure a stale `‹ 1 / 2 ›` can't linger?

gadenbuie (commented):
Really nice work — I read through the JS, Python, and R sides and the design reads well. The implicit-fork approach (rewind `current_leaf`, replay, then let the resubmit flow through the normal turn path) is elegant, and the client/server message-index alignment appears to hold because sibling metadata is only sent at settle points after a replay.

I've left a few inline comments below.

pkg-r/R/chat_history.R
cpsievert: Yes — the earlier block should have been removed; the duplicate was a rebase artifact. Fixed in `2a276ee6`: dropped the first `message_response_effect` block and the extra `message_response_effect$destroy()` in `cancel()`. The observer now registers once (so `on_response()` fires once per echo) and is destroyed once.

js/src/chat/ChatMessage.tsx
cpsievert: Real risk — the deps include `message.content`/`message.attachments`, so a new `ChatMessageData` for the same id arriving mid-edit re-runs `setInputValue` and both clobbers the in-progress text and re-steals focus. Fixed in d6dd7b4d by seeding only on the `false -> true` `isEditing` transition (guarded on a `wasEditingRef`), so later content/attachment changes while editing no longer re-seed.

js/src/chat/ChatMessages.tsx
cpsievert: Agreed, id is more robust. 1e1a6b1f switches the open-edit state from `editingIndex` to `editingId` and compares `msg.id === editingId`, so it stays anchored to the right message if the list reorders or shrinks while an edit is open. (Same commit also does the `useCallback` change from the comment below.)

pkg-py/src/shinychat/_history_types.py
cpsievert: Confirmed the drift and fixed it in fa13fe51 with the shared helper you suggested: `ConversationNode.ui_message_count()` (and R's `record_ui_message_count()`), returning `len(ui)` or `1`. It's now used by `node_id_for_message_index()` and the sibling-metadata builder in both languages, so they count a `ui=None` node as the one fabricated message `replay_ui()` renders — client indices and the server mapping stay aligned.

One direction call worth flagging: the "correct" count is subtly state-dependent (a `ui=None` node shows a fabricated message on *restore* but not during *live* streaming). I aligned everything to `replay_ui`'s fabricate-1 because the `ui` field is documented as "re-render from turns on restore (lossy but never broken)" — restore must render *something* — which makes `replay_ui` the single source of truth the counters mirror. A live `ui=None` node on the active path isn't reachable in normal operation (tool-call rounds group into one node *with* ui). This did mean rewriting `test_node_id_for_message_index_skips_empty_ui_nodes`, which encoded the under-counting side of the drift. R had the identical drift (your comment was Python-only), so I fixed it there too for parity.

pkg-py/src/shinychat/_history.py
cpsievert: You're right, it's dead. Removed both assignments in fa13fe51 (and the test that only asserted the flag). Python doesn't need it: the count-based guard in `on_response` already skips restore echoes. R keeps its `suppress_next_save` because R's `replay_ui` re-sends the stored UI to the client, so it suppresses the resulting echo explicitly — that asymmetry is now the documented reason.

pkg-r/R/chat_history.R
cpsievert: Matched Python in 8aaa65d1: after `validate_attachments()`, R now reconstructs a canonical `{mime, data_url, name, size}` record per attachment, dropping any extra client-sent fields before echoing them back through `update_input`. (The prior comment already claimed this pattern, so the code now actually matches it.)

js/src/chat/ChatMessages.tsx
cpsievert: Fixed the `ChatMessages` case in 1e1a6b1f: `onStartEdit`/`onCancelEdit` are now stable `useCallback`s (with `onStartEdit` taking the message id), so `ChatMessage`'s `memo()` can skip untouched subtrees during streaming.

On `AttachmentTray.tsx:51-54`: I left that one as-is. `AttachmentPreview`'s inline `onRemove`/`registerRef` do break its `memo()`, but the attachment tray is a short list that isn't re-rendered on the streaming path, so the memo savings are negligible there. Happy to stabilize it too if you'd prefer consistency — just wanted to avoid churn for no measurable win.

pkg-py/src/shinychat/_history_types.py
cpsievert: Implemented the `selected_child` direction you proposed, in fa13fe51:

- Optional `selected_child` per `ConversationNode` (unset => today's newest-child behavior, so it's backward-compatible and round-trips through both stores).
- A new `set_current_leaf()` records the selected child at each node along the active path wherever the leaf moves (the three sites: `append_linear`, `handle_navigate`, `handle_edit` — mirrored in R).
- `subtree_leaf()` follows `selected_child` when set, falling back to the newest child otherwise.

Off-path nodes are untouched, so each subtree keeps its own remembered position — switching between two sibling branches with their own sub-branches now returns to where you left each one. Tests added in both languages.

js/src/chat/ChatMessage.tsx
cpsievert: Wrapped the `1 / 2` counter in `aria-live="polite"` in 64a6a97b, so screen readers announce the new position when navigating versions.

pkg-py/src/shinychat/_history.py
cpsievert: Confirmed it's a non-issue and documented why, in fa13fe51 (comment at the early return). A badge only exists at a fork point (a node with >1 sibling), and forks are permanent: navigating between siblings keeps that node's sibling count > 1, and nothing prunes nodes from a record. So a path that currently has no forks never had one — empty metadata always means the client already shows no badges, and there's nothing stale to clear.

PR #270, opened 2026-07-02
feat: refine chat history trigger, drawer, and fill layout

gadenbuie:
## Summary

A refinement pass over the multi-conversation history UI (shipped in #266), polishing the drawer chrome and tightening how it integrates with the surrounding layout. Purely front-end (JS/SCSS) plus the generated dist assets copied into both packages; the Python/R APIs are unchanged except for one internal container-style helper.

**Layout anchoring.** The history trigger and drawer are now direct children of `<shiny-chat-container>` and anchored to the full-width container, while a new `.shiny-chat-wrapper` holds the width-limited, centered chat column. The container stays `display: grid` (element-selector SCSS beats bslib's flex rule) so bslib's fill child-rules stay inert and the messages area still grows. The chat column width is driven by `--_chat-width` (default `min(680px, 100%)`); `data-history-placement="right"` mirrors the trigger + drawer to the right edge (pure CSS, settable via `chat_ui(**kwargs)`).

**Sidebar clearance.** In sidebar layouts the trigger no longer collides with sidebar controls: a JS overlap-nudge probes the trigger's leading edge and shifts it inward past any obstacle, re-running on resize, sidebar transitions, and drawer open/close. While the drawer is open over a collapsed same-side sidebar, the bslib collapse-toggle is hidden (it would otherwise paint over the drawer). The `min-height` floor is set at element specificity with the 400px default supplied by a zero-specificity `:where()` rule, so both `min-height` and `--_min-height` remain author-overridable.

**Drawer redesign.** `_history.scss` and `ChatHistoryDrawer.tsx` were rewritten to follow the repo's `_command-chip.scss` token idiom (public `--shiny-chat-history-*` → private `--_` → `--bs-*` fallback, with `light-dark()` + `color-mix()` tints for hover/active states). The drawer is now a flex column with a pinned header (title + close) and toprow (ghost "New" button + search), and a scrollable list of dense single-line rows. Unicode-glyph controls were replaced with inline SVG icons; the trigger icon is now clock-rewind (history) rather than a hamburger. New/search share height via toprow `align-items: stretch` and a shared border token; type hierarchy runs header > controls (0.875rem) > titles > meta. Initial focus lands on the dialog container, and Close is first in tab order (matches visual reading order, WCAG 2.4.3).

**Fill-padding transfer.** When a `fill` chat is the sole child of a padded bslib fillable container (sidebar main, `card()` body, `page_fillable`), the trigger/drawer would inherit that padding and sit inset from the parent edge. `ChatContainer` now climbs the single-child fillable chain to the first padded ancestor, zeroes its inline padding, and re-applies the captured value to the centered content column via a public `--shiny-chat-fill-padding` token — so the trigger/drawer sit flush with the parent's outer edge while messages/input keep equivalent breathing room.

## Verification

Run this app and open the history drawer (move the `chat_ui(...)` call between the marked spots to exercise each layout):

```python


def app_ui(request: object) -> ui.Tag:
    return ui.page_sidebar(
        ui.sidebar("Sidebar content"),
        # (A) Chat inside a padded card body: trigger/drawer sit flush with the
        #     card edge while the card keeps its surrounding `main` padding.
        ui.card(
            ui.card_header("Chat with history", ui.input_dark_mode()),
            ui.card_body(
                chat_ui("chat", data_history_placement="right"),
                padding="64px",
            ),
        ),
        # (B) Move the same chat_ui(...) call here (drop the card) to put the
        #     chat directly in the sidebar `main`: trigger/drawer go flush with
        #     the main edge and the content column keeps the 24px padding.
        fillable=True,
    )


def server(input: Inputs, output: Outputs, session: Session) -> None:
    Chat(id="chat")  # attach any chatlas/ellmer client


app = App(app_ui, server)
```

Check:
- Full-width grid container with the chat column centered at 680px; scrim spans the whole container.
- `data_history_placement="right"` mirrors trigger + drawer to the right edge; dropping it (or setting `"left"`) mirrors them back left.
- The trigger clears the sidebar toggle, and the sidebar collapse-toggle hides while the drawer is open over a collapsed same-side sidebar.
- In both spots (A) and (B) the trigger/drawer are flush with the parent's outer edge while the content column keeps the parent's padding.

Automated: `cd js && npm run lint` (clean); 636 JS tests pass; `ruff` + `pyright` clean; R `chat`/`greeting` snapshot tests pass.


js/src/chat/ChatContainer.tsx
```diff
@@ -192,6 +192,162 @@ export const ChatContainer = forwardRef<
...
+    }
+  }, [scrollRef])
+
+  // Keep the history trigger clear of other visible UI. It anchors to a corner
+  // of the container, where it can collide with e.g. a bslib sidebar reveal
+  // button. When the trigger's home position overlaps another element that
+  // isn't part of the chat, nudge it inward (toward the chat column) via the
+  // --_history-trigger-shift var until it clears, capped so it never travels
+  // too far into the content.
+  useEffect(() => {
```
cpsievert: Would it make sense to extract the logic for these two effects into helper functions? I'm mainly just worried about an already monolithic component becoming even larger.

cpsievert (approved):

js/src/chat/ChatContainer.tsx
gadenbuie: Good call — done in feadf35. Both effects now live in custom hooks alongside the existing `useFullscreen`/`useInputHistory` ones:

- `useFillPaddingTransfer(scrollRef)` — a verbatim lift.
- `useOverlapNudge(elementRef, options)` — I generalized this one into a reusable "keep a corner-anchored control clear of an overlapping sibling" hook rather than hardcoding history. The history trigger is now just one caller, passing its boundary, shift property, placement, drawer-open skip guard, and mutation watch as config. Behavior is unchanged.

That drops ~145 lines from the component down to two declarative calls.

PR #272, opened 2026-07-03
feat: make the browser the source of truth for chat messages

cpsievert:
Faithfully saving and restoring a conversation means tracking two things: the
**LLM-facing state** (the turns to and from the model) and the **user-facing
state** (what's actually displayed). They aren't always the same — apps can
transform a response in a way that isn't (or shouldn't be) visible to the LLM 
or shinychat.

Now that the UI runs on React, the user-facing state already lives in the
client. This PR leans into that: instead of the server tracking displayed
messages on its own, the browser reports its state and the server treats it as
authoritative — one mechanism shared by both R and Python.

## What changes

- **The client reports its rendered messages.** The React app sends a
  settled-message snapshot to the server as a new `${id}_messages` input; the
  server reads that as the record of the user-facing conversation.
- **Python** already tracked user-facing state, but in a separate server-side
  accumulator (`_messages` / `_store_message`) — a redundant copy of what the
  browser holds. That's removed; `.messages()`, bookmark serialization, and
  history now read the client's report. History's save-idempotency guard drops
  its two replay-suppression flags for a simpler count-based check.
- **R** never tracked user-facing state, so its history (#266) restored
  straight from turns and lost every transformation. It gains the same protocol
  and now restores the exact displayed message, falling back to turn-derived
  markdown only for conversations captured before this change.

## Why this is better

- **R and Python history now share one model.** Previously each answered "what
  did the user see?" differently — Python kept a server-side accumulator, R
  didn't track it at all — so the two implementations diverged in both structure
  and fidelity. Now both consume the same client report through mirrored
  `HistoryController` logic (a reported-messages read, a `ui_offset` cursor, and
  the same heuristic for attaching UI messages to the record). Fixes and
  behavior port directly across languages, and keeping the two in parity is much
  easier.
- **Less server-side state to drift.** A single source of truth lets Python
  delete its duplicate ledger and the two replay-suppression flags that guarded
  save idempotency — fiddly bookkeeping that still missed partial mid-restore
  reports. The guard is now a plain count comparison.
- **Faithful by construction.** The record is whatever the browser rendered, so
  transformations are captured automatically instead of relying on every server
  code path to mirror them.

## Tests

New Playwright coverage for the client-authoritative behavior (`client_state/`)
and its interaction with history (`history_idempotent_save`,
`history_out_of_band`, `history_cross_session_deps`,
`history_ui_offset_after_restore`), plus expanded R `testthat` coverage for the
new store format and node/UI attachment logic.


cpsievert:
@gadenbuie ready for your eyes when available

pkg-r/R/chat_history.R
```diff
@@ -79,10 +85,14 @@ HistoryController <- R6::R6Class(
         rlang::abort("History controller partition not set")
       }
 
+      messages <- get_reported_messages(private$session, private$chat_id)
```
gadenbuie, line 95: This reads whatever the client last reported, but `on_response()` is invoked from the stream-completion promise (`chat.R` → `chat_history_on_response`), which resolves once the *server* finishes streaming. At that point the browser hasn't yet received the final chunk, rendered it, and reported its new snapshot back — that's a round-trip that lands after this runs. So I *think* `messages` here is the snapshot from *before* this response, missing the assistant message we just finished.

If that's the case, `extend_record_linear` would create the new turn node but find no new UI to attach (`ui = NULL`), and on restore that node falls back to `turn_fallback_markdown()` — losing the transformation this change is meant to preserve, at least for the last message of a conversation.

Compare Python, where the save is `@reactive.event(chat.messages)` — driven *by* the report arriving, so turns and UI are consistent when it fires.

Am I misreading the ordering? If not, would it be worth keying the R save off the `_messages` input changing (as Python does) rather than off stream completion? A real-browser test around last-message fidelity would also help pin this down — I didn't spot Playwright coverage for the R round-trip (the `client_state/`/`history_*` suites look Python-only), and the testthat cases pre-set `chat_messages` with the full transcript before calling `on_response`, which sidesteps the timing.

pkg-r/R/chat_history.R
```diff
@@ -79,10 +85,14 @@ HistoryController <- R6::R6Class(
...
+
       first_save <- is.null(self$record)
       if (!first_save) {
-        existing_count <- length(record_path_node_ids(self$record))
-        if (length(recorded_turns) <= existing_count) {
+        if (
+          length(recorded_turns) <= record_turn_count(self$record) &&
+            length(messages) <= record_ui_count(self$record)
```
gadenbuie, line 101: Minor observation: R now has both the `is_replaying`/`suppress_next_save` flags (above) *and* this count-based guard, whereas the PR description mentions Python dropped its two replay-suppression flags in favor of the count comparison. Not necessarily wrong to keep both, but if the count guard is sufficient here too it might be worth removing the flags for symmetry — or, if they're still needed in R, a one-liner on why R differs would help the next reader. Your call.

pkg-py/src/shinychat/_chat.py
```diff
@@ -833,7 +830,7 @@ def messages(
                 "Use your LLM provider (e.g., chatlas, LangChain) to manage conversation context instead."
             )
 
-        messages = self._messages()
+        messages = self._reported_messages()
```
gadenbuie, line 843: Worth surfacing that this makes `.messages()` eventually-consistent: it returns `()` until the client has reported, and a server-side `append_message` won't show up here until the browser renders it and reports back (previously `_store_message` made it synchronous). That's a reasonable consequence of client-authoritative state, but it is a public-API behavior change and I didn't see a CHANGELOG entry for it. Might be worth a note in the docstring + changelog so downstream users who read `.messages()` right after appending aren't surprised. Non-blocking.

pkg-py/src/shinychat/_input_handler.py
```diff
@@ -41,3 +50,40 @@ def _(value: Any, _name: "ResolvedId", _session: "Session") -> UserInputValue:
...
+                segments[0].html_deps = html_deps
+            attachments = [
+                Attachment.model_validate(a) for a in (m.get("attachments") or [])
+            ]
+            validate_attachments(attachments)
+            message = StoredMessage(
+                role=m["role"], segments=segments, attachments=attachments
+            )
+        except (KeyError, TypeError, ValueError) as e:
+            logger.warning("Skipping malformed message on shinychat.messages: %s", e)
```
gadenbuie: Since this snapshot is now the authoritative record for persistence, silently skipping a malformed message (warn + `continue`) could mean invisible data loss on save. Probably rare given the client controls the shape, but might be worth considering whether a louder failure (or a user-visible notification) is preferable to a log line here. Also note this diverges from the R handler (`chat_history_types.R`), which has no per-message guard and would raise on the same input — might be nice to align the two postures. Just a thought.

pkg-r/R/chat_history_types.R
```diff
@@ -1,3 +1,28 @@
+messages_input_value <- function(value) {
+  if (!is.list(value)) {
+    rlang::abort(paste0(
+      "Expected a list from shinychat.messages, got ",
+      class(value)[1]
+    ))
+  }
+  lapply(value, function(m) {
```
gadenbuie, line 8: For symmetry with the Python handler (`_input_handler.py`), which wraps per-message parsing in try/except and skips on error: this version accesses `m$role`/`m$segments` directly, so a single malformed message would abort the whole handler. Depending on how you land the "silent skip vs. loud failure" question on the Python side, it might be worth making the two behave the same way. Not urgent.

pkg-r/R/chat_history_store.R
```diff
@@ -191,7 +191,7 @@ safe_conv_path <- function(scope_dir, conv_id) {
   if (!grepl(CONV_ID_RE, conv_id)) {
     rlang::abort(paste0("Invalid conversation ID: ", conv_id))
   }
-  file.path(scope_dir, paste0(conv_id, ".json"))
+  file.path(scope_dir, conv_id)
```
gadenbuie, line 194: This changes the on-disk layout from one `{conv_id}.json` per conversation to a directory-per-conversation (`record.json` + `turns.jsonl` + `ui.jsonl`), and `list()`/`get()` now look for the new layout. If the previous R history format shipped in a release, conversations saved under the old layout would silently stop appearing — would a migration (or a documented break) be warranted? If R history is still unreleased, please disregard. Flagging just in case.

js/src/chat/ChatApp.tsx
```diff
@@ -101,6 +111,49 @@ export function ChatApp({
...
+  stateRef.current = state
+
+  const reportSnapshot = useCallback(() => {
+    transport.sendMessagesSnapshot(
+      elementId,
+      buildMessagesSnapshot(stateRef.current),
+    )
+  }, [transport, elementId])
+
+  useEffect(() => {
```
gadenbuie, line 128: This reports the *entire* settled transcript on every `state.messages` change (`buildMessagesSnapshot` serializes all messages plus their retained `htmlDeps`). For a long conversation that's roughly O(n²) bytes over a session, and the retained HTML deps could make each snapshot sizable. Probably fine for typical usage, but if you expect long transcripts it might be worth considering a delta/append protocol or at least noting the ceiling somewhere. Non-blocking.

js/src/chat/ChatApp.tsx
```diff
@@ -101,6 +111,49 @@ export function ChatApp({
...
+        type: "INPUT_SENT",
+        content,
+        role: "user",
+        ...(attachments.length > 0 ? { attachments } : {}),
+      })
+      // Build the snapshot from CURRENT settled state, then append the just-
+      // submitted user turn. Co-send userInput + snapshot in the SAME tick so
+      // Shiny batches them into one flush (server sees the turn in
+      // on_user_submit).
+      const snapshot = buildMessagesSnapshot(stateRef.current)
```
gadenbuie, line 145: If I'm reading this right, submit sends a snapshot here manually, and the `dispatch({ type: "INPUT_SENT" })` above also mutates `state.messages`, which triggers the `reportSnapshot` effect to send a second, near-identical snapshot the same tick. It looks correct (the downstream idempotency guard absorbs the duplicate), but relying on the guard to dedupe a self-inflicted double-send felt worth a comment — either a note explaining it's intentional, or dropping one of the two sends. Minor.

js/src/transport/shiny-transport.ts
```diff
@@ -68,8 +79,15 @@ export class ShinyTransport implements ChatTransport, ShinyLifecycle {
 
   sendInput(id: string, value: string | UserInputValue): void {
     if (!window.Shiny?.setInputValue) return
-    window.Shiny.setInputValue(`${id}:shinychat.userInput`, value, {
-      priority: "event",
+    const composite =
+      typeof value === "string" ? { text: value, attachments: [] } : value
+    this.inputSeq += 1
```
gadenbuie, line 84: Nice touch using the `seq` nonce so identical resubmissions still fire while co-batching with the snapshot. One thing worth calling out for future readers: dropping `priority: "event"` turns `input$<id>_user_input` into a persistent regular input rather than an event, which is a subtle semantics change for anyone reading that input directly. Blast radius looks small (it's bookmark-excluded), but a changelog line might save someone a debugging session later. Non-blocking.

gadenbuie (commented):
Really like the direction here — making the browser the source of truth is a clean way to collapse the R/Python divergence, and the Python side in particular reads as a genuine simplification (dropping `_store_message` and the replay-suppression flags for a count-based guard is a nice win).

Most of my comments are questions rather than requests, and I've left them inline. The one I'd most like your read on is a timing question on the R save path: Python's save is *triggered by* the client's message report (`@reactive.event(chat.messages)`), so the reported UI and the server turns are consistent by construction. R's save is chained onto the server stream-completion promise and then isolate-reads whatever the client last reported — which, in a real browser, I *think* is the snapshot from before this response, since the client hasn't rendered/reported the just-finished message yet. If that's right, the newest assistant message's `ui` may be empty at save time and restore would fall back to turn-derived markdown. I could easily be missing something about the ordering, so I've framed it as a question inline.

pkg-r/R/chat_history.R
cpsievert: Confirmed — you read the ordering exactly right, this was losing the last assistant message's UI on restore. Fixed in df49bf81: the save now keys off the `_messages` input (like Python) instead of stream completion, and only fires on a settled transcript ending in an assistant message. Added a regression test that drives the save through the echo (not a direct `on_response`) and asserts the displayed/transformed UI is stored. No `shinytest2` in the R package, so it exercises the observer via `testServer` rather than a real browser.

pkg-r/R/chat_history.R
cpsievert: They're not redundant: the count guard is steady-state idempotency, the flags guard the restore transaction (replay re-sends UI to the client, whose echo would otherwise trigger a save). Python can drop its flags because it doesn't re-send UI on restore. Kept both with a comment explaining why — c48feaa6.

pkg-py/src/shinychat/_chat.py
cpsievert: Good call — documented on the method and in the changelog in 8927ba61.

pkg-py/src/shinychat/_input_handler.py
cpsievert: Agreed. ef986c73 makes Python raise instead of warn-and-skip, matching R — now that the snapshot is the persistence record, a malformed message is a bug to surface, not silently drop.

pkg-r/R/chat_history_types.R
cpsievert: Aligned on loud failure: R already aborts, and Python now matches (ef986c73).

pkg-r/R/chat_history_store.R
cpsievert: R history is unreleased (dev version only), so no migration needed — thanks for flagging it in case.

js/src/chat/ChatApp.tsx
cpsievert: Real ceiling; deferring a delta protocol as more than typical usage needs. Noted it in a comment (6003ef1a) as the revisit path.

js/src/chat/ChatApp.tsx
cpsievert: Intentional — the manual send co-batches with userInput in one flush and the idempotent save absorbs the follow-up. Commented in 7612cbf7.

js/src/transport/shiny-transport.ts
cpsievert: Good call — changelog note in both packages (87368293), since R users read `input$<id>_user_input` directly.

js/src/chat/ChatApp.tsx
cpsievert: Opened #277 to track this as a follow-up — it captures the O(n²) reporting, why a delta protocol reintroduces server-side accumulation, and the options (append-delta, content-addressed messages, htmlDeps dedup).

PR #279, opened 2026-07-24
fix(pkg-r): faithfully restore displayed chat UI from bookmark, not re-derived turns

cpsievert:
## Why

#272 made the browser the authoritative source for what a chat currently
displays, so the server always saves the exact rendered snapshot the user
saw (transformations included) rather than re-deriving it from LLM turns.

`chat_restore()` never got the other half of that: on restore, it still
called `client_set_ui()`, which regenerates the UI straight from the
`client`'s turns. Any transformation applied between receiving a message
and displaying it (custom rendering, markdown post-processing, etc.) was
silently dropped on restore, even though the browser's own snapshot -
the thing #272 introduced - had it the whole time. Bookmark restoration
and live chat state also disagreed about what "the chat" looked like.

There was a second, related gap: the "bookmark on response" trigger fired
when the assistant's stream completed, but the browser echoes its settled
message snapshot back to the server in a later round trip. Bookmarking
on stream completion could persist a snapshot missing the reply that had
just finished, since the browser hadn't reported it yet.

## What changed

- `chat_restore()` now captures the browser's reported message snapshot
  into the bookmark's `state$values` (server bookmark store only - the
  gzip+base64 payload isn't URL-bookmark friendly) and replays that
  snapshot on restore, preserving display-only transformations exactly
  as shown.
- Falls back to the old turn-derived `client_set_ui()` path for
  `bookmarkStore = "url"` and for bookmarks saved before this snapshot
  existed, so nothing breaks for existing bookmarks.
- The "bookmark on response" observer now fires off the client's
  `_messages` echo (once it reports a settled transcript ending in an
  assistant message), instead of stream completion - mirroring the same
  fix #272 already made for history saves, applied here to bookmarks.
- Guards the response-bookmark trigger against firing during initial UI
  population (a fresh session's pre-seeded turns, or a restore itself) -
  only a genuine user-submitted exchange should mint an automatic
  bookmark.
- Fixes a separate cross-R-version bug in bookmark/UI-snapshot
  decompression, and removes bookmark-on-response code made fully dead
  by the trigger change above.
- Adds `encode_ui_snapshot()`/`decode_ui_snapshot()` (reusing the
  `serializeJSON` -> gzip -> base64 pipeline already used for client
  state) plus regression tests covering round-tripping, htmlDeps/
  attachments preservation, and the fallback path for pre-snapshot
  bookmarks.

## Test plan

- [x] New testthat coverage for snapshot encode/decode, faithful replay,
      and turn-derived fallback (`pkg-r/tests/testthat/test-chat_restore.R`)
- [x] `devtools::test()` / `devtools::check()` in `pkg-r/` — 599 passed, 0 failures


pkg-r/R/client_state.R
```diff
@@ -103,6 +103,45 @@ method(client_set_ui, S7::new_S3_class(c("Chat", "R6"))) <-
...
+    is.null(str) ||
+      !is.character(str) ||
+      length(str) != 1 ||
+      is.na(str) ||
+      !nzchar(str)
+  ) {
+    return(NULL)
+  }
+  json <- memDecompress(base64enc::base64decode(str), asChar = TRUE)
+  jsonlite::unserializeJSON(json)
```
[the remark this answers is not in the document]
cpsievert: Fixed in 76ff8154: `decode_ui_snapshot()` now wraps the decompress/deserialize step in `tryCatch`, returning `NULL` on any decode error instead of erroring. That lets `bookmark_restore_ui()` fall through to the turn-derived UI fallback (as already documented in `?chat_restore`) instead of aborting the whole `onRestore` handler on a corrupted payload. Added a regression test with a corrupted base64/gzip payload.

pkg-r/R/chat_restore.R
```diff
@@ -199,12 +202,26 @@ chat_restore <- function(
...
+      shiny::observeEvent(
+        session$input[[id_messages]],
+        label = "on_response_do_bookmark",
+        ignoreInit = TRUE,
+        {
+          if (messages_end_with_assistant(get_reported_messages(session, id))) {
+            session$doBookmark()
+          }
+        }
+      )
```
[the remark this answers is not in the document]
cpsievert: Fixed in 76ff8154: removed `chat_update_bookmark()`, `has_session_bookmark_on_response()`, `set_session_bookmark_on_response()`, and `ON_RESPONSE_KEY` from chat_restore.R, plus their call sites in `chat_append_stream()` (chat.R) and `markdown_stream()` (markdown-stream.R). Confirmed via grep that `set_session_bookmark_on_response()` had no remaining callers, so the flag could never be set and `chat_update_bookmark()` was a permanent no-op. Full test suite (596 tests) still passes.

pkg-r/tests/testthat/test-chat_restore.R
```diff
@@ -0,0 +1,165 @@
+test_that("encode/decode UI snapshot round-trips a simple message", {
```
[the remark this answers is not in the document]
cpsievert: Verified this doesn't actually break under R CMD check. `tests/testthat.R` runs `testthat::test_check("shinychat")`, which calls `test_dir(..., load_package = "installed")`. Traced through testthat's source: regardless of `load_package`, the environment each test file runs in comes from `testthat:::test_env(package)`, which does `env_clone(asNamespace(package))` — a clone of the package's full namespace, not the search-path attachment `library()` produces. A namespace always contains every object the package defines (exported or not); NAMESPACE-based export filtering only controls what's re-exposed to the search path, not what's inside the namespace itself.

Confirmed empirically:
- With the package actually installed and only `library(shinychat)` attached (no testthat), `encode_ui_snapshot(list())` throws `could not find function` — so it's genuinely internal, not accidentally exported.
- Running `test_dir("tests/testthat", filter = "chat_restore", load_package = "installed", package = "shinychat")` in a fresh process against the installed package (the exact code path `test_check()`/R CMD check uses) — all tests pass calling `encode_ui_snapshot`/`decode_ui_snapshot`/etc. unqualified.

This also matches the existing convention elsewhere in the suite (e.g. `test-chat_history_store.R` calls `InMemoryConversationStore()` directly, which isn't exported either). No `:::`/`getFromNamespace()` needed here.

pkg-r/R/chat_restore.R
```diff
@@ -236,51 +319,44 @@ chat_restore <- function(
...
-  # Bookmark has been flagged for `id`.
-  # When the stream ends, update the URL.
-  prom <-
-    promises::then(stream_promise, function(stream) {
-      # Force a bookmark update when the stream ends!
-      shiny::isolate(session$doBookmark())
-    })
-
-  return(prom)
+  state$values[[paste0(id, "_ui")]] <- encode_ui_snapshot(
```
gadenbuie, line 313: **[Blocking] Persists browser-controlled UI metadata that replays as script.**

`bookmark_save_ui()` serializes the raw browser-reported `_messages` snapshot, and on restore `restore_chat_ui()` → `restore_history_message()` forwards its `htmlDeps` unchanged (`chat.R:816`), where the client turns them into `<script>` elements appended to `document.head`.

Because a server bookmark is shareable via its `_state_id_` URL, an attacker can submit a forged `_messages` value in their own session, let `chat_restore()` mint the bookmark, and share the URL — the recipient's session then loads attacker-selected dependency scripts / raw HTML in the app origin during restore.

The `_messages` input handler (`messages_input_value()`) normalizes field names and validates attachments, but it does **not** establish a trust boundary for `htmlDeps` or `html` segments. Please reconstruct/allowlist dependencies from server-owned state (or otherwise ensure replayed HTML + deps cannot execute client-supplied code) rather than persisting them from an untrusted client input. Worth an adversarial test proving a forged snapshot cannot load a script.

pkg-r/R/chat_restore.R
```diff
@@ -192,19 +232,62 @@ chat_restore <- function(
...
+  # messages we populated ourselves (initial turn replay or restore) -- see
+  # has_user_submitted's comment above.
+  cancel_bookmark_on_response <-
+    if (bookmark_on_response) {
+      shiny::observeEvent(
+        session$input[[id_messages]],
+        label = "on_response_do_bookmark",
+        ignoreInit = TRUE,
+        {
+          if (!has_user_submitted) {
```
gadenbuie, line 250: **[Required] Response-bookmark guard order can eat the first real bookmark.**

The guard checks `!has_user_submitted` before `suppress_next_bookmark`. During the whole replay window `has_user_submitted` is `FALSE`, so we `return()` on the first line every time and **never consume `suppress_next_bookmark` during replay**. The stale flag leaks into the first genuine interaction.

Whether it does harm depends on echo ordering: if the first post-submit `_messages` echo ends in `"user"` it is consumed harmlessly, but if the user+assistant echoes coalesce into one update ending in `"assistant"`, suppress eats *that* one → no bookmark for the first assistant reply, which is exactly the miss this PR set out to prevent.

`HistoryController$on_response` (`chat_history.R:84-90`) checks `suppress_next_save` first with no `has_user_submitted` gate, so its flag is consumed by the post-replay echo as designed — this ordering is inverted relative to that. Per your own comment at `:159-169`, `has_user_submitted` is sufficient on its own; if so, dropping `suppress_next_bookmark` (and likely `is_replaying_ui`, never `TRUE` once `has_user_submitted` is) removes the misfire. Otherwise, please reorder so suppress is consumed during replay and document the scenario where `has_user_submitted` alone is insufficient.

pkg-r/R/client_state.R
```diff
@@ -103,6 +104,54 @@ method(client_set_ui, S7::new_S3_class(c("Chat", "R6"))) <-
...
+    return(NULL)
+  }
+  tryCatch(
+    {
+      json <- memDecompress(
+        base64enc::base64decode(str),
+        type = "gzip",
+        asChar = TRUE
+      )
+      jsonlite::unserializeJSON(json)
```
gadenbuie: **[Required] Syntactically valid but malformed snapshots still abort restore.**

`decode_ui_snapshot()` catches base64/gzip/JSON errors but does not validate the decoded value, so a well-formed-but-wrong payload slips through. E.g. `decode_ui_snapshot(encode_ui_snapshot(1L))` returns `1L`, and replay then fails on `message$role` rather than taking the documented turn-derived fallback. Please validate the transcript container and each message before replay, and route validation failures into the fallback. (Python's `_restore_bookmark_message()` is a useful parity target.)

Relatedly, `error = function(e) NULL` swallows genuine corruption silently — a `rlang::warn()` in the catch would make a lost faithful-restore diagnosable instead of a silent downgrade to turn-derived UI.

pkg-r/tests/testthat/test-chat_restore.R
```diff
@@ -0,0 +1,197 @@
...
```
gadenbuie, line 1: **[Required] Response-bookmark state machine is untested.**

Current tests cover the codecs and `restore_history_message()` replay, but never register `chat_restore()` and drive `_user_input` / `_messages` through a reactive session — which is exactly why the guard-ordering issue (the `!has_user_submitted` check running before `suppress_next_bookmark`) would not be caught. Please add observer-level coverage: startup/restore replay does not bookmark; a user-only snapshot does not response-bookmark; the first settled assistant response bookmarks exactly once; both `bookmark_on_input` settings; callback teardown.

pkg-r/R/client_state.R
```diff
@@ -103,6 +104,54 @@ method(client_set_ui, S7::new_S3_class(c("Chat", "R6"))) <-
...
 
+# Serialize the browser-reported message snapshot for storage in a bookmark's
+# state$values. Uses the same serializeJSON/base64 pipeline as
+# client_get_state() (see its comment for why, over toJSON()/fromJSON()).
+encode_ui_snapshot <- function(messages) {
+  if (is.null(messages) || length(messages) == 0) {
+    return(NULL)
+  }
+  json <- jsonlite::serializeJSON(messages)
+  base64enc::base64encode(memCompress(json, "gzip"))
```
gadenbuie: **[Optional] Duplicated serialize→gzip→base64 pipeline.**

The `serializeJSON → memCompress(gzip) → base64encode` pair (and its decode inverse) now lives here and in `client_get_state()`/`client_set_state()` (`:44-45`, `:68-72`). This is the exact pair the cross-R-version fix (`ec897ab`) had to touch twice; a shared `gzip_b64_encode`/`decode` helper would localize the next such fix.

gadenbuie (commented):
Thanks for closing the restore/transform gap left by #272 — the snapshot codec and the reuse of `restore_history_message()` are clean, and the `has_user_submitted` reasoning is well documented.

A few things I'd like to resolve before merge. The blocking one is a trust-boundary problem: a server bookmark now persists the browser-controlled `_messages` snapshot (including `htmlDeps`) and replays it verbatim into any session that opens the bookmark URL, which is a stored-script vector. The response-bookmark guard ordering and the malformed-snapshot fallback also need a look, and the new tests don't yet exercise the observer where the guard bug would surface.

cpsievert:
Splitting this into four PRs rather than merging it as one — it mixed an
urgent, narrow fix (markdown can smuggle shinychat's internal element names
into a raw-HTML sink) with a much larger feature (faithful UI restore on
bookmark) and a second security fix that turned out not to need the feature
at all. Closing this PR; your five comments carry over as threaded replies
below, each pointing at the PR that actually has the fix now:

- **#287** — the markdown-escaping fix, on its own, no bookmark code. Patch
  release candidate.
- **#288** — the client-report trust boundary (comment 1). Independent of the
  restore feature: it turns out this doesn't touch any of that feature's
  code, and it closes the same gap in the already-shipped
  `chat_enable_history()`, so there's no reason to gate it on the feature's
  review.
- **#289** — a small, unrelated Python fix (part of comment 3) that surfaced
  while working the rest of this.
- **#290** — the feature itself: faithful UI restore on `chat_restore()`
  bookmarks, plus your comments 2, 4, 5, and the R half of comment 3. Stacked
  on #287.

pkg-r/R/chat_restore.R
cpsievert: Both halves fixed, same shape of fix, same input boundary — and reworked once
more since the first pass. Now on #288 (`44d71933`, `b68e11e9`, `0fa88f24`).

**Deps**, in `44d71933`: `send_chat_action()` records every dependency it
sends, keyed by `name@version`. At the `_messages` input boundary each
reported dependency is replaced by the server's own copy of it, and anything
the server never sent is dropped. Nothing downstream — bookmark state, history
store, replay — ever sees a client-supplied dependency object again.

**Html segment content**, in `b68e11e9`/`0fa88f24`: the raw-html sink you
flagged (`RawHTML.tsx:26`, `el.innerHTML = html`) needed the same trust
boundary as deps, but there's no server-owned dependency object to substitute
here — the displayed string *is* the thing worth preserving, so it can't just
be swapped for a canonical copy.

The first pass (`b68e11e9`) recorded a hash for every *prefix* of a streaming
html run, because it couldn't predict where the client would close a segment.
On a second look that turned out to be defending a path that doesn't exist:
`buildMessagesSnapshot()` on the client filters out anything still streaming,
so the browser only ever reports a *settled* message — never an intermediate
prefix. `0fa88f24` replaces the prefix-hashing with a ledger that performs the
same merge the client's `chunk` reducer performs (append vs. replace,
content-type boundaries, `chunk_start`/`chunk`/`chunk_end`) and records the one
resulting string when the message settles. Same trust boundary, no more
`O(n²)` hashing, and a test now pins the invariant directly: an in-flight
prefix is asserted *not* trusted, only the settled merge is.

`messages_input_value()` degrades any reported `content_type: "html"` segment
that doesn't match the ledger to `"markdown"` — content untouched, so it
renders as inert literal text (the markdown branch escapes
`shinychat-raw-html`/`shiny-tool-request`/`shiny-tool-result` as of #287).

Also ported the same boundary to Python (`_html_trust.py`), which had the
identical gap — `messages_input_value()` there took `content_type` and
`htmlDeps` from the client verbatim into `StoredMessage` and on into the
history store. Wasn't in scope for your original comment since this PR never
touched Python, but it's the same sink, so it's part of #288 now.

Same adversarial test shape as before, in both languages: forge a `_messages`
value with attacker html, sanitize it through the input handler, feed it into
`restore_history_message()` (the primitive a history-conversation switch and
bookmark restore both call per stored message) into a fresh session, and
assert `"html"` never appears among the segment types — or a dependency
object — that reach the render/send path there. Added an end-to-end R/Python
pair that drives a *real* `chat_append()`/`append_message()` call through the
actual send path and feeds the wire content back through the input handler
the way the browser would, to check the ledger agrees with genuine server
output and not just with my model of it.

Two more bugs found while getting the invariant solid enough to build the
registry on, now on different PRs: `chat_ui(messages = list(div(...)))` was
rendering its `<shinychat-raw-html>` wrapper as literal text instead of the
actual content — fixed in #287, travels with the escaping fix rather than
the trust boundary. And `chat_ui(messages =)` content was being included in
the browser's reported snapshot even though it's static page markup, so a
bookmark restore rendered it twice — fixed in #290, since it only matters
once the restore feature exists to do the double-rendering.

Still open, deliberately not closed here: a forger with a *trusted* html
segment can still reorder it, duplicate it, or attach it to a different
message. Every byte reaching the sink is still server-authored, so this is
transcript spoofing, not script injection — closing it needs positional or
role binding per segment, which felt like its own PR.

pkg-r/R/chat_restore.R
cpsievert: Now on #290 (`b07e645c`).

Your read is right, including that `has_user_submitted` is sufficient on its
own. `is_replaying_ui` is only ever `TRUE` during startup or restore, which is
strictly before the first submission, so both it and `suppress_next_bookmark`
were unreachable-but-harmful. Both are gone, along with the `onFlushed`
bookkeeping that maintained them.

Confirmed the failure mode before fixing it: reintroducing the old ordering
fails three of the new observer tests, including the coalesced-echo case.

One thing worth flagging directly, since it's still open and not something
your comment covered: `has_user_submitted` only guards the *startup* replay
window. If an app pairs `chat_restore()` with `chat_enable_history()`,
switching conversations replays N messages and produces N `_messages` echoes
after the user has already submitted once this session — each ending in an
assistant message mints its own bookmark. Not proposing a fix for this PR;
flagging it as a follow-up (also noted in #290's description).

pkg-r/R/client_state.R
cpsievert: Split across #290 (`7feac78b` → simplified by `475c4601`) and #289 (`b37f7fa5`).

R side, #290: `decode_ui_snapshot()` initially grew a set of structural
predicates validating the transcript container and each message (role,
segments, and the optional `attachments`/`htmlDeps` fields) before returning
it, so `decode_ui_snapshot(encode_ui_snapshot(1L))` would take the
turn-derived fallback instead of failing on `message$role`. On review that
felt like validating a shape shinychat itself writes into its own bookmark
state dir — the only realistic producer of a wrong shape is version skew (a
bookmark URL outlives the shinychat that wrote it). `475c4601` replaces the
predicates with a version envelope on the snapshot, matching what
`client_get_state()` already does for the turns payload: one check instead of
a structural walk, same fallback, same warning.

Both failure modes (unreadable envelope, corrupted payload) still
`rlang::warn()` — agreed a silent downgrade is worse than a noisy one.

Python side, now its own PR (#289) rather than folded in here:
`_restore_bookmark_message()` was raising on a malformed message, which
aborts the whole restore loop and drops every message after the bad one too
(Shiny's generic `on_restore` error handling just shows a banner, it doesn't
resume the loop). Python has no turns-derived fallback to route into the way
R does, so the proportionate fix there is per-message: warn and skip just the
bad message, restoring everything else. Test seeds one malformed message
between two valid ones and asserts both valid ones still restore.

pkg-r/tests/testthat/test-chat_restore.R
cpsievert: Now on #290 (`b07e645c` + `pkg-r/tests/testthat/test-chat_restore.R`).

All of it observer-level through `shiny::testServer()` with `doBookmark()`
counted: startup/restore replay doesn't bookmark, a user-only snapshot
doesn't, the first settled assistant reply bookmarks exactly once, both
`bookmark_on_input` settings, `bookmark_on_response = FALSE`, and teardown.

The one that would have caught the guard bug drives the coalesced post-submit
echo specifically, since that's the only ordering where the miss is visible.

pkg-r/R/client_state.R
cpsievert: Now on #290 (`2891dd4e`).

One `gzip_b64_encode()`/`gzip_b64_decode()`, with the serializeJSON rationale
and the explicit `type = "gzip"` note (the thing an earlier commit on this
branch had to fix twice) documented once at the helpers.

PR #282, opened 2026-07-30
fix(markdown-stream): decide auto-scroll pinning before content grows

### LINKED ISSUE #2378: MarkdownStream auto-scroll can stall on Firefox: pinned state is derived only from async scroll events
## Summary

`test_auto_scroll_pinning[firefox]` failed again on the v1.7.0 release PR CI ([#2377](https://github.com/posit-dev/py-shiny/pull/2377), a docs-only diff), and passed on rerun.

Filing this as a **new** issue rather than reopening #2351, because the interesting part is not the test — it's that the failure signature is *byte-identical* to the pre-fix occurrence, which suggests the underlying product behavior, not test timing, is the thing left unaddressed.

**Job (attempt 1):** https://github.com/posit-dev/py-shiny/actions/runs/30419592356/job/90473449678 — `playwright-shiny (3.12, firefox, 2)`

```
FAILED tests/playwright/shiny/components/MarkdownStream/auto_scroll/test_stream_auto_scroll.py::test_auto_scroll_pinning[firefox]
  RuntimeError: .card-body:has(#auto_scroll_stream) never settled at bottom:
  {'scrollTop': 1998, 'scrollHeight': 2816, 'clientHeight': 258}
```

## Why this looks like more than a flake

| Occurrence | scrollTop | scrollHeight | clientHeight |
|---|---|---|---|
| #2351 (2026-07-09, **before** the #2355 fix) | 1998 | 2816 | 258 |
| This run (2026-07-29, **after** the #2355 fix) | **1998** | **2816** | **258** |

Identical geometry across three weeks and a landed fix. The container parks at exactly the pre-chunk-5 bottom — 560px short of the 2558px maximum (`scrollHeight - clientHeight`) — rather than at some arbitrary intermediate offset. A purely random timing wobble would be expected to produce varying values.

## Suspected root cause (from #2355's own analysis)

[#2355](https://github.com/posit-dev/py-shiny/pull/2355) diagnosed this precisely and then fixed only the test side:

> the vendored `markdown-stream.js` updates its pinned/unpinned state only inside its `scroll` event handler, and `updated()` auto-scrolls only when that flag says pinned. Scroll events are dispatched asynchronously — on Firefox from paint ticks, which can lag under CI load while Playwright's timer-based polls keep running.

That fix made the *test* wait for scroll-event delivery before streaming. It did not change the design where pinned state is knowable only via an async `scroll` event. If a paint tick is late in a real browser under real load, the same stall should be reachable outside tests: `ui.MarkdownStream` silently stops following new content while appearing pinned.

## Suggested direction

Rather than widening test waits again, consider computing pinnedness on demand from live geometry at the moment `updated()` needs it (e.g. comparing `scrollTop + clientHeight` against `scrollHeight` within a tolerance), using the `scroll` handler only as a hint. That removes the dependency on event delivery timing entirely.

## Notes

- Not a v1.7.0 regression — pre-existing, and unrelated to the release diff, which touched only `CHANGELOG.md` and `CITATION.cff`.
- Supersedes/relates to #2351 (closed as completed by #2355).
- Labeled `flaky test` so it is picked up by the staleness sweep, though the argument above is that the product path deserves a look independent of the flake.

schloerke:
Fixes posit-dev/py-shiny#2378.

## The bug

`useAutoScroll` recomputes `stickToBottom` **only** inside its `scroll` handler, and that handler reads geometry when the event is *delivered*, not when the scroll happened. Browsers dispatch scroll events asynchronously — on Firefox from a paint tick that can lag under load — so a content chunk can grow `scrollHeight` in the gap.

When it does, the user's at-bottom position reads as:

- not `isAtBottom` — the comparison is against the **grown** height
- not `isScrollingUp` — the user scrolled *down*

Neither branch fires, so `stickToBottom` stays `false`. Nothing else sets the flag and no further scroll events arrive, so the stream **silently stops following new content while still appearing pinned**. It never recovers for the life of the element.

This is the root cause of a py-shiny CI flake whose failure geometry was byte-identical across three weeks and an intervening test-side fix — the tell that the stall point was structural rather than a timing wobble.

## The fix

Add `repinIfAtBottom()` to `useAutoScroll` and call it from `MarkdownStream.appendContent` immediately before `setContent`. At that moment the DOM still holds the pre-growth `scrollHeight`, so "is the user at the bottom?" is unambiguous and needs no event.

It only ever **engages**, never disengages. That asymmetry is load-bearing: chunks routinely arrive while the previous chunk's smooth scroll is still animating, so "not at the bottom" at apply time does not mean the user scrolled away. Disengaging remains the scroll handler's job, where `isScrollingUp` is a genuine user-intent signal.

## Verification

Reproduced first, against py-shiny's existing `MarkdownStream/auto_scroll` app, by delaying delivery of shinychat's scroll callback by 400 ms (emulating a lagging paint tick) while leaving the test's own listener undelayed:

```js
window.__delayScroll = true;
const origAdd = EventTarget.prototype.addEventListener;
EventTarget.prototype.addEventListener = function (type, cb, opts) {
  if (type === "scroll" && window.__delayScroll && typeof cb === "function") {
    return origAdd.call(this, type, (...a) => setTimeout(() => cb(...a), 400), opts);
  }
  return origAdd.call(this, type, cb, opts);
};
```

| | `scrollTop` | `scrollHeight` | `clientHeight` |
|---|---|---|---|
| py-shiny CI failure (both occurrences) | 1998 | 2816 | 258 |
| Repro on `main` | **1998** | **2816** | **258** |
| Repro on this branch | **2558** (= bottom) | 2816 | 258 |

Also checked:

- `npm run test` — 691 pass, 0 fail; `npm run lint` clean.
- New unit tests: the late-scroll-event regression, the mid-animation guard (must not disengage), and the negative case (must not re-pin while the user is scrolled away).
- New `MarkdownStream` test asserting pinnedness is settled *before* the appended chunk reaches the DOM — ordering is the whole point, so it's worth pinning down.
- py-shiny's full `MarkdownStream` playwright suite passes on Firefox and Chromium against the rebuilt asset.
- Fast-stream regression check (200 chunks at 10 ms) ends flush at the bottom on both browsers, identical to the `main` baseline — the engage-only asymmetry holds up in a real browser.

`js/dist` + `pkg-py` / `pkg-r` assets are rebuilt and synced per `CLAUDE.md`. Changelog entries to follow in a second commit once this has a PR number.


cpsievert:
Thanks!

PR #283, opened 2026-07-31
feat: render tool calls as quiet, groupable rows

gadenbuie:
A single agentic turn that called three tools produced three full cards — header, arguments, result body, stacked. The transcript read as tool machinery with an answer buried somewhere in it. Tool activity is *subordinate* to the assistant's answer, and it should read that way.

Tool calls now render as a **quiet one-line row**: an identity glyph, the tool's title, a short peek at the result, and a chevron. Calls to the same tool in one contiguous tool loop collapse into a single row with a `×N` badge. Prose or thinking starts a new loop. Everything that used to be on screen is still one click away.

    Before                              After
    ┌────────────────────────────┐
    │ ▸ Searched the docs        │      ◇ Thought for 2s
    │   query: "net revenue"     │      · Searched the docs ×3            ▸
    ├────────────────────────────┤      · Read page: handbook.internal    ▸
    │ ▸ Searched the docs        │
    │   query: "discounts"       │
    ├────────────────────────────┤
    │ ▸ Read page                │
    │   url: handbook.internal…  │
    └────────────────────────────┘

Layered onto the existing request → result → streaming protocol, not a rewrite of it. **Both languages, one behaviour.**

## Three disclosure tiers

1.  **The resting row** — dimmed, `0.8em`, normal weight. Quiet by default.
2.  **The per-call list** — for a grouped row, revealed on expand.
3.  **The full card** — arguments and result body, exactly what shipped before.

A single-call row skips tier 2 and opens straight to the card. The card's own header is hidden in the morph, so the row *becomes* the header; fullscreen keeps it. Row space is reserved identically for the single and aggregated shapes, so a call landing and promoting a row to `×2` causes no layout jump.

## New API

**Chat-level**, on `chat_ui()` / `Chat.ui()`:

- **`tool_grouping`** — `"tool"` (default) groups calls to the same tool within a contiguous tool loop, order-independent rather than only consecutive; `"all"` groups every call in that loop; `"none"` gives each call its own activity row. It does not restore the old card stack. Prose and thinking split loops. It's also a **live custom-element attribute** — changing it re-routes the transcript already on screen with no server round-trip.
- **`icon_assistant=False`** removes the assistant icon and its streaming dots.

**Per-result**, on `ToolResultDisplay` / `tool_result_display()`:

- **`label`** — the per-call identifying value, rendered as `{title}: {label}`. A filename, a query, a URL.
- **`value_preview`** — a terse display-only peek: `"1,204 rows"`, `"3.1 kB · 200 OK"`.

**Per-tool**, as a definition-level annotation next to `title` / `icon`:

- **`grouping`** — a per-tool override of the chat-level mode. Grouping is a per-tool decision, not a per-result one. Chat-level `"none"` deliberately beats the annotation: it means grouping is *off*, so a tool can opt out of the chat's choice of how to group but can't opt back into grouping the app switched off.

R also gains an exported **`tool_result_display()`** constructor, matching Python's `ToolResultDisplay`. Bare named lists still work and are promoted internally.

## Fully custom tool UI now works end to end

Both packages already let an app render a tool result as *entirely custom UI* — R via a `contents_shinychat()` method, Python via `message_content()` / `message_content_chunk()` handlers. That bare return left **no `<shiny-tool-result>` in the transcript at all**, so nothing could tell the request had been answered and **the request row spun forever.**

shinychat now wraps that output in a real `<shiny-tool-result>` itself. **No author-facing API change** — existing apps of this shape are fixed retroactively. Detection is by artifact-marking rather than by shape, which is what preserves the extend-vs-replace distinction: an author who follows the documented `S7::super()` pattern (mutate `res$value`, return shinychat's own card) correctly reads as *not* custom.

While a custom call is pending, it remains in the activity row. Once it settles, that call leaves the row and its custom UI renders standalone below any remaining tool activity. Turning that bypass into a supported path is what let an imperative wire action be deleted: whether a request has been superseded is now **derived from content**, so the same lifecycle survives preload, bookmark restore, and history replay instead of depending on a signal that only exists live.

## Behaviour change worth calling out

**The client no longer conjugates tool titles.** The `"Running {title}"` / `"{title} failed"` template is gone. Present → past now comes entirely from two author-supplied strings: the tool's definition title (static, present tense) and an optional result title (dynamic, past tense).

The definition title appears while a call runs and remains the shared title for a multi-call group. For a settled single-call row, its result title replaces the definition title; within an expanded multi-call group, distinct result titles identify individual calls. A neutral definition title such as `"Weather Forecast"` is valid when no result-level title is supplied. Failures are a separate status cue — a `failed` / `N failed` note and icon — rather than text appended to the title. The count is always a separate `×N` badge, never inline, so **there is no pluralization or numeral-agreement API at all**; zero new phrasing strings ship here.

## How it works

Grouping is a **pure pre-render content router**, not React-tree grouping. One pass parses tool custom-element attributes out of the raw content string and rewrites the message into typed blocks, where a run of tool elements uninterrupted by prose or thinking becomes one `tool_loop` carrying a group model. The per-element bridges shrank to leaf renderers.

The reason it's shaped that way is **replayability**: history restore re-runs the same pure function over the same content and reconstructs identical grouping, independent of render or mount order. React-tree grouping is simpler but render-order-dependent, so restore and live streaming would disagree. Because the router is pure in `(blocks, grouping, role)` and a loop retains the raw slice it was parsed from, changing `tool_grouping` at runtime is just a re-route — which is where the live attribute comes from, for free.

## Documentation

- **R** — the canonical Tool Calling UI article now covers grouping, labels, value previews, title lifecycles, and custom displays. API references were updated for `tool_result_display()`, `contents_shinychat()`, and `chat_ui(tool_grouping=)`.
- **Python** — the API reference has a new Tool display section exposing `ToolResultDisplay`, `message_content`, and `message_content_chunk`.

## Tests

~5,500 added test lines, roughly 59% of the diff:

- **JS** — `js/tests/chat/ToolGroup.test.tsx` (+1,768) and large additions to `state.test.ts` and `ChatMessage.test.tsx`. 845 passing.
- **Python** — new `pkg-py/tests/pytest/test_tool_condensed_view.py` (+721), plus a `tools/custom_output` Playwright app.
- **R** — new `test-tool-custom-output.R` (+399) and substantial `test-contents_shinychat.R` additions.

A structural test pins the invariant that a **homogeneous group renders byte-identical markup to the pre-segment implementation**, so the multi-tool header work can't silently change the common case.

## Reviewing this

- **Skip `87f38c5`** — a mechanical `ruff format` pass touching 27 files, 25 of them unrelated to this feature. Verified formatting-only (identical ASTs once docstrings are normalized; the docstring diffs are `docstring-code-format`). Isolated in its own commit precisely so it can be skipped. `ruff check` had been passing independently, which is how the drift went unnoticed.
- **`2bc5cb7` and `46eaab0` cancel out.** A "thinking-lift" feature — hoisting a loop's reasoning into one merged row above its calls at `tool_grouping="all"` — was built, tested in a browser, and reverted the same day: hoisting rearranges the transcript into an order the model never produced, so a thought that plainly followed one tool and motivated the next appeared to precede both. Neither commit's code is in the final diff. Kept in history because the reasoning is worth having. Thinking now renders in its linear position at every mode and breaks the loop, so `"all"` aggregates only genuinely adjacent calls.

## Deliberately not in this PR

Each of these wants its own issue:

- **Serializing rendered UI into R's bookmark state.** R replays *turns*, never a snapshot, so it has neither an `html_deps` round-trip nor UI serialization. Includes the open question of whether R's **history replay** path — the default, and probably the higher-impact half — is covered by the same fix.
- **A server-side `update_tool_grouping()`** mirroring `update_cancel` / `update_upload`. The client half ships here; only the documented server API waits.
- **Hoisting the router to transcript scope**, which collides with \#269's sibling/edit navigation.
- **`prefers-reduced-motion` doesn't cover the tool chevrons** — a pre-existing one-line gap in `chat-tools.scss`.
- Plus: dropping `merge_ellmer_turn_group()`, unflattening `message_content(Turn)`, servers not emitting `content-type`, and a latent validation-escalation issue in Python's `ToolResultDisplay`.


js/src/chat/ToolResultBridge.tsx
```diff
@@ -1,11 +1,10 @@
-import { useContext, useEffect } from "react"
 import { ToolResult } from "./ToolResult"
-import { ChatDispatchContext } from "./context"
 
 interface ToolResultBridgeProps {
```
cpsievert, line 4: The four-layer split is really nice — `tool-presentation.ts` being pure policy with no React and no icon HTML is a pleasure to read.

One thing I'd like to understand better: the bridges are now a second rendering path for the same elements, and they've already drifted from the router — no `value-preview`, `grouping`, or `custom-display`, plus a private `isTruthy` duplicating `attrTruthy`.

I think the path is still live rather than vestigial: `markdownProcessor` passes raw HTML through (`allowDangerousHtml` + `rehypeRaw`, no `remarkEscapeHtml` — that's user-only), so any element the router declines becomes a real HAST node and hits the tag map. The case I'd expect in practice is a tag split across stream chunks, which seems plausible whenever `value` carries a whole HTML table. If that's right, the visible consequence of the missing `custom-display` is that a wrapped custom result renders as a normal tool card with the author's HTML in the body instead of migrating out.

I haven't built a repro, so you may know it's unreachable. Either way it'd help to write the reachability contract down next to `chatTagToComponentMap` — and if it is reachable, maybe the bridges should read the three new attributes (or deliberately render nothing, so the fallback is a known-empty state rather than a subtly wrong card).

pkg-r/R/contents_shinychat.R
```diff
@@ -162,6 +167,106 @@ new_tool_card <- function(type, request_id, tool_name, ...) {
...
+contents_shinychat_wrapped <- function(content) {
+  if (!S7::S7_inherits(content, ellmer::Content)) {
+    return(content)
+  }
+
+  wrap_custom_tool_result(content, contents_shinychat(content))
+}
+
 #' @export
 as.tags.shinychat_tool_card <- function(x, ...) {
```
cpsievert, line 271: `names(x) <- gsub("_", "-", names(x))` on line 291, just below, means any key in the card list becomes an attribute, so a field name the client doesn't read fails silently. That's the mechanism behind the `res$title` / `res$tool_title` doc bug fixed in e817f2da — the docs shipped a no-op for a release and nothing surfaced it.

Suggestion, non-blocking: validate the names against a known set and `cli_warn` on anything unrecognized. There's already exactly this pattern for display fields in `as_tool_result_display()` (around line 564), so it'd be consistent rather than new machinery — and it would have caught the original bug at the first `devtools::test()`.

js/src/chat/ChatMessage.tsx
```diff
@@ -85,6 +93,64 @@ export const ChatMessage = memo(function ChatMessage({
...
+    [message.streaming, message.blocks, message.role, toolGrouping],
+  )
+
+  // Tool UI is never legitimate in a user message, so don't hand the bridges to
+  // one. Defense in depth alongside the router's role gate: the router covers
+  // markdown (it just leaves the tags as text), but an html-typed user block
+  // skips the router's effect entirely and goes through `htmlProcessor` — no
+  // remarkEscapeHtml, no rehypeSanitize — so without this the tags would still
+  // resolve to real tool cards. Withholding the map leaves them inert elements.
+  const tagToComponentMap = isUser ? undefined : chatTagToComponentMap
```
cpsievert, line 117: Purely a note for later (take or leave): this now owns route-while-streaming, entity re-derivation, icon resolution, and empty-message suppression on top of the editing UI. Nothing wrong with any of it — just flagging that the two tool-specific `useMemo`s look like they'd lift out into a hook cleanly if it keeps growing.

pkg-py/src/shinychat/_chat_types.py
```diff
@@ -197,6 +191,20 @@ class ChatMessageDict(TypedDict):
...
+    # author's custom UI that bypassed it -- not declared here as an
+    # `__init__` parameter, since authors never construct it themselves.
+    #
+    # Typed as `Any` (not the precise `ToolResultComponent | None`) on
+    # purpose: this module has no chatlas dependency, and pyright checking a
+    # `setattr(msg, "_tool_result", ...)` call elsewhere against a pydantic
+    # `ToolResultComponent`-typed annotation forces resolution of that class's
+    # recursive `TagChild` alias from a foreign module context, which pyright
+    # cannot do (spurious `reportInvalidTypeForm` in `_chat_normalize_chatlas.py`).
+    _tool_result: Any
```
cpsievert: `_tool_result: Any` is a bare class-level annotation with no default, so it's typing-only — a normal `ChatMessage()` doesn't actually carry this attribute until `setattr()` marks it in `_chat_normalize_chatlas.py:522`. Every current read goes through `getattr(msg, "_tool_result", None)`, but any future direct access would raise `AttributeError`, and since it's typed `Any`, pyright won't flag the risk.

Also: only presence/absence is ever checked (no call site reads the stored value), so a whole `ToolResultComponent` is being carried around — plus an `Any` and a paragraph of comment about a pydantic recursive-type limitation — just to answer a yes/no question. The R side already has the idiomatic version of this exact check: `inherits(msg, "shinychat_tool_card")` in `contents_shinychat.R`. The natural Python equivalent is an empty marker subclass checked via `isinstance()`, not a monkey-patched private attribute — that also sidesteps the pyright issue entirely, since the marker class never needs to type the `ToolResultComponent` payload at all.

Separately, the comment just above this (`_chat_normalize.py:289`) claims `_chat.py` uses the presence of `_tool_result` — it doesn't; the only reader is `_wrap_custom_tool_result` in `_chat_normalize.py` itself. Worth fixing so it doesn't send someone chasing the wrong file.

Suggestion, non-blocking:

```python
# _chat_normalize_chatlas.py — carries no fields; existence of the type is the signal
class ShinyToolCardMessage(ChatMessage):
    pass

# _chat_normalize.py, in the ContentToolResult handler
cls = ShinyToolCardMessage if isinstance(result, ToolResultComponent) else ChatMessage
msg = cls(content=result)

# _wrap_custom_tool_result
if isinstance(msg, ShinyToolCardMessage):
    return msg
```

This drops the `_tool_result: Any` annotation (and its explaining-away comment) from `_chat_types.py` entirely, removes the `setattr`/`getattr(..., None)` dance, and keeps `_chat_types.py` — which has no chatlas dependency today — exactly as ignorant of chatlas as it already is, since the marker class lives alongside `ToolResultComponent` in `_chat_normalize_chatlas.py`.

cpsievert (commented):
The wire vocabulary is now written out four times independently: `tool-protocol.ts`, the two bridges, the Python `tagify()` methods, and R's free-form card list. Nothing ties them together, and each language tests its own half against hand-written strings.

Three drifts I noticed while reading, to make the case concrete: the R `tool_title` one (see the inline comment on `contents_shinychat.R`); `value-preview`/`grouping`/`custom-display` missing from the bridges (see the inline comment on `ToolResultBridge.tsx`); and Python emitting `expanded` on `<shiny-tool-request>` (`_chat_normalize_chatlas.py:106`) which nothing reads — `tool-protocol.ts`'s request branch doesn't parse it and `ToolRequest.tsx` has no such prop. That last one predates this PR, which is sort of the point: it's been dead for a while and there was no way to find out.

Cheapest fix I can think of: have Python and R each emit a canonical request+result pair to a checked-in fixture, and have a vitest case assert the parser round-trips that same file. One test catches all three, and there's precedent already — `pkg-r/tests/testthat/fixtures/history-behavior-matrix.json`. Happy to write it if you'd rather not expand the diff.

js/src/chat/ToolResultBridge.tsx
gadenbuie: Good catch. The bridges are reachable as a fallback for incomplete/unrouted assistant elements (and greeting/thinking Markdown), while complete elements are consumed by `routeToolBlocks()` first. I documented that contract beside the tag map in 95ba957. `grouping` and `value-preview` remain router-only because they have no truthful leaf-card behavior; for `custom-display`, the fallback now deliberately renders nothing until the complete router path can preserve standalone-payload semantics. The bridge and router also now share one presence-attribute truthiness helper, with focused fallback tests.

pkg-r/R/contents_shinychat.R
gadenbuie: We had actually litigated this locally when fixing `res$title`: the card object is a deliberately low-level extension point, and arbitrary fields provide an escape hatch for attributes such as `data-*` used by app JS. A closed allow-list here would make that deep path stricter without enforcing the cross-language contract, so I am going to pass on this safeguard. The substantive drift concern is addressed in 57976eb instead: R and Python now pin real serializer output in checked-in fixtures, and Vitest verifies both parse to the same semantics. Extra attributes remain open-world.

js/src/chat/ChatMessage.tsx
gadenbuie: Agreed on the threshold. I am leaving this in place for now: the memos still share message-rendering state and there is not yet duplication that makes a hook pay for itself. I recorded it as a refactor to take if this component grows further.

pkg-py/src/shinychat/_chat_types.py
gadenbuie: Done in 4bfccf3. `ShinyToolCardMessage` is now an empty marker subclass beside `ToolResultComponent`; `_wrap_custom_tool_result()` uses `isinstance()`, and `_tool_result: Any` plus the setattr/getattr dance and stale `_chat.py` comment are gone. I kept the `ToolResultComponent` narrowing/construction in a small helper inside `_chat_normalize_chatlas.py` so the recursive `TagChild` pyright issue stays contained. Existing complete/streaming/custom/display-none/legacy paths remain covered, and the full Python unit suite and pyright pass.

gadenbuie:
Thanks for the wire-contract read. I took the shared-fixture direction in 57976eb: R and Python each pin actual request/result serializer output, and Vitest verifies both fixtures describe and parse to the same semantics. This intentionally remains open-world for extra attributes rather than enforcing an attribute allow-list.

I also removed the stale Python request-side `expanded` capability in 5a277e6 and moved that field to `ToolResultComponent`; a targeted test proves it cannot leak onto request serialization. The field dated back to the old Lit shared-card inheritance and survived the React split.

The bridge fallback and Python marker points are handled in 95ba957 and 4bfccf3 respectively; details are in the inline replies.

pkg-py/src/shinychat/_chat_normalize.py
```diff
@@ -249,6 +348,80 @@ def _(chunk: Turn):
...
+
+        return isinstance(value, ContentToolResult)
+    except ImportError:
+        return False
+
+
+def _wrap_custom_tool_result(message: Any, msg: ChatMessage) -> ChatMessage:
+    """Wrap custom tool-result UI in a routable result element."""
+    if not _is_tool_result(message) or message.request is None:
+        return msg
```
cpsievert: I'd find this slightly easier to follow -- `_is_tool_result()` could also type narrow via `TypeGuard` (like it was before)

```suggestion
    if not _is_tool_result(message):
        return msg
    
    if message.request is None:
        return msg
```

pkg-py/src/shinychat/_chat_normalize_chatlas.py
```diff
@@ -383,6 +504,58 @@ def tool_result_contents(x: "ContentToolResult") -> Tagifiable:
...
+    # two shapes callers pass -- tags for `value_type="html"`, a plain string
+    # for every other mode, matching the split in `tagify()`.
+    value: Union[Tagifiable, str],
+    value_type: ValueType,
+    grouping: Optional[GroupingValue],
+) -> Tagifiable:
+    """Build the `<shiny-tool-result>` wrapper for an author's custom result UI.
+
+    Kept as a factory here so the caller only sees the opaque `Tagifiable`
+    return type.
```
cpsievert, line 540: Not sure I follow the reasoning here. Was this a conscious decision to keep this helper? I think I'd find it slightly easier to follow the if `ToolResultComponent()` was used directly as needed?

pkg-py/src/shinychat/_chat.py
```diff
@@ -350,7 +352,9 @@ def __init__(
 
         # Keep track of effects so we can destroy them when the chat is destroyed
         self._effects: list["Effect_"] = []
-        history_config = history if isinstance(history, HistoryOptions) else None
+        history_config = (
+            history if isinstance(history, HistoryOptions) else None
+        )
```
cpsievert, line 357: If it isn't a hassle, since this PR is pretty large, it'd be nice to avoid formatting only changes.

cpsievert (approved):
Nicely done! I have a few last nitpicks, but excited to see this land.

pkg-py/src/shinychat/_chat_normalize_chatlas.py
gadenbuie: Conscious call, yes. Two reasons to keep it as a factory here rather than constructing `ToolResultComponent` directly in `_chat_normalize.py`:

- `_chat_normalize.py` never has to know `ToolResultComponent`'s actual shape — if its fields change, only this module needs updating. It already treats `resolve_tool_annotations`, `tool_display_override`, `is_legacy`, and `ShinyToolCardMessage` the same way (guarded import from here), so this function is consistent with that boundary rather than an exception to it.
- The "wire minimal" invariants (`show_request=False`, `custom_display=True` — nothing renders for a migrated call) stay next to the model they configure, with the comment explaining why, instead of migrating that reasoning to the call site.

The counter-case is real too: it's called from exactly one site, so the indirection isn't buying reuse today. But I'd rather keep the layering than inline it for a single caller.

pkg-py/src/shinychat/_chat_normalize.py
gadenbuie: Done in 5c59fed: `_is_tool_result` now returns `TypeGuard[ContentToolResult]` (import guarded under `TYPE_CHECKING`), narrowing `message` for the rest of the function, and the combined guard is split into two early returns as you suggested.

pkg-py/src/shinychat/_chat.py
gadenbuie: Yeah, the hassle here is that the ruff version update caused formatting changes but our CI would have blocked the PR. I normally would rather do things separately but knew we'd need clean CI in this PR to move forward and I didn't have bandwidth/desire to wait for a separate PR to land. At this point, I'd much rather just land the formatting changes than move them around just for the sake of moving them

pkg-py/src/shinychat/_chat_normalize_chatlas.py
gadenbuie: Took a closer look at inlining this, and there's a concrete reason to keep it as a factory beyond style: pyright can't resolve this module's recursive `TagChild = TypeAliasType(...)` (line 37) when `ToolResultComponent` is *instantiated from a different module*. I confirmed this directly — even a trivial `ToolResultComponent(request_id="x", tool_name="y")` dropped into `_chat_normalize.py` produces `error: Variable not allowed in type expression` on the `TagChild` definition here, regardless of which fields are set. Calling it from within this module (as `tool_result_contents` already does) is fine; calling it from `_chat_normalize.py` isn't.

So the docstring's "caller only sees the opaque `Tagifiable` return type" wasn't just an abstraction preference — keeping construction inside this module is what avoids that resolution failure. I don't have a clean fix for the underlying pyright limitation, so I'm leaving this as a factory rather than inlining it into the one call site.

PR #295, opened 2026-08-10
Fix history persistence for tool-result htmlwidgets

### LINKED ISSUE #294: Conversation history fails to save when a tool result contains an htmlwidget
## Problem

When a tool result's `extra$display$html` contains an htmlwidget (e.g. a mapgl map), the widget renders correctly in the chat, but conversation-history persistence fails with a red toast:

> Could not save conversation: No method asJSON S3 class: htmlwidget

The save path (the `history_on_response` observer in `chat_history.R`, which calls `controller$on_response(get_turns_recorded(...))`) ultimately serializes the ellmer turns with `jsonlite::toJSON`, and the `ContentToolResult`'s `extra$display$html` holds the live htmlwidget object, which jsonlite cannot serialize.

## Repro

1. Register a tool that returns an interactive map:

```r
library(shiny)
library(bslib)
library(ellmer)
library(shinychat)
library(sf)
library(mapgl)

map_tool <- tool(
  function() {
    pts <- st_as_sf(
      data.frame(lon = runif(100, -125, -116), lat = runif(100, 32, 49)),
      coords = c("lon", "lat"), crs = 4326
    )
    map <- maplibre(style = carto_style("positron")) |>
      fit_bounds(pts, animate = FALSE) |>
      add_circle_layer(id = "pts", source = pts, circle_color = "#1b9e77")
    ContentToolResult(
      value = "Mapped 100 points.",
      extra = list(display = list(
        title = "Map", html = map, open = TRUE, full_screen = TRUE
      ))
    )
  },
  name = "show_map",
  description = "Show an interactive map of points."
)

ui <- page_fillable(chat_ui("chat"))
  client <- chat_anthropic()
  client$register_tools(list(map_tool))
  chat_server("chat", client)  # history = TRUE (default)
}
shinyApp(ui, server)
```

2. Ask: "Show me a map."
3. The map renders inline (great!), then the toast appears and the turn is not saved to history.

## Expected behavior

History persistence should either:

- Strip/skip non-serializable display content (e.g. drop `display$html` when it's an htmlwidget, or replace it with a placeholder) before saving, so the rest of the turn still persists, and/or
- Document that tools returning htmlwidgets are incompatible with `history = TRUE`, so app authors know to choose.

Ideally a restored conversation would show the tool card with a graceful fallback (title + text value) instead of failing the entire save.

## Environment

- shinychat 0.4.0.9000 (also reproduces against current `main` of posit-dev/shinychat — `chat_history.R` has no special handling for htmlwidgets)
- mapgl 0.5.0, ellmer (current), R 4.5.2, macOS

## Context

Hit while building a commons data agent whose custom `map_plots` tool returns mapgl htmlwidgets via `extra = list(display = list(html = ...))` — the htmlwidget-in-tool-card rendering works beautifully; only history save breaks.

gadenbuie:
Closes #294

## Summary
- Use object-safe JSON serialization when computing in-memory conversation metadata.
- Add a regression test for recorded tool results whose display HTML is a Leaflet widget.

## Verification
- `Rscript -e \"devtools::test()\"` (820 passing)
- `air format --check pkg-r/`

### DIFF
--- a/pkg-r/R/chat_history_store.R
+++ b/pkg-r/R/chat_history_store.R
@@ -175,7 +175,7 @@ InMemoryConversationStore <- R6::R6Class(
 )
 
 record_json_size <- function(record) {
-  as.double(nchar(jsonlite::toJSON(record, auto_unbox = TRUE), type = "bytes"))
+  as.double(nchar(jsonlite::serializeJSON(record), type = "bytes"))
 }
 
 CONV_ID_RE <- "^[A-Za-z0-9_-]{1,80}$"
--- a/pkg-r/tests/testthat/test-chat_history_store.R
+++ b/pkg-r/tests/testthat/test-chat_history_store.R
@@ -12,6 +12,42 @@ test_that("InMemoryConversationStore: put and get", {
   expect_equal(result$id, rec$id)
 })
 
+test_that("InMemoryConversationStore handles recorded HTML widgets", {
+  skip_if_not_installed("ellmer")
+  skip_if_not_installed("leaflet")
+
+  tool_result <- ellmer::ContentToolResult(
+    value = "Map created.",
+    extra = list(
+      display = list(
+        html = leaflet::leaflet() |> leaflet::addTiles()
+      )
+    )
+  )
+  recorded_turn <- ellmer::contents_record(
+    ellmer::UserTurn(contents = list(tool_result))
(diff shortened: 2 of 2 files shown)


cpsievert (approved):
Great, thanks!

PR #297, opened 2026-08-10
Fix Python tool-result display serialization

gadenbuie:
Related to #295.

## Walkthrough

A chatlas tool result can provide custom display HTML using the supported dictionary form:

```python
ContentToolResult(
    value="done",
    extra={"display": {"html": widget}},
)
```

The equivalent typed form, `ToolResultDisplay(html=widget)`, already serialized correctly because `ToolResultDisplay` converts htmltools objects into JSON-safe rendered HTML and dependency metadata. The dictionary form rendered correctly in the chat UI, but it bypassed that field serializer. When shinychat saved conversation history or bookmark state, `Turn.model_dump(mode="json")` encountered the live `htmltools.Tag` and raised `PydanticSerializationError`.

This has a similar symptom to #295, but it is orthogonal: #295 fixes R's in-memory conversation-size calculation, while this bug occurs earlier in Python when chatlas turns are converted to JSON-safe dictionaries.

The new shared serializer detects dictionary-form displays on `ContentToolResult`, copies the turn, normalizes the display through `ToolResultDisplay`, and then serializes the copy. Both conversation history and bookmarks use this path, and the live chatlas turn remains unchanged.

## Summary

- Serialize typed and dictionary-form tool-result displays consistently.
- Share the normalization path between conversation history and bookmarks.
- Add regressions for HTML dependencies, history metadata, and non-mutating dictionary normalization.
- Document the fix in the Python changelog.

## Verification

- `uv run pytest -q --ignore=pkg-py/tests/playwright` (400 passed)
- `uv run ruff format --check pkg-py/src/shinychat/_chatlas_serialization.py pkg-py/src/shinychat/_chat_bookmark.py pkg-py/src/shinychat/_history_client.py pkg-py/tests/test_history_client.py pkg-py/tests/pytest/test_bookmark_serialization.py pkg-py/tests/test_history_store.py --config pyproject.toml`
- `uv run pyright`
- `uv build`


pkg-py/src/shinychat/_chatlas_serialization.py
```diff
@@ -0,0 +1,30 @@
...
+        if not isinstance(extra, dict) or not isinstance(
+            extra.get("display"), dict
+        ):
+            continue
+
+        if normalized is turn:
+            normalized = turn.model_copy(deep=True)
+        normalized_content = normalized.contents[index]
+        normalized_content.extra = dict(normalized_content.extra)
+        normalized_content.extra["display"] = ToolResultDisplay(
```
cpsievert: This replacement can remove fields from the original display dictionary because `ToolResultDisplay` ignores unknown fields. Can we preserve the original dictionary and overlay the serialized known fields? That uses the `ToolResultDisplay` serializers without discarding application metadata.

pkg-py/tests/test_history_store.py
```diff
@@ -54,6 +55,35 @@ async def test_put_get_round_trip(store: FileConversationStore):
...
+        )
+    )
+    turn = chatlas.Turn(
+        role="user",
+        contents=[
+            chatlas.ContentToolResult(value="done", extra={"display": display})
+        ],
+    )
+    rec = new_conversation_record(title="Widget")
+    rec.append_linear([turn.model_dump(mode="json")])
```
cpsievert: I do not think this test protects the reported regression. It constructs a typed `ToolResultDisplay` and calls `turn.model_dump()` directly, so this path already worked before the change. Can we use a dictionary display and pass it through `TurnsAdapter`? Ideally, the test would also write through `FileConversationStore`, restore, and serialize again.

pkg-py/src/shinychat/_chatlas_serialization.py
```diff
@@ -0,0 +1,30 @@
+from __future__ import annotations
+
+from typing import Any
+
+
+def serialize_chatlas_turn(turn: Any) -> dict[str, Any]:
```
cpsievert: Since this function only accepts chatlas turns, can we type `turn` as `Turn[Any]` through a `TYPE_CHECKING` import? Using `Any` prevents Pyright from checking the assumptions about `contents`, `model_copy()`, and `model_dump()`.

cpsievert (changes requested):
Thanks for putting this together. I confirmed that the earlier fix only covered typed `ToolResultDisplay` instances. Dictionary displays still need normalization for both bookmarks and history, and this is the right layer for it.

The overall approach looks good. I left two substantive changes and one typing cleanup inline.

pkg-py/src/shinychat/_chatlas_serialization.py
gadenbuie: Implemented in ddc7a0e8. Dictionary displays are copied before known ToolResultDisplay fields are serialized over them, and the regression test verifies arbitrary application metadata is retained.

pkg-py/tests/test_history_store.py
gadenbuie: Implemented in e433dcea. The test now uses a dictionary display through TurnsAdapter, persists with FileConversationStore, restores it, and serializes it again.

pkg-py/src/shinychat/_chatlas_serialization.py
gadenbuie: Implemented in eca8db69. serialize_chatlas_turn now accepts Turn[Any] through a TYPE_CHECKING import, with an explicit ContentToolResult assertion after copying.

gadenbuie (commented):
Addressed the review comments @cpsievert can you re-review and approve? Thanks!

cpsievert (approved):

PR #298, opened 2026-08-10
support new ellmer API for grabbing the model name

simonpcouch:
Related to https://github.com/tidyverse/ellmer/pull/1053. Resolves a warning in the dev version of the package with dev ellmer:

```
Warning: Could not save conversation
Caused by error:
! Can't find property <ellmer::ProviderAnthropic>@model
```

`$get_model()` works for both CRAN and dev ellmer, so don't need any version-specific logic.


cpsievert:
Thanks!

PR #303, opened 2026-08-11
test(r): fix leaking warnings/messages in history & slash-command tests

gadenbuie:
## Summary
- Replace deprecated `bookmark_on_input`/`bookmark_on_response` chat_server() args in test-slash-commands.R with `history = FALSE`, and fix an invalid `title = "fallback"` value in test-chat_history.R (falls through to the same LLM auto-titling path as `"auto"`, which errored against the mock client) — both were leaking warnings into `devtools::test()` output.
- Add direct `lifecycle::expect_deprecated()` coverage for the `bookmark_on_input`/`bookmark_on_response` deprecations, so they're tested explicitly rather than as an incidental side effect.
- Add `shinychat.history_options.store_auto.quiet` option to silence the once-per-session "which storage backend was picked" messages from `resolve_store("auto")`, defaulting to quiet under `testthat`. Add snapshot tests pinning the exact message text.
- Narrow `@import S7` to `@importFrom S7 new_generic method "method<-" new_property S7_dispatch S7_inherits` — the blanket import pulled in `S7::validate()`, which shadowed `shiny::validate()` and triggered a "masked" warning on load. Fully qualify `S7::class_character` since it isn't in the new allowlist.

## Test plan
- [x] `devtools::test()` passes with 0 warnings (827 tests)

### DIFF
--- a/pkg-r/tests/testthat/test-chat_history.R
+++ b/pkg-r/tests/testthat/test-chat_history.R
@@ -353,7 +353,7 @@ test_that("on_response_saved fires on every response", {
   ctrl <- HistoryController$new(
     chat_id = "chat",
     client = client,
-    options = history_options(store = store, title = "fallback"),
+    options = history_options(store = store, title = NULL),
     session = session
   )
   ctrl$partition <- conversation_partition("chat", "test-user")
@@ -380,7 +380,7 @@ test_that("on_pre_switch returning TRUE skips the in-session swap", {
   ctrl <- HistoryController$new(
     chat_id = "chat",
     client = client,
-    options = history_options(store = store, title = "fallback"),
+    options = history_options(store = store, title = NULL),
     session = session
   )
   ctrl$partition <- conversation_partition("chat", "test-user")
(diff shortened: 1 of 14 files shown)


pkg-r/tests/testthat/test-chat_history.R
```diff
@@ -353,7 +353,7 @@ test_that("on_response_saved fires on every response", {
   ctrl <- HistoryController$new(
     chat_id = "chat",
     client = client,
-    options = history_options(store = store, title = "fallback"),
+    options = history_options(store = store, title = NULL),
```
gadenbuie, line 356: Flagging for confirmation while I was in here: `history_options(title = )` docs only recognize `"auto"`, a `function`, or `NULL`, but `HistoryController$initialize()` (R/chat_history.R:60-65) never validates that. Before this fix, this test passed `title = "fallback"`, which isn't any of the documented values — but since the code only checks `is.function(title)` / `is.null(title)`, it silently fell through to the same auto-titling path as `"auto"` and attempted a real LLM call against the mock client (causing the warning this PR removes). Should `history_options()` throw on unrecognized string values for `title` instead of silently treating them as `"auto"`?

cpsievert: > Should `history_options()` throw on unrecognized string values for `title` instead of silently treating them as `"auto"`?

Yeah, that seems sensible.

cpsievert (approved):

pkg-r/tests/testthat/test-chat_history.R
gadenbuie: Done in 361271b7423b5195177105baff9c92163e4c28c9

PR #308, opened 2026-08-13
Fix responsive layout for standalone asides

cpsievert:
## Why this matters

Standalone `<shiny-aside>` elements can become unusable when their labels are unusually long, particularly on narrow viewports or with increased text size. In the minimal reproduction, the trigger grows to 634 px tall at 200% text size, while the open popover is positioned 454 px above the viewport and its heading consumes 576 px before the body begins.

This is a general aside-component problem rather than citation-specific behavior.

## What changes

- Bound aside triggers to the available message width and a compact `18em` maximum.
- Keep trigger labels on one line with ellipsis instead of allowing pathological wrapping.
- Clamp popover headings to two lines so the body remains discoverable.
- Keep the navigation and source heading fixed while only the potentially long body scrolls.
- Enable Floating UI cross-axis shifting so tall popovers remain within the viewport.
- Add a standalone `<shiny-aside>` reproduction and Playwright coverage for desktop width, narrow layouts, 200% text size, truncation, heading height, and all four viewport boundaries.
- Rebuild and distribute the updated JavaScript and CSS assets to the Python and R packages.

## Before

At 320 x 480 and 200% text size, the trigger expands to 634 px tall. Opening it places the popover at `y = -454`, leaving most of its heading outside the visible viewport.

### Trigger at 200% text size

Before: a long aside label produces a trigger more than 600 pixels tall at 200 percent text size

### Open popover at 200% text size

Before: the popover begins above the viewport and its long heading consumes most of the visible panel

## After

The trigger remains a compact single line. The popover is constrained to the viewport at `(8, 8)`, its heading is limited to two lines, and the body begins within the visible panel.

### Trigger at 200% text size

After: the aside trigger remains a compact single line at 200 percent text size

### Open popover at 200% text size

After: the popover is fully contained within the viewport with a two-line heading and visible body

## Verification

- `cd js && npm run build`
- `cd js && npm test` - 972 passed, 23 skipped
- `uv run pytest pkg-py/tests/playwright/chat/aside_responsive pkg-py/tests/playwright/chat/web_asides -q` - 9 passed
- `uv run ruff check pkg-py/tests/playwright/chat/aside_responsive --config pyproject.toml`
- `uv run pyright`
- `git diff --check`


cpsievert:
@gadenbuie this should help with popovers that contain a lot of content. Feel free to comment or followup if you had something more particular in mind.

Issue #312, opened 2026-08-13
Reject unsupported conversation-history schema versions


gadenbuie:
## Problem

Python and R conversation-history records include `schema_version`, currently version 1, but neither package validates the value before using the record. The file stores continue decoding records with any version, and `put()` can rewrite an existing record without first checking the on-disk version. A package version that encounters a record written by newer code could therefore misinterpret or overwrite it.

## Required behavior

Implement the guard concurrently in the Python and R packages with equivalent behavior:

- Define one authoritative supported record-schema version per package (currently `1`).
- Validate a record schema before version-dependent decoding in every history read path. Missing `schema_version` must continue to mean version 1 for compatibility with existing records.
- Reject unsupported versions with a clear compatibility error that reports the encountered and supported versions. Do not treat an unsupported schema as a generic malformed record.
- Validate records returned by custom `ConversationStore` implementations before the history controller uses them.
- Make writes fail closed:
  - reject an input conversation record with an unsupported version;
  - when `record.json` already exists, preflight its schema before initializing write state, appending JSONL data, touching files, or replacing `record.json`;
  - never overwrite an existing record whose schema is unsupported.
- Apply equivalent validation to each package's in-memory store or shared store boundary so an unsupported record cannot bypass the file-store checks.

## Acceptance criteria

- Python and R `get()` and `list()` cannot decode a record with a schema newer than the package supports.
- Attempting to write a supported record over an existing unsupported record raises the compatibility error and leaves all files byte-for-byte unchanged.
- Attempting to write an unsupported record into an empty store fails without creating a conversation directory or files.
- A custom store returning an unsupported record is rejected before the record affects controller or client state.
- Existing version 1 records, including records without an explicit `schema_version`, continue to round-trip.
- Both package test suites cover unsupported reads, unsupported input writes, overwrite protection, custom-store ingress, and the legacy missing-version default.

## Out of scope

- Defining or migrating to a new record schema
- Changing the physical store layout
- General history-store refactoring

PR #313, opened 2026-08-13
fix(py,r): reject unsupported conversation-history schema versions

gadenbuie:
#312

## Summary
- Conversation-history records carry a `schema_version` field, but neither package validated it, so a record written by newer code could be silently misinterpreted or overwritten.
- Added `MIN_SCHEMA_VERSION`/`MAX_SCHEMA_VERSION` constants (both `1` for now) and a strict `check_schema_version()` helper in each package; missing/`None`/`NULL` still defaults to version `1` for backward compatibility.
- `FileConversationStore` validates disk records before decoding or writing, and both built-in stores reject unsupported input records before mutation. Rejected file writes leave the filesystem untouched.
- `HistoryController` validates full records returned by arbitrary custom `ConversationStore` implementations before they reach controller or client state.

## Test plan
- [x] Python: `uv run pytest`, `uv run pyright`, `uv run ruff check`
- [x] R: `devtools::test()`, `air format --check`, `devtools::check(document = FALSE)`
- [x] New tests cover unsupported schema on read/write, validation before version-dependent decoding, strict version types, byte-for-byte-unchanged rejected writes, custom-store ingress, and legacy missing-version defaults.

### DIFF
--- a/pkg-py/CHANGELOG.md
+++ b/pkg-py/CHANGELOG.md
@@ -33,6 +33,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * Fixed `MarkdownStream` permanently stopping following new content after the user scrolled back to the bottom. Pinning was decided only from `scroll` events, which browsers dispatch asynchronously; if a chunk grew the container first, the user's at-bottom position no longer read as at-bottom and auto-scroll silently disengaged for good. (#282)
 
+* `FileConversationStore` and `InMemoryConversationStore` now reject conversation records with an unsupported `schema_version` on read (`get`/`list`) and write (`put`), instead of silently misinterpreting or overwriting them.
+
 ## [0.6.0] - 2026-07-06
 
 ### New features
--- a/pkg-py/CHANGELOG.md
+++ b/pkg-py/CHANGELOG.md
@@ -33,8 +33,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * Fixed `MarkdownStream` permanently stopping following new content after the user scrolled back to the bottom. Pinning was decided only from `scroll` events, which browsers dispatch asynchronously; if a chunk grew the container first, the user's at-bottom position no longer read as at-bottom and auto-scroll silently disengaged for good. (#282)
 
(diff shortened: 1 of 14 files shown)


pkg-py/src/shinychat/_history_store.py
cpsievert, line 175: We don't have to do this here, but more so just calling it out that this should probably be a pydantic model

pkg-py/CHANGELOG.md
```diff
@@ -33,6 +33,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * Fixed `MarkdownStream` permanently stopping following new content after the user scrolled back to the bottom. Pinning was decided only from `scroll` events, which browsers dispatch asynchronously; if a chunk grew the container first, the user's at-bottom position no longer read as at-bottom and auto-scroll silently disengaged for good. (#282)
 
+* `FileConversationStore` and `InMemoryConversationStore` now reject conversation records with an unsupported `schema_version` on read (`get`/`list`) and write (`put`), instead of silently misinterpreting or overwriting them. (#313)
```
cpsievert: Maybe not worth mentioning?

pkg-r/NEWS.md
```diff
@@ -45,6 +45,8 @@
 
 * Fixed the copy button on code blocks not working in some embedded contexts. (@thisisnic, #247)
 
+* `FileConversationStore` and `InMemoryConversationStore` now reject conversation records with an unsupported `schema_version` on read (`get()`/`list()`) and write (`put()`), instead of silently misinterpreting or overwriting them. (#313)
```
cpsievert: Maybe not worth mentioning?

cpsievert (approved):
Thanks!

pkg-py/src/shinychat/_history_store.py
gadenbuie: Agreed. The persisted `record.json` shape deserves its own Pydantic model because its nodes contain `turn_ids` rather than the inline turns used by `ConversationRecord`. I kept that broader wire-format refactor out of this focused compatibility fix.

pkg-py/CHANGELOG.md
gadenbuie: Agreed, removed the Python changelog entry in 71098c27.

pkg-r/NEWS.md
gadenbuie: Agreed, removed the R NEWS entry in 71098c27.

Issue #276, opened 2026-07-13
R pkg: argument to control how errors are displayed in the chat


xx02al:
When an error occurs while streaming an assistant response (e.g. inside `chat$stream_async()`), [`chat_append_stream()`](https://github.com/posit-dev/shinychat/blob/e74ed266016038667a78ae8d1bf6a81089bb548f/pkg-r/R/chat.R#L790-L834) catches it and appends (via `chat_append_message()`) `sanitized_chat_error(reason)` into the chat UI as a normal chat bubble.

[`sanitized_chat_error()`](https://github.com/posit-dev/shinychat/blob/e74ed266016038667a78ae8d1bf6a81089bb548f/pkg-r/R/utils.R#L23) lives in `pkg-r/R/utils.R`, along with further helpers:

```r
needs_sanitized <- function(err) {
  isTRUE(getOption("shiny.sanitize.errors")) &&
    !inherits(err, "shiny.custom.error")
}

sanitized_error_message <- function(err) {
  if (needs_sanitized(err)) {
    "An error occurred. Please try again or contact the app author."
  } else {
    strip_ansi(conditionMessage(err))
  }
}

notify_error <- function(prefix, err) {
  shiny::showNotification(
    paste0(prefix, ": ", sanitized_error_message(err)),
    type = "error",
    duration = NULL
  )
  rlang::warn(prefix, parent = err)
}

sanitized_chat_error <- function(err) {
  if (needs_sanitized(err)) {
    sprintf("\n\n**%s**", sanitized_error_message(err))
  } else {
    sprintf(
      "\n\n**An error occurred:**\n\n```\n%s\n```",
      sanitized_error_message(err)
    )
  }
}
```

As we see, if `shiny.sanitize.errors` is explicitly set to `TRUE`, the chat shows the generic message. Otherwise, the chat bubble will show: `**An error occurred:**` followed by the raw `conditionMessage()`. A `conditionMessage()` text is whatever it is—it may be satisfactory, but chances are that it is not helpful, reveals confidential information, ...

`shiny.sanitize.errors` is a global option. If an app developer wants to scope it to a single shinychat function, or to otherwise take ownership of a specific error message shown by shinychat, things get complicated. Not because of a flaw in the implementation—`shiny.sanitize.errors` being read asynchronously, inside `chat_append_stream()`'s own `promises::catch()` handler, is absolutely fine as is.

## Proposal

Add an argument to the following functions, that allows app authors to define how the error is displayed in the chat:

[...]

PR #314, opened 2026-08-14
fix(r): report a response that fails before it streams anything

### LINKED ISSUE #304: A stream that fails before its first yield is dropped, leaving a spinner that never resolves
The error path in `chat_append_stream()` catches the rejection and appends `sanitized_chat_error(reason)` ([chat.R#L863-L887](https://github.com/posit-dev/shinychat/blob/64330fc2a703ea3bc005b221e42aace6389433ce/pkg-r/R/chat.R#L863-L887)). That only covers an error raised *after* `chat_append_stream_impl()` suspends at its first `await`. A provider that rejects the request outright fails before that, synchronously, so those handlers are never attached and nothing reaches the UI.

## Mechanism

The `ellmer::Chat$stream_async(stream = "content")` call returns a synchronous `coro_generator_instance`, and the request runs on its first advance — surfacing as a plain R error, not a rejected promise:

```r
client <- ellmer::chat_openai(
  base_url = "http://127.0.0.1:1/v1",
  model = "gpt-4.1-nano",
  credentials = function() list(Authorization = "Bearer x")
)

gen <- client$stream_async("hi", stream = "content")
gen()
#> Error in `req_perform_connection()`: Failed to perform HTTP request.
#> Caused by error in `open.connection()`: ! cannot open the connection
```

The consuming loop ([chat.R#L924](https://github.com/posit-dev/shinychat/blob/64330fc2a703ea3bc005b221e42aace6389433ce/pkg-r/R/chat.R#L924)) sits in the segment `coro::async()` runs eagerly, ahead of the first `await`. So the error propagates synchronously out of the *first* statement of `chat_append_stream()` ([chat.R#L856](https://github.com/posit-dev/shinychat/blob/64330fc2a703ea3bc005b221e42aace6389433ce/pkg-r/R/chat.R#L856)), skipping everything after it — both `promises::catch()` calls included. Tracing confirms it: `chat_append_message` is entered exactly once, for the `chunk = "start"`, and the `rlang::warn()` in the error handler never fires.

The error then lands in `append_stream_task`, whose result nothing reads — `on_stream_complete` branches only on `"success"`, and `status()` collapses `"success"` and `"error"` alike to `"idle"`, so a consumer cannot see it either:

```
task$status(): error | mod$status(): idle | result: httr2_failure: Failed to perform HTTP request.
```

The browser therefore keeps the loading bubble it raised on submit, with an empty `.shiny-chat-message-content` and `.shiny-chat-input` still carrying `disabled` — so the turn cannot be retried. Nothing later clears either: the `chunk = "end"` that would is in the handler that never runs.

This is the shape of every provider-side rejection — exhausted quota, over-long context, dropped connection, bad key. None yield a chunk, so none are the mid-stream case that is handled. A rejected turn is indistinguishable from a slow one, and the context-length case never recovers for the rest of the session.

## Reproducing

Nothing listens on port 1, so the first request is refused. Type anything and submit.

```r
library(shiny)
library(shinychat)

client <- ellmer::chat_openai(
  base_url = "http://127.0.0.1:1/v1",
  model = "gpt-4.1-nano",
  credentials = function() list(Authorization = "Bearer x"),
  echo = "none"
)

ui <- bslib::page_fluid(
  chat_ui("chat"),
  verbatimTextOutput("probe")
)

  mod <- chat_server("chat", client, history = FALSE)

  task <- get("append_stream_task", envir = environment(mod$append))

  output$probe <- renderText({
    err <- tryCatch(task$result(), error = conditionMessage)
    paste0(
      "task$status(): ", task$status(),
      " | mod$status(): ", mod$status(),
      " | result: ", err
    )
  })
}

shinyApp(ui, server)
```

Permalinks are pinned to 64330fc, which is what I ran; `main` is unchanged in all three respects ([chat.R#L991](https://github.com/posit-dev/shinychat/blob/main/pkg-r/R/chat.R#L991), [chat_app.R#L714](https://github.com/posit-dev/shinychat/blob/main/pkg-r/R/chat_app.R#L714)). Versions: shinychat 0.4.0.9000 built from source, ellmer 0.4.1.9000, R 4.5.3.

## Fix

Turning the synchronous throw into a rejection routes these into the handling that already exists, since everything downstream of that line is promise-shaped:

```r
result <- tryCatch(
  chat_append_stream_impl(id, stream, role, icon, session),
  error = function(e) promises::promise_reject(e)
)
```

Applied to `chat.R:856` in a source build of 64330fc and re-running the app above:

| | before | after |
| --- | --- | --- |
| assistant bubble text | `''` | `An error occurred:` + the httr2 message |
| `.shiny-chat-input` has `disabled` | `TRUE` | `FALSE` |
| `chat_append_stream()` warning | absent | fires |

Note it deliberately leaves `chat_append_stream()` returning the rejected promise, so `append_stream_task$status()` stays `"error"`. Reading `append_stream_task$result()` in `on_stream_complete` is therefore still worth doing on its own — it would catch anything else escaping the append path and let a consumer tell a failed turn from a finished one, which `status()` currently cannot.

Happy to open a PR for either or both if that's useful — just say which you'd prefer.

Related to #276, but distinct: that asks for control over how a displayed error renders, and presumes it is displayed. This class is not displayed at all.

nbenn:
Fixes #304.

The error path in `chat_append_stream()` never runs for a response that fails before it streams anything. Its stream is consumed inside a coroutine ahead of the first `await`, so the failure is raised synchronously rather than rejecting, skips both `promises::catch()` calls, and ends up in the `chat_server()` stream task whose result nothing reads. The browser keeps the loading indicator it raised on submit, with the composer disabled. That is the shape of every turn a provider rejects outright, since `ellmer::Chat$stream_async()` performs the request on the generator's first advance.

**Commit 1, `fix(r)`** — catch the synchronous throw and turn it into a rejection, which routes it into the handling already there. The error then renders through `sanitized_chat_error()` and the composer is released.

**Commit 2, `feat(r)`** — expose the condition as `last_error` on the `chat_server()` return, derived from the task rather than stored. The Python package already reads `_stream_task.error()` in an effect, so R was the outlier. Drop this commit if you would rather take the fix alone.

## Calls I made

Some judgement calls you might want to weigh in on:

- **The `last_error` value is additive, rather than making `status()` three-valued.** An `"error"` state is more honest but breaks callers testing `status() == "idle"` for completion. Exposing the task itself, as Python's `latest_message_stream` does, is a third option.
- **It stores nothing and re-raises nothing.** Python raises `NotifyException` from an effect; the R equivalent is an `observe()`, where a raised error takes the session down. Control over presentation belongs in #276.
- **A failed turn now warns twice** — Shiny's plus `chat_append_stream()`'s. Reading `$result()` does not mark the task handled. A mid-stream failure already does this, so the two classes now match.
- **The `last_error` test keeps a `MockShinySession` open** rather than using `testServer()`, which leaks an unhandled rejection when an `ExtendedTask` settles after teardown. Happy to use a house pattern if there is one.


pkg-r/tests/testthat/test-chat.R
```diff
@@ -327,3 +352,38 @@ test_that("chat_append_message() emits segment payloads incl. thinking", {
...
+  shiny::withReactiveDomain(session, {
+    expect_null(shiny::isolate(mod$last_error()))
+
+    suppressWarnings({
+      session$setInputs(chat_user_input = "hi")
+
+      while (is.null(shiny::isolate(mod$last_error()))) {
+        later::run_now(0.05)
+        session$flushReact()
+      }
```
gadenbuie: This wait loop has no upper bound, and testthat applies no per-test timeout. If a regression ever stops the rejection from reaching the task, `mod$last_error()` stays `NULL` and this spins until the CI job itself times out — an indefinite hang instead of a readable failure. Capping the iterations lets the existing `expect_equal(conditionMessage(...), "boom")` below fail cleanly instead:

```suggestion
      for (i in seq_len(200)) {
        if (!is.null(shiny::isolate(mod$last_error()))) break
        later::run_now(0.05)
        session$flushReact()
      }
```

pkg-r/tests/testthat/test-chat.R
```diff
@@ -182,6 +182,31 @@ test_that("chat_append_stream() returns the stream contents as list if not all t
...
+    })
+
+    p <- chat_append_stream("chat", stream())
+    expect_warning(
+      res <- tryCatch(sync(p), error = identity),
+      regexp = "chat_append_stream"
+    )
+
+    expect_promise(p, "rejected")
+    expect_equal(conditionMessage(res), "boom")
```
gadenbuie, line 223: Its sibling test (`chat_append_stream() handles errors in the stream`) asserts `expect_s3_class(res, "shiny.silent.error")`, but this one only checks the message. Since the fix's whole point is that the fail-before-yield path converges on the *same* handling as the mid-stream path, asserting the class here is what proves that convergence rather than just "some rejection happened." Consider adding it:

```suggestion
    expect_equal(conditionMessage(res), "boom")
    expect_s3_class(res, "shiny.silent.error")
```

gadenbuie (commented):
Two test-hardening notes on an otherwise solid fix, both minor. The `chat_append_stream()` change correctly routes the synchronous throw into the existing rejection handling, and the `last_error` wiring reads cleanly. Comments are inline.

pkg-r/tests/testthat/test-chat.R
gadenbuie: Superseded — see the new inline `suggestion` below. It bounds the loop with `flush_promises()`'s `deadline <- Sys.time() + timeout` idiom (keeping the `flushReact()` interleave this test needs) rather than an iteration cap, so it's more consistent with the existing helpers. Resolving this thread as outdated.

pkg-r/tests/testthat/test-chat.R
```diff
@@ -327,3 +352,38 @@ test_that("chat_append_message() emits segment payloads incl. thinking", {
...
+      session$setInputs(chat_user_input = "hi")
+
+      while (is.null(shiny::isolate(mod$last_error()))) {
+        later::run_now(0.05)
+        session$flushReact()
+      }
+    })
+
+    expect_equal(shiny::isolate(mod$status()), "idle")
+    expect_equal(conditionMessage(shiny::isolate(mod$last_error())), "boom")
```
gadenbuie, line 434: Bounds the wait loop so a regression fails cleanly instead of hanging (testthat has no per-test timeout, so an error that never lands spins until the CI job itself times out). This borrows `flush_promises()`'s `deadline <- Sys.time() + timeout` idiom while keeping the `flushReact()` interleave the test needs — which `sync()`/`flush_promises()` don't do — and adds an `expect_false()` guard so a missing error surfaces as a readable failure.

```suggestion
    suppressWarnings({
      session$setInputs(chat_user_input = "hi")

      deadline <- Sys.time() + 5
      while (is.null(shiny::isolate(mod$last_error())) && Sys.time() < deadline) {
        later::run_now(0.05)
        session$flushReact()
      }
    })

    expect_false(is.null(shiny::isolate(mod$last_error())))
    expect_equal(shiny::isolate(mod$status()), "idle")
    expect_equal(conditionMessage(shiny::isolate(mod$last_error())), "boom")
```

gadenbuie:
Weighing in on the "Calls I made" — this is a well-reasoned change, and `last_error` is the right shape. Notes below.

> * **The `last_error` value is additive, rather than making `status()` three-valued.** An `"error"` state is more honest but breaks callers testing `status() == "idle"` for completion. Exposing the task itself, as Python's `latest_message_stream` does, is a third option.

I think keeping it additive is the right call and it's not worth adding a new `"error"` state at this point. It's also consistent with a decision the R API already made: `status` is exposed as a derived scalar (`"streaming"`/`"idle"`), not the raw task, and `last_turn` already exposes the successful result. `last_error` completes that same decomposition — status / last_turn / last_error — rather than introducing a competing idiom.

**On the third option — exposing the task.** I took a look at this and landed on `last_error()` being the right call; recording why, mostly just for future reference:

- R's `ExtendedTask` public surface is `initialize` / `invoke` / `status` / `result`. `invoke` here is `function(client, ui_id, user_input, controller)` — handing back the raw task lets an app author fire a turn directly and bypass the whole `on_chat_user_input` orchestration (input tracking, the shared controller, the running-state guard, `pending_swap`). So the task can't be exposed as-is; it would need an `invoke`-stripped, read-only proxy.
- Once stripped, that proxy would expose `status()` + `result()`, which would collide with the existing `status` reactive (two vocabularies — `running/success/error` vs `streaming/idle`) and with `last_turn`. And `result()` would re-raise on error / `req(FALSE)` on idle, pushing reactive-only semantics onto every consumer. The coherent version of "expose the task" is really a *curated* handle (`status()`/`result()`/`error()`) — a new API to name and stabilize, not "surface the ExtendedTask."
- Crucially, `last_error` doesn't preclude that. Because the turn stays internally orchestrated, if a real need for the live handle shows up later (a partial-result reader, a cancel affordance), it can be added as another purpose-built accessor beside `last_turn`/`last_error` without breaking anyone. Exposing the task would have been the hard-to-reverse move.

So: `last_error` as implemented. 👍

> * **The `last_error` test keeps a `MockShinySession` open** rather than using `testServer()`, which leaks an unhandled rejection when an `ExtendedTask` settles after teardown. Happy to use a house pattern if there is one.

Good news! Driving a `MockShinySession` directly rather than `testServer()` *is* the house pattern. I left a comment above that fixes the timeout problem and uses more in-house helpers. 

> * **A failed turn now warns twice** — Shiny's plus `chat_append_stream()`'s. Reading `$result()` does not mark the task handled. A mid-stream failure already does this, so the two classes now match.

At this point, I'm okay with the duplication. If it ends up being annoying, we can chase down a resolution later.

pkg-r/tests/testthat/test-chat.R
nbenn: Applied. You are right that this is the assertion that carries the claim — "some rejection happened" would pass even if the two paths diverged, and convergence is the whole point of the change. Asserted now, matching the sibling test.

pkg-r/tests/testthat/test-chat.R
nbenn: Applied verbatim, including the `expect_false()` guard.

Checked that it does what it is for rather than assuming: with the client patched to succeed so the error never lands, the test now fails in a few seconds on `Expected is.null(...) to be FALSE` instead of spinning. Thanks for the pointer to the `flush_promises()` deadline idiom — I had not found it, which is also why I asked about a house pattern.

nbenn:
Rebased onto `main` to clear a `NEWS.md` conflict — both entries wanted the top of `## Bug fixes`. Kept yours from #275 unmodified and put this one above it, following the newest-landed-first ordering the section already uses. Say the word if you would rather they were the other way round.

Nothing else drifted: `air format --check` is clean, docs regenerate to the same output under the pinned roxygen, and the suite is green at 869. The one failure, `test-chat_history_integration.R:972` on `@finish_reason`, reproduces on a clean checkout of current `main` here too, so it is unrelated to this branch.

The workflow runs are sitting at `action_required` — they need an approval to start, since this is a first contribution from a fork.

gadenbuie (approved):
Thank you!

PR #315, opened 2026-08-14
Refine Python tool-result display serialization

cpsievert:
## Why stack this on #297

#297 correctly fixes dictionary-form tool displays bypassing `ToolResultDisplay`'s Pydantic serializers. This follow-up keeps that behavior and its persistence coverage, while moving the conversion to the JSON serialization boundary.

That distinction matters because a chatlas turn can contain application metadata beyond `extra["display"]`. Walking and copying the turn makes the generic persistence layer understand one shinychat schema; Pydantic's serialization fallback lets the turn remain opaque and only handles the htmltools values that are not otherwise JSON serializable.

## What changes

- Reuse one private htmltools serializer from typed display fields and dictionary-form turn metadata.
- Serialize bookmark and history turns directly with `model_dump(mode="json", fallback=...)`.
- Preserve sparse dictionaries, unknown display metadata, and the live turn.
- Require Pydantic 2.11+, where `fallback` became public API.

The helper deliberately preserves shinychat's current dependency dictionary format for this release. Its module documentation marks the intended follow-up: replace this temporary boundary with a source-preserving htmltools codec once that API is released.

## Verification

- `uv run pytest -q --ignore=pkg-py/tests/playwright` (429 passed)
- `uv run ruff format --check` for all changed Python files
- `uv run pyright`
- `uv build`

The repository-wide Ruff format check also reports a pre-existing formatting difference in `pkg-py/tests/playwright/chat/aside_responsive/test_aside_responsive.py` from the stacked base.


pkg-py/src/shinychat/_chat_bookmark.py
```diff
@@ -87,7 +87,10 @@ async def get_state() -> Jsonifiable:
         turns: list[Turn[Any]] = client.get_turns()
         return {
             "version": 1,
-            "turns": [serialize_chatlas_turn(turn) for turn in turns],
+            "turns": [
+                turn.model_dump(mode="json", fallback=serialize_htmltools)
```
gadenbuie: Non-blocking: this `turn.model_dump(mode="json", fallback=serialize_htmltools)` expression is duplicated here and in `_history_client.py:41`. The removed `_chatlas_serialization` module used to centralize how a chatlas turn becomes the wire format; inlining it in two places lets the `mode=`/`fallback=` contract drift between call sites.

Consider a small shared helper — e.g. `serialize_chatlas_turn(turn)` in `_chat_bookmark.py`, next to `is_chatlas_chat_client` (which `_history_client` already imports from):

```python
def serialize_chatlas_turn(turn: Turn[Any]) -> dict[str, Any]:
    return turn.model_dump(mode="json", fallback=serialize_htmltools)
```

That keeps `_htmltools_serialization` a chatlas-agnostic codec and restores the single boundary.

gadenbuie (approved):
Looks good! Just one small comment to create a shared helper so we can more easily coordinate changes in this area in the future.

Also, to record our thoughts from our Slack conversation, we'll merge this and #297 into main, take #297 into a maintenance v0.6.1 and land #316 separately when htmltools is ready.

pkg-py/src/shinychat/_chat_bookmark.py
gadenbuie: Went ahead and did this in febe6e83 — added a `serialize_chatlas_turn` helper in `_chat_bookmark.py` and pointed both call sites at it.

PR #317, opened 2026-08-14
Prepare shinychat for RenderedHTML value objects

cpsievert:
## Why this matters

posit-dev/py-htmltools#123 changes `RenderedHTML` from a structural `TypedDict` into the dict-compatible value object returned by `render()`. Shinychat currently creates two plain dictionaries under that annotation, which makes downstream Pyright fail against the proposed htmltools release.

Using the public constructor keeps shinychat compatible with both the current htmltools `TypedDict` constructor and the proposed value object, so this can land independently before htmltools is released. Runtime output is unchanged.

## Verification

Verified from a fresh `uv sync --all-extras` environment after installing the wheel built from posit-dev/py-htmltools#123:

- `uv run --no-sync pyright`
- `uv run --no-sync pytest -q --ignore=pkg-py/tests/playwright` (424 passed)
- `uv run --no-sync ruff check pkg-py --config pyproject.toml`
- Changed-file Ruff format check
- Source distribution and wheel build

### DIFF
--- a/pkg-py/src/shinychat/_chat_normalize_chatlas.py
+++ b/pkg-py/src/shinychat/_chat_normalize_chatlas.py
@@ -210,10 +210,10 @@ class ToolResultComponent(ToolCardComponent):
         if self.value_type == "html":
             value_ui = TagList(self.value).render()
         else:
-            value_ui: "RenderedHTML" = {
-                "html": str(self.value),
-                "dependencies": [],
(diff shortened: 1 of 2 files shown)


cpsievert:
Superseded by #315: the merged tree already includes the two  constructor changes needed for py-htmltools#123.

PR #323, opened 2026-08-17
fix(pkg-r): preserve filename on PDF attachments

### LINKED ISSUE #321: PDF attachments lose the user's filename before reaching ellmer
In the R package, a PDF attachment's original filename is discarded before the content object is built, so the model receives a synthetic name like `file_001.pdf`. The Python package preserves the real filename. This is a cross-language behavior difference that affects live conversations, not just persistence.

### Cause

`content_from_attachment()` in `pkg-r/R/attachments.R` computes the attachment name and uses it for the text-attachment branch, but does not pass it to the PDF branch:

```r
content_from_attachment <- function(att) {
  mime <- att[["mime"]]
  data_url <- att[["data_url"]]
  name <- att[["name"]] %||% ""      # computed...

  if (startsWith(mime, "image/")) {
    return(ellmer::content_image_url(data_url))
  }
  if (identical(mime, "application/pdf")) {
    return(ellmer::content_pdf_url(data_url))   # ...but not used here
  }
  # ...text branch does use `name`
```

`ellmer::content_pdf_url()` has no filename parameter, so it substitutes `unique_pdf_name()`, a sequential `sprintf("file_%03d.pdf", counter)`. The real name is destroyed before the `ContentPDF` object exists — it isn't merely dropped at serialization time.

The Python path does the opposite, in `pkg-py/src/shinychat/_attachments.py`:

```python
if att.mime == "application/pdf":
    _require_data_url(att)
    data = decode_data_url(att.data_url)
    return ContentPDF(data=data, filename=att.name or "document.pdf")
```

### Reproducible example

```r
url <- paste0(
  "data:application/pdf;base64,",
  jsonlite::base64_enc(charToRaw("%PDF-1.4 fake"))
)

ellmer::content_pdf_url(url)@filename
#> [1] "file_001.pdf"
ellmer::content_pdf_url(url)@filename
#> [1] "file_002.pdf"
```

### Impact

The filename is part of the prompt context. A user who uploads `Q3-earnings.pdf` and asks "summarize the attached earnings report" gives the model a useful cue in Python and `file_001.pdf` in R. It also means the same app, ported between languages, sends the model different information.

### Suggested fix

Construct `ContentPDF` directly rather than going through `content_pdf_url()`, mirroring the Python branch. Verified this works:

```r
p <- jsonlite::base64_enc(charToRaw("%PDF-1.4 fake"))
ellmer::ContentPDF(
  type = "application/pdf",
  data = p,
  filename = "Q3-earnings.pdf"
)@filename
#> [1] "Q3-earnings.pdf"
```

A fallback of `"document.pdf"` when the name is empty would match Python exactly.

### Note

Image attachments lose their filename in **both** languages, because `ContentImageInline` has no filename field at all. That's an upstream limitation rather than a shinychat bug, so it's out of scope here — mentioning it only so the asymmetry isn't surprising.

gadenbuie:
Fixes #321

## Summary
- `content_from_attachment()` in `pkg-r/R/attachments.R` computed the attachment `name` but only used it for text attachments; the PDF branch went through `ellmer::content_pdf_url()`, which has no filename parameter and substitutes a synthetic `file_NNN.pdf` name.
- Now constructs `ellmer::ContentPDF()` via `content_pdf_url()`, and then overwrite's ellmer's generated filename (keeping ellmer's generated file name when empty).
- Extended `test-attachments.R` to assert the PDF content's `@filename` matches the original name, and falls back to ellmer's `"file_NNN.pdf"` when the name is empty.


pkg-r/R/attachments.R
```diff
@@ -173,7 +173,15 @@ content_from_attachment <- function(att) {
     return(ellmer::content_image_url(data_url))
   }
   if (identical(mime, "application/pdf")) {
-    return(ellmer::content_pdf_url(data_url))
+    comma <- regexpr(",", data_url, fixed = TRUE)
```
gadenbuie: actually, on second look, maybe this belongs in ellmer?

gadenbuie: after looking at ellmer... _maybe_ but probably not worth it.

From ellmer's perspective, only OpenAI uses the filename. It's not enough to warrant coordination with versioning, so we'd have to keep our own implementation in shinychat until we have a better reason to bump ellmer.

So I just went with a simpler version that re-uses `content_pdf_url()`'s input validations but then overwrites `@filename` before returning.

pkg-r/NEWS.md
```diff
@@ -37,6 +37,8 @@
 
 ## Bug fixes
 
+* Fixed PDF attachments losing the user's original filename. `content_pdf_url()`'s synthetic filename is now overridden with the user's original filename when one was provided, matching the existing behavior in Python. (#323)
```
cpsievert: I think this could be omitted given that we haven't shipped file attachments?

cpsievert (approved):

PR #329, opened 2026-08-20
feat: add page chat layouts, navigation, and drawer

gadenbuie:
## Summary

This PR adds a first-class full-window chat application layout for shinychat. `page_chat()` makes chat the primary application surface while preserving `chat_ui()` for embedded or mixed layouts.

### Desktop layout

https://github.com/user-attachments/assets/4c02115a-7f0b-4600-9a13-1245a79063fe

### Mobile layout

https://github.com/user-attachments/assets/37138596-6216-45b3-9162-3539dcc65470

### Compose a page

`page_chat()` registers secondary application pages through `pages`:

```python
page_chat(
    "Assistant",
    sidebar=chat_sidebar(...),
    toolbar=...,
    toolbar_global=...,
    pages=[
        chat_nav_panel("Settings", ..., sidebar=chat_sidebar(...), toolbar=...),
        chat_nav_panel("About", ..., sidebar=False),
    ],
)
```

In R, pass the same `chat_nav_panel()` configurations as `pages = list(...)`.

- `page_chat(sidebar=)` configures the chat home sidebar. `True` uses the default conversation-history sidebar, `False` removes it, and `chat_sidebar(...)` supplies home-specific content, history, width, initial open state, and desktop resizing.
- `chat_nav_panel(sidebar=)` configures the selected secondary page's sidebar. It can omit the sidebar, reuse the default history sidebar, or replace it with page-specific `chat_sidebar(...)` content.
- The home sidebar and each panel sidebar are independently configured. A page must opt into the default history sidebar or supply its own sidebar; this avoids showing stale home controls on a settings or informational page.
- `page_chat(toolbar=)` supplies actions scoped to the chat home. `chat_nav_panel(toolbar=)` replaces that scoped toolbar for its panel; `toolbar=True` reuses the home toolbar and `toolbar=False`/`None` omits it.
- `page_chat(toolbar_global=)` supplies persistent actions that remain mounted and visible on every page. This is appropriate for actions such as help, account controls, or application-wide settings.

`chat_nav_panel()` and `chat_sidebar()` are shinychat configuration objects rather than generic Shiny navigation/sidebar components because the page shell must coordinate one persistent chat root, per-page sidebar/toolbar metadata, responsive menu placement, history ownership, resizing, and accessible navigation. On mobile, the same controls are moved into the app menu rather than duplicated, preserving Shiny input/output IDs and state.

### New public API

- Add `page_chat()` and `page_chat_theme()` for Shiny for Python and Shiny for R, plus a Shiny Express `page_chat()` adapter.
- Add `chat_nav_panel()`, `chat_sidebar()`, and `chat_ui_history()` for page navigation and externally mounted conversation history.
- Add `chat_drawer()` to configure an adjacent drawer with initial content, title, width, visibility, and resizing.
- Add server-side drawer controls:
  - Python: `await chat.drawer.show()`, `.update()`, `.hide()`, and `.toggle()`.
  - R: `chat_drawer_show()`, `_update()`, `_hide()`, and `_toggle()`.

`chat_ui()` also gains `drawer` and `show_history` options, making drawers and externally mounted history available outside `page_chat()`.

### Interaction and lifecycle

- The chat home remains mounted while a `chat_nav_panel()` is selected, preserving conversation state, streaming work, composer draft, scroll position, and drawer state.
- A shared per-chat history store lets the page shell and chat root consume restored history without competing ownership.
- Drawers are adjacent to chat on sufficiently wide containers and become a focused full-width takeover view in narrow containers. Responsive behavior follows container width, so embedded chats behave correctly too.
- `show()` can update and reveal a drawer atomically; `update()` preserves visibility; `hide()` preserves content. Dynamic drawer UI processes dependencies and unbinds/rebinds Shiny inputs safely.
- `page_chat_theme()` provides a composable baseline for page surfaces, typography, density, safe-area insets, and theme tokens. Supplying an ordinary Shiny/bslib theme intentionally bypasses that baseline.

### Compatibility and intentional constraints

- `page_chat()` owns one full-window chat root, history placement, navigation, and the responsive app menu. Do not wrap it in another page/layout container or add unrelated top-level UI; use `chat_ui()` for those layouts.
- `height`, `fill`, and `show_history` are page-owned and cannot be overridden through `page_chat()`.
- `navbar_options` supports title-bar styling but rejects `position` and `collapsible`, which conflict with the fixed shell and responsive menu.
- Python accepts only `chat_sidebar()` configurations. R also normalizes compatible left-positioned `bslib::sidebar()` values; right-positioned bslib sidebars remain unsupported.
- Drawer content is application-supplied UI and may contain ordinary Shiny inputs and outputs. Conversation changes deliberately do not clear drawers automatically.
- R `chat_app()` now composes `page_chat()` with `chat_server()`. Its `...` now configures the page rather than `shinyApp()`; use `app_options`, `bookmark_store`, or direct composition for the previous `shinyApp()` controls.
- The Python package now requires Shiny >= 1.6 and `libsass`; the R package requires bslib >= 0.11.

## Documentation and verification

- Added Python Core and Express guides, R reference/docs, API reference pages, and credential-free navigation and drawer-control examples for both packages.
- Added JS unit coverage plus Python and R unit, snapshot, and browser coverage for page layouts, history, drawers, resizing, navigation, and responsive behavior.
- Added two example apps, mirrored in both packages, that can be run with `shiny run shinychat.examples.page_chat.navigation:app` or `shiny::runExample("page-chat-navigation", package = "shinychat")` to exercise navigation, sidebars, history, mobile menu behavior, and drawer controls. The second example is `shinychat.examples.page_chat.drawer_control` or `page-chat-drawer-controls`.


docs/bslib-resize-handle-proposal.md
cpsievert, line 1: Was adding this file intentional?

pkg-py/src/shinychat/_page_chat.py
```diff
@@ -0,0 +1,1147 @@
...
+    navbar_options: Any = None,
+    sidebar: bool | ChatSidebar = True,
+    artifact: bool | ChatArtifact = True,
+    window_title: str | None = None,
+    lang: str | None = None,
+    theme: Any = None,
+    messages: Optional[
+        Iterable[str | TagChild | "ChatMessageDict" | "ChatMessage" | Any]
+    ] = None,
+    greeting: Optional[Union[str, HTML, Tag, TagList, "ChatGreeting"]] = None,
```
cpsievert, line 97: I've been having feelings lately along the lines of "`messages` aren't needed now that we have both greetings and history". Does that resonate with you?

pkg-py/src/shinychat/_page_chat.py
```diff
@@ -0,0 +1,1147 @@
...
+
+
+def page_chat(
+    title: TagChild,
+    icon: TagChild | None = None,
+    *,
+    id: str = "chat",
+    pages: Sequence[ChatNavPanel] | None = None,
+    toolbar: TagChild | None = None,
+    toolbar_global: TagChild | None = None,
```
cpsievert: I find some of the global vs local stuff in this PR hard to wrap my head around. 

What if we went in a direction where everything at the top-level `page_chat()` gets inherited on each page (i.e., global by default), then you opt-out at with `chat_nav_panel(inherit_toolbar = F, inherit_sidebar = F)`, etc. This way, I think you could also de-couple history from sidebar (i.e., `inherit_history = F`), which I also found a bit confusing.

pkg-r/R/page_chat.R
```diff
@@ -0,0 +1,1051 @@
...
+#' @param ... UI content to display when the panel is active.
+#' @param value An optional unique navigation value.
+#' @param icon An optional icon to display with the title.
+#' @param sidebar Whether to use the default sidebar (`TRUE`), no
+#'   page-specific sidebar (`FALSE`), or a [chat_sidebar()] or
+#'   [bslib::sidebar()] configuration.
+#' @param toolbar `NULL` (the default) for no page-scoped toolbar, or UI
+#'   content for a page-specific toolbar. For backward compatibility, `TRUE`
+#'   reuses the home `page_chat()` toolbar and `FALSE` omits the page-scoped
+#'   toolbar.
```
cpsievert: I think the boolean value support here needs to be dropped?

Also interesting that it seemed like you were kind of going for this "inherit" idea at one point, but dropped it. I feel like it would be worth revisiting, but separating "inherit" from "this is (extra) content I want".

pkg-py/src/shinychat/_page_chat.py
```diff
@@ -0,0 +1,1147 @@
...
+    navbar_options: Any = None,
+    sidebar: bool | ChatSidebar = True,
+    artifact: bool | ChatArtifact = True,
```
cpsievert: I'm starting to have feelings along the lines of "artifact is going to become a very overloaded term". Relatedly, just yesterday, I renamed the query chat `/artifact` feature to `/handoff`. From what I've seen so far, we at least currently don't do a whole lot here to actually surface artifacts -- this is more of a layout thing. Maybe we consider a more general name, like `chat_side_panel()` or `chat_dock()`?

pkg-py/src/shinychat/examples/page_chat/navigation/app.py
```diff
@@ -0,0 +1,192 @@
...
+        shiny.__version__ if shiny_version is None else shiny_version
+    )
+    if installed_version < required_shiny_version:
+        raise RuntimeError(
+            "The page-chat navigation example requires shiny >= 1.7.0 "
+            "for offcanvas settings."
+        )
+
+
+require_offcanvas_support()
```
cpsievert: Could we just require 1.7.0 in `pyproject.toml` and drop this?

cpsievert: Oh, it's example, feel free to ignore if you'd like, but still think it'd be fine to drop

pkg-py/src/shinychat/_page_chat.py
cpsievert: Since we're going down the road of respecting user preferences for dark/light mode, maybe we should default to including a dark mode toggle in the toolbar?

pkg-py/src/shinychat/_page_chat.py
```diff
@@ -0,0 +1,1147 @@
...
+        segment.
+    toolbar_global
+        Optional persistent HTML child displayed after the page-scoped toolbar
+        in the navigation controls. It remains mounted while pages are
+        selected and while controls move between desktop and mobile layouts.
+    navbar_options
+        Optional :func:`shiny.ui.navbar_options` that styles the page title bar.
+        Its ``bg``, ``theme``, ``underline``, and HTML attributes are supported.
+        ``position`` and ``collapsible`` are unsupported because ``page_chat()``
+        owns the full-window layout and responsive app menu.
```
cpsievert, line 156: Maybe we should have a `TypedDict` for this that is independent of shiny?

pkg-py/src/shinychat/_page_chat.py
```diff
@@ -0,0 +1,1147 @@
...
+    id: str = "chat",
+    pages: Sequence[ChatNavPanel] | None = None,
+    toolbar: TagChild | None = None,
+    toolbar_global: TagChild | None = None,
+    navbar_options: Any = None,
+    sidebar: bool | ChatSidebar = True,
+    artifact: bool | ChatArtifact = True,
+    window_title: str | None = None,
+    lang: str | None = None,
+    theme: Any = None,
```
cpsievert: Looks like shiny has an actual type for `theme`. Any reason why this needs to be different?

pkg-py/src/shinychat/_page_chat.py
```diff
@@ -0,0 +1,1147 @@
...
+    sidebar: bool | ChatSidebar
+    toolbar: bool | TagChild | None
+    content_width: str
+
+
+def page_chat(
+    title: TagChild,
+    icon: TagChild | None = None,
+    *,
+    id: str = "chat",
```
cpsievert, line 81: I think I'd prefer `icon` to be a keyword param. Also, we really need better icon utilities in python

pkg-py/src/shinychat/_page_chat.py
```diff
@@ -0,0 +1,1147 @@
...
+def _validate_content(content: tuple[TagChild, ...], name: str) -> None:
+    if any(isinstance(item, dict) for item in content):
+        raise TypeError(
+            f"`{name}` content must be HTML children, not an attribute dictionary."
+        )
+
+
+def chat_sidebar(
+    *content: TagChild,
+    history: bool = False,
```
cpsievert: What's the reasoning behind the `False` default? I think it's pretty surprising for `sidebar = chat_sidebar()` to remove it, at least at the `page_chat()` level.

pkg-py/src/shinychat/_page_chat.py
```diff
@@ -0,0 +1,1147 @@
...
+    if not isinstance(id, str):
+        raise TypeError(f"`id` must be a string, not {type(id).__name__}.")
+    if not id.strip():
+        raise ValueError("`id` must not be an empty string.")
+    _validate_page_child(title, "title")
+    if isinstance(title, str) and not title.strip():
+        raise ValueError("`title` must not be an empty string.")
+    _validate_page_child(icon, "icon", allow_none=True)
+    _validate_page_child(toolbar, "toolbar", allow_none=True)
+    _validate_page_child(toolbar_global, "toolbar_global", allow_none=True)
```
cpsievert: These validate helpers smell like R code ported to Python.

I'd rather see the if statements and errors inline. That way the type checker will eliminate those types in downstream code. That might also help will narrowing input types to downstream functions

pkg-py/src/shinychat/_page_chat.py
```diff
@@ -0,0 +1,1147 @@
...
+            )
+            for page in normalized_pages
+            if page.toolbar_key is not None and page.toolbar_key != "home"
+        ),
+        class_="shiny-chat-page-toolbar-sources",
+        hidden=True,
+    )
+
+    shell = Tag(
+        "shiny-chat-page",
```
cpsievert: I'd find it more readable if each child was assigned to an informative name before assembling the `<shiny-chat-page>`. This would also make the downstream `_apply_page_chat_navbar_options()` feel like less of a hack

pkg-py/src/shinychat/__init__.py
```diff
@@ -1,15 +1,37 @@
...
+from ._page_chat_theme import page_chat_theme
 
 __all__ = [
     "Attachment",
     "attachment_to_content",
     "Chat",
+    "ChatArtifact",
+    "ChatArtifactController",
+    "ChatNavPanel",
+    "ChatSidebar",
```
cpsievert: Users aren't expected to call these constructors, are they? If not, lets only export them from `shinychat.types`

pkg-py/src/shinychat/examples/page_chat/navigation/app.py
```diff
@@ -0,0 +1,192 @@
...
+                "review the fieldwork plan."
+            ),
+            sidebar=chat_sidebar(
+                ui.p("Notebook resources"),
+                history=False,
+                width="18rem",
+                open="always",
+                resizable=False,
+            ),
+        ),
```
cpsievert, line 106: Seems there's no way to collapse the sidebar on this page?

gadenbuie: Right, that's from the `open="always"` setting in `chat_sidebar()` (which is itself an extension of `bslib::sidebar()`

docs/bslib-resize-handle-proposal.md
gadenbuie: No, thanks, I'll remove it.

pkg-py/src/shinychat/_page_chat.py
gadenbuie: I'm also good with that.

pkg-py/src/shinychat/_page_chat.py
gadenbuie: I don't think that would work. In practice, you'll want to have page-specific toolbars and global toolbars. I started from a place where pages could choose to inherit the toolbar from the main page or not and you immediately run into problems when you add a button to the home page toolbar that is only effective for the chat page.

So you really do need to separately decide what's page-specific and what's global.

I'm open to reconsidering sharing history in the sidebars. My thinking was that there are cases where people would want to have the conversation selector stay visible across pages, e.g. having a a setup page visible but not hiding the history controls, or when the content of the pages changes depending on the selected bit of history (like with a dedicated artifact or dashboard page).

gadenbuie: Sure, I think it's a reasonable default. That also makes it possible to disable the light/dark mode following if they want to.

pkg-py/src/shinychat/_page_chat.py
gadenbuie: Yeah, I hear you. I think also people might pick up the function thinking that it provides more than it does. On the other hand, in every case that I've thought of using it, the code you'd write reads so much better if you just call it an artifact. And in all the UIs that I surveyed, it's most frequently called an artifact. What that artifact actually **is** changes by chat application, so it seems okay to me to use the word knowing it'll change meaning slightly for each app author.

I don't like _side panel_ or _dock_. They're certainly unopinionated about what you use them for, and honestly I think that's part of the problem. There are enough UI affordances and design layouts available to users to achieve what they want; being opinionated in the naming encourages people to use artifact the way that we want them to use it.

I'm open to other names if we want to keep brainstorming though!

pkg-py/src/shinychat/_page_chat.py
gadenbuie: This was just an oversight, will fix

pkg-py/src/shinychat/_page_chat.py
gadenbuie: Absolutely; part of the other PR I'm preparing removes them from the UI functions.

pkg-py/src/shinychat/_page_chat.py
gadenbuie: Yeah, that sounds reasonable.

pkg-py/src/shinychat/_page_chat.py
gadenbuie: My operating assumption here is that most uses of `chat_sideabr()` will be on pages _other than_ the home page. So most people won't need to use `chat_sidebar()` in `page_chat()`, but will instead use this function in `chat_nav_panel()`. Also someone who is customizing the sidebar in `page_chat()` probably has a non-zero chance of wanting to make big changes to what shows up in the sidebar. So overall my feeling was that more usage of `chat_sidebar()` would _not_ want the history in the sidebar.

I think I briefly looked at doing a `None` and defaulting to `True` in `page_chat()` and `False` in `chat_nav_panel()`; I'd be happy to revisit that.

There's also another angle we could take: `chat_sidebar()` turns on history by default and a regular `sidebar()` (from `shiny` or `bslib`) doesn't carry `history` and turns it off by default. (For convenience, you should be able to pass either `sidebar()` config object into shinychat's `sidebar` arguments.)

pkg-py/src/shinychat/examples/page_chat/navigation/app.py
gadenbuie: Yeah, I didn't want to raise the version floor just for an example. But we might as well since things work better with offcanvas support!

pkg-r/R/page_chat_theme.R
cpsievert, line 1: The SCSS in this file is duplicated for both languages. Would it be worth reducing/eliminating duplication by importing SCSS files? If not, we should have some commentary about why as well as a note for our agents to remember to update both files.

pkg-r/R/page_chat.R
gadenbuie: Good point, will drop.

Mentioned in the other comment, but yes, I explored the inheriting idea. Like you said, you need to separate between extra content for this page and content that's okay to be shared across all pages, which is how I ended up at `toolbar=` (content for this page) and `toolbar_global=` (content for all pages).

It's certainly possible that you might want to have controls that are shared across _some but not all pages_, but it's an awkward scenario that would require an awkward API and I'd rather optimize for the two obvious cases.

pkg-py/src/shinychat/_page_chat.py
cpsievert: Yeah, I'm mainly trying to avoid a situation where we do actually add support an API for extracting artifacts from the chat into that this container, leading to potential naming confusion/conflicy. Maybe just something like `artifact_panel` would help in that regard?

pkg-r/R/page_chat.R
gadenbuie: The `toolbar = <bool>` overload was removed from both R and Python; sidebar boolean compatibility remains. The follow-up instead makes `chat_sidebar(history = NULL)` defer its history choice to the owner, with `page_chat()` defaulting to `TRUE` and `chat_nav_panel()` to `FALSE`, in cfc7c0a and d357b5ae.

pkg-py/src/shinychat/examples/page_chat/navigation/app.py
gadenbuie: Done. The Python package now requires `shiny>=1.7.0`, and the navigation example’s startup guard and obsolete test were removed in c0db1223.

pkg-py/src/shinychat/_page_chat.py
gadenbuie: Agreed. `chat_sidebar()` now defers its history value using `MISSING`; `page_chat()` resolves an unset value to `True`, while `chat_nav_panel()` resolves it to `False`. Explicit history values remain unchanged. This is implemented in d357b5ae.

pkg-py/src/shinychat/_page_chat.py
gadenbuie: The page-chat and Express signatures now mirror Shiny’s accepted theme union, including `str`, `Path`, `Theme`, and `ThemeProvider`, in 0cf274c1.

docs/bslib-resize-handle-proposal.md
gadenbuie: This has been removed in 77c0cc69.

pkg-py/src/shinychat/_page_chat.py
gadenbuie: Done. The page shell now assigns the header and body children to named variables before assembling the root, and navbar options are applied directly to the named header in 272511f8.

pkg-py/src/shinychat/__init__.py
gadenbuie: Agreed. `ChatArtifact`, `ChatNavPanel`, and `ChatSidebar` are no longer top-level exports; they remain available from `shinychat.types`, in 272511f8.

pkg-py/src/shinychat/_page_chat.py
gadenbuie: Done. The page-chat validation checks are now inline, allowing the type checker to narrow values for downstream code, in 272511f8.

pkg-py/src/shinychat/_page_chat.py
cpsievert: Maybe `chat_sidebar(history=)` should default to `"auto"` and resolve to `T` for the home page, `F` otherwise? In that case, I think we could also drop `chat_nav_panel(sidebar: bool)` support? 

One other concern I have with sidebars is that, when history is present, extra UI is gonna start to look busy and clunky pretty quickly. Partially for this reason, I wonder if something like `chat_sidebar_panel()` (the equivalent of putting an accordion in the sidebar) would be worthwhile? In that case too, we could place `new chat` and `search` in the history panel's title, making it so that you can collapse history, while also easily starting a new chat.

pkg-py/src/shinychat/_page_chat.py
cpsievert: Okay, yeah, I can see how you landed here given that global is the secondary case. See https://github.com/posit-dev/shinychat/pull/329#discussion_r3833565361 for other related thoughts. Feel free to resolve this thread.

gadenbuie: Implemented in 5815d2f8 and 0c9db9e0. The default global toolbar now wraps the dark-mode control with `ui.toolbar()`/`bslib::toolbar()`, and the demo apps explicitly include `input_dark_mode()` when supplying custom global toolbars. Passing `None`/`NULL` still opts out.

pkg-py/src/shinychat/_page_chat.py
```diff
@@ -0,0 +1,1155 @@
...
+    toolbar: TagChild | None
+    content_width: str
+
+
+def page_chat(
+    title: TagChild,
+    *,
+    id: str = "chat",
+    icon: TagChild | None = None,
+    pages: Sequence[ChatNavPanel] | None = None,
```
cpsievert: Not that I necessarily want to add this as a requirement right now, but worth considering before we ship, is navigation in the sidebar. Claude, Gippity, and Gemini all take this approach. I wonder if, because of that, it's closer to what folks would want, especially in the (most common?) case where there are no `pages` (i.e., no navbar at all and sidebar owns title)?

It'd be neat if you could easily swap back and forth between sidebar / header nav layout. From what I can see here, I think it's only `navbar_options` that locks us into a navbar?

**Update**: thinking on it more, we'd probably want a different page container for the "monolithic sidebar" layout. Maybe that's reason enough to rename `page_chat()` to `page_chat_navbar()`? I guess if we feel opinionated enough that navbar is the right default layout, we could keep it as is, then eventually add `page_chat_sidebar()`? I must say though that `page_chat_navbar()` feels more self-describing than `page_chat_sidebar()`

pkg-py/src/shinychat/examples/page_chat/navigation/app.py
```diff
@@ -0,0 +1,172 @@
...
+            sidebar=False,
+            toolbar=None,
+        ),
+    ],
+    artifact=chat_artifact(
+        artifact_content("Use the home toolbar to open this preview."),
+        title="Working preview",
+        width=420,
+        open=False,
+    ),
```
cpsievert: When I go to the `Sources` tab, then click on `Show Preview`, this doesn't appear to open. Is the intention that artifacts are only visible on the home/chat page?

cpsievert: Also, on the home page, if I open artifact while the sidebar is open, I can no longer see the chat (even though it feels like there are space for all 3). I think it'd be worth making the artifact less aggressive about going full width.

pkg-py/src/shinychat/_page_chat.py
```diff
@@ -0,0 +1,1162 @@
...
+    toolbar
+        Optional home-page-scoped HTML child displayed with the navigation
+        controls. A page's ``chat_nav_panel(toolbar=)`` can replace this
+        segment.
+    toolbar_global
+        Optional persistent HTML child displayed after the page-scoped toolbar
+        in the navigation controls. When omitted, it contains Shiny's
+        dark/light mode toggle; pass ``None`` to opt out. It remains mounted
+        while pages are selected and while controls move between desktop and
+        mobile layouts.
```
cpsievert: Especially since toolbars are a new thing, it'd be helpful to explicitly mention the shiny toolbar API in some fashion. 

Also, I think I know what you mean by this, but I also think it could be more clear:

>  A page's ``chat_nav_panel(toolbar=)`` can replace this segment.

pkg-py/src/shinychat/__init__.py
```diff
@@ -1,15 +1,31 @@
...
+    chat_ui_history,
+    page_chat,
+)
+from ._page_chat_theme import page_chat_theme
 
 __all__ = [
     "Attachment",
     "attachment_to_content",
     "Chat",
+    "ChatArtifactController",
```
cpsievert: This probably needs to get moved to `types` (or removed completely)?

pkg-py/src/shinychat/_page_chat.py
gadenbuie: Implemented as chat_artifact_panel() and artifact_panel= in commits 21ea0183 (R) and 68b6aac2 (Python). The panel controller/configuration are now named ChatArtifactPanelController and ChatArtifactPanel, with the controller exported from shinychat.types.

pkg-py/src/shinychat/_page_chat.py
gadenbuie: Implemented the single-container direction in e7424f2d: pages is now pages_navbar, leaving page_chat() available for a future pages_sidebar argument rather than adding separate page constructors. Commit 8dbfb169 also extends pages_navbar with the supported standard nav items.

pkg-py/src/shinychat/examples/page_chat/navigation/app.py
gadenbuie: Resolved in 71a4158a: the navigation example Show Panel action now returns to the chat home before revealing the chat-owned artifact panel. The companion responsive layout fix is 6bb3f541, which defers panel takeover until the layout is no longer viable.

pkg-py/src/shinychat/_page_chat.py
gadenbuie: Resolved in e23cc64e. The Core and Express docs explicitly point to shiny.ui.toolbar(), distinguish home-scoped, panel-scoped, and global segments, and state that chat_nav_panel(toolbar=...) replaces the active scoped segment.

pkg-py/src/shinychat/__init__.py
gadenbuie: Resolved in 68b6aac2. The controller is no longer exported from the shinychat package root; ChatArtifactPanel and ChatArtifactPanelController are available from shinychat.types.

pkg-py/src/shinychat/_page_chat.py
gadenbuie: Yeah, that's a good point, and I agree that a future API for extracting artifacts from a conversation should not collide with this layout API. The current API is now `chat_artifact_panel()` with `artifact_panel=`, making clear that it is the opinionated adjacent panel rather than an extracted artifact object. This is implemented in 21ea0183 and 68b6aac2.

pkg-py/src/shinychat/__init__.py
gadenbuie: The public Python controller/configuration types now live in `shinychat.types` rather than the package root. The controller retains its suffix because it is the server-side command interface, distinct from the `ChatArtifactPanel` configuration object. This is in 68b6aac2.

pkg-py/src/shinychat/_page_chat.py
gadenbuie: Yeah, that sounds reasonable. The page-chat configuration `TypedDict`s are now independent of Shiny's types and exported from `shinychat.types` in 272511f8.

pkg-py/src/shinychat/_page_chat.py
gadenbuie: This is a great point, and we should definitely leave room for sidebar navigation. I'd strongly prefer that we don't end up with two separate `page` functions; I'd rather accommodate both navigation styles in one `page_chat()`. That said, I'd also prefer to wait until we've fully designed the API in bslib/py-shiny before we commit to it here, or we can fold sidebar navigation into `chat_sidebar()` or `sidebar()`, depending on how we feel. Regardless, in a world with sidebar navigation, we would not want a bare `pages=` argument on `page_chat()`.

My proposal is that we align the `page_chat()` API in a way that leaves room for the future sidebar-navigation API. To that end, instead of a bare `pages=` argument, I renamed it to `pages_navbar=` in e7424f2d. That leaves `pages_sidebar=` available for the future design.

`pages_navbar=` also now supports the standard navigation-item shapes that page chat can render today, including nested menus, controls, and spacers, in 8dbfb169.

pkg-py/src/shinychat/examples/page_chat/navigation/app.py
gadenbuie: The Sources-page button was intentionally demonstrating why inheriting the home toolbar is not always appropriate, but it exposed that the demo did not provide a complete path back to the chat-owned artifact panel. I updated it so the button returns to the home page and then reveals the panel in 71a4158a.

That led to the related question of navigation management. I think we should support the regular `nav_select()`, `nav_show()`, and `nav_hide()` APIs rather than introduce chat-specific variants. I will add that work to the PR shortly or handle it as a follow-up; once it exists, the example can replace its temporary scoped message with the normal navigation API.

I also adjusted the responsive threshold in 6bb3f541 so the chat and artifact only enter takeover once their minimum widths plus the gap no longer fit. The panel no longer takes over merely because the sidebar is open while there is still room for all three regions.

pkg-py/src/shinychat/_page_chat.py
gadenbuie: Agreed. The Python Core and Express docs now explicitly reference `shiny.ui.toolbar()`, distinguish home-scoped, panel-scoped, and global segments, and explain that `chat_nav_panel(toolbar=...)` replaces the active scoped segment. This is in e23cc64e.

pkg-py/src/shinychat/_page_chat.py
```diff
@@ -0,0 +1,1372 @@
...
+                    normalize_item(child, f"{location}.{index}", True)
+                    for index, child in enumerate(controls, start=1)
+                ),
+            )
+
+        if _is_shiny_nav_panel(item):
+            content = getattr(item, "content")
+            nav = getattr(item, "nav")
+            if not isinstance(nav, Tag):
+                raise TypeError(f"`pages_navbar` item {location} is malformed.")
```
cpsievert: If py-shiny changes its types, this code could start failing at runtime in possibly surprising ways. If instead, we imported the relevant types, we'll get notified by the type checker if our code is incompatible. 

Please revisit the python code (in commit 8dbfb169b970b25f980a6cea22778f1fc1914103) with this in mind. I also prefer `TypeGuard` for the `is_` helpers, and there is no need for the `_` prefix on the helper functions. BTW, `page_navbar()` input args are `NavSetArg | MetadataNode | Sequence[MetadataNode]`

pkg-py/src/shinychat/_page_chat.py
```diff
@@ -0,0 +1,1372 @@
...
+    )
+
+
+def chat_artifact_panel(
+    *content: TagChild,
+    title: str | None = None,
+    width: "CssUnit" = 400,
+    open: bool = True,
+    resizable: bool = True,
+) -> ChatArtifactPanel:
```
cpsievert: Thanks for adding the `_panel` suffix, I think that's an improvement.

That said, I'd still prefer a more general name. I could see this getting used for a variety of things like data sources (a la Plot Pilot) or more general chat configuration stuff (e.g., enabled tools, enabled knowledge stores, current model, prompts, etc).

I think a name like `chat_drawer()` could work, but I'll leave the final decision up to you.

pkg-py/src/shinychat/_page_chat.py
gadenbuie: Good call! Done in e7a18cc7 and 4c1dd713

pkg-py/src/shinychat/_page_chat.py
gadenbuie: Okay, I like `chat_drawer()`! Working on making the change now

cpsievert (approved):
Great work, thanks for this! 🚀 

Please remember to update anything that's now stale in the PR title/description

PR #333, opened 2026-08-21
keep popover portals out of app layout

### LINKED ISSUE #332: Hovering on `shiny-aside` causes chat input box to jump up
With `main` shinychat:

https://github.com/user-attachments/assets/2b7496b9-7363-46d6-bf62-c6c54429c262

Reprex:

```r
library(bslib)
library(shiny)
library(shinychat)

ui <- page_fillable(
  chat_ui(
    "chat",
    messages = list(
      list(
        role = "assistant",
        content = paste(
          "Governed result.",
          '<shiny-aside label="Verified answer">',
          "This answer comes from a governed calculation.",
          "</shiny-aside>"
        )
      )
    )
  )
)

server <- function(input, output, session) {}

shinyApp(ui, server)
```

simonpcouch:
Closes #332. With this PR:

https://github.com/user-attachments/assets/b6825bf5-33bf-4a6a-a474-6cb01cbee325


js/src/chat/chat-tools.scss
```diff
@@ -973,6 +973,14 @@ shinychat-raw-html {
   }
 }
 
+// Keep portal hosts from participating in flex or grid app-shell layout.
+[data-floating-ui-portal]:has(
+  > .shiny-aside-popover,
+  > .shiny-sources-popover
+) {
+  position: fixed;
+}
```
gadenbuie, line 979: If every use of `FloatingPortal` is intended to be out of document flow, we could generalize this to


```suggestion
// Keep portal hosts from participating in flex or grid app-shell layout.
[data-floating-ui-portal] {
  position: fixed;
}
```

Otherwise, if we do need to differentiate between portal children, we could give the ones that need to be out of document flow a general class

```suggestion
// Keep portal hosts from participating in flex or grid app-shell layout.
[data-floating-ui-portal]:has(> .shiny-floating-overlay) {
  position: fixed;
}
```

cpsievert (commented):
Whoops, sorry, this review is outdated now.

js/src/chat/chat-tools.scss
cpsievert: I would rather us go in the 2nd direction than the 1st. That first direction has potential to conflict with other portals we don't control.

simonpcouch: Applied the second suggestion! 8bcaf4b

cpsievert:
Thanks!

PR #334, opened 2026-08-21
feat(chat): add input toolbar

gadenbuie:
This adds a `toolbar_input` argument to both `chat_ui()` and `page_chat()` across Python and R.

The goal is to differentiate between toolbar-style inputs that should be "attached" to the chat input and letting `footer` contain content that's shown at and pinned to the bottom of the chat area.

## Greeting state

In the greeting state, the `toolbar_input` moves with the chat input, `footer` is always pinned to the bottom of the chat area.

file-5a069af315a0e56fabd70138b23697d4

## While chatting

In the chatting state, both the input and its toolbar moves down as expected.

file-4518f8fc2e310818ec870e1e2b9ba69f

## Future work

I think in the long run we'll want to move `toolbar_input` _into_ the chat input area,  but this will require re-designing the chat input. That said, we can still use this same argument and general structure, we'll just end up moving where it shows up in the DOM. In the short, term this unblocks a lot of additional features for most shinychat users.


pkg-py/src/shinychat/examples/page_chat/navigation/app.py
```diff
@@ -137,7 +155,12 @@ def artifact_content(label: str):
 def server(input, output, session):
     chat = Chat(
         "chat",
-        client=EchoChatClient(),
+        client=EchoChatClient(
+            lambda user_input: (
+                f"{input.model()} [{input.reasoning()}] replied to your "
+                f"message: {user_input}"
+            )
+        ),
```
cpsievert, line 146: Since it's non-trivial to wire this up in a real client, it might be nice to at least document how this would work in practice with a real LLM.

gadenbuie: Agreed in the long run; for now though I'd rather have this be a follow-up than a blocker

cpsievert: Ok, are you looking to get this one in the release? What's preventing it from getting out of draft? Just approval on #329?

pkg-py/src/shinychat/express/_page_chat.py
```diff
@@ -31,6 +31,7 @@ def page_chat(
     id: str = "chat",
     pages: Sequence[ChatNavPanel] | None = None,
     toolbar: TagChild | None = None,
+    toolbar_input: TagChild | None = None,
     toolbar_global: TagChild | None = None,
```
cpsievert: I prefer to see the order swapped since `toolbar_global` is semantically closer to `toolbar`

```suggestion
    toolbar_global: TagChild | None = None,
    toolbar_input: TagChild | None = None,
```

pkg-py/src/shinychat/examples/page_chat/navigation/app.py
gadenbuie: Draft was an oversight. Hoping to merge with the other PR. Any feedback before I do that?

pkg-r/R/chat.R
```diff
@@ -268,7 +268,11 @@ chat_greeting <- function(
 #'   controlled globally by the `SHINYCHAT_MAX_ATTACHMENT_SIZE` environment
 #'   variable (a raw byte count; defaults to approximately 30 MB). Files that
 #'   would push the total over this cap are rejected in the browser with a notice.
-#' @param footer Optional HTML content to display below the chat input.
+#' @param toolbar_input Optional HTML content to display directly below the chat
+#'   input. Use it for compact controls such as a model selector or input
+#'   actions.
```
cpsievert: IIRC, you recently made some changes to the other toolbar parameters, pointing folks to the actual toolbar API. Might be nice to do the same here?

cpsievert (approved):

pkg-py/src/shinychat/express/_page_chat.py
gadenbuie: Updated both Python page_chat() signatures and their docstrings so toolbar_global appears before toolbar_input, and aligned the signature contract test.

gadenbuie: 7cc53cf0

pkg-r/R/chat.R
gadenbuie: Updated chat_ui(toolbar_input=) documentation to point users to bslib::toolbar(), and regenerated the affected R documentation. 214119ef

gadenbuie:
Merged manually into #329 in c6427ddf5a7d2590f8c69dceebbfc01be0c24b8b

PR #337, opened 2026-08-24
Programmatic navigation for `page_chat()` (`nav_{select,show,hide}`)

gadenbuie:
## Summary

`page_chat()` now supports standard bslib/Shiny programmatic navigation. No shinychat-specific functions are required — apps use the same `bslib::nav_select()`, `bslib::nav_show()`, `bslib::nav_hide()` (R) and `shiny.ui.update_navset()` (Python) they already know from `page_navbar()` and friends.

The page shell's root `<shiny-chat-page>` element carries a derived `id = "<id>_page"` (e.g. `page_chat(id = "chat")` → `"chat_page"`), and a Shiny input binding exposes the active page server-side as `input$<id>_page` / `input["<id>_page"]()`. The reserved value `"__home__"` represents the main chat page; users may use `"home"` as a regular page value.

## Why we intercept `shiny-change-tab-visibility` instead of using a public handler

This is the most unusual design choice in the PR, so it's worth addressing first.

A design goal for `page_chat()` was that it should work with the standard nav-interaction functions from bslib and py-shiny out-of-the-box — `nav_select()`, `nav_hide()`, `nav_show()`, `update_navset()` — rather than introducing a new set of shinychat-specific primitives. That goal introduced an unexpected wrinkle, but we think the end result is worth it.

The wrinkle: `nav_hide()` / `nav_show()` dispatch through Shiny's built-in `shiny-change-tab-visibility` top-level message, and there is no public API to register a handler for a top-level message type. `addCustomMessageHandler()` only sees `msgObj.custom.*` messages (things sent via `session$sendCustomMessage`), and the internal `addMessageHandler()` is module-private, throws on duplicates, and its registry is not exposed. Monkey-patching `dispatchMessage` would be strictly more fragile (raw websocket payload parsing, load-order sensitivity).

Instead we use the interception point Shiny's dispatcher deliberately exposes. Before dispatching to internal handlers, `dispatchMessage` triggers a `shiny:message` jQuery event and checks `isDefaultPrevented()` — this is Shiny's intended extension point for exactly this use case, and it's identical in py-shiny's bundled client. Our listener binds at script evaluation time (before Shiny connects), so no message can arrive before we're listening.

We don't call `preventDefault()` — that would kill output values batched in the same websocket frame. Instead, when the message's `inputId` belongs to one of our `<shiny-chat-page>` elements, we delete just the `shiny-change-tab-visibility` key from the (mutable) message object and apply our own semantics. Everything else in the batch dispatches normally. For all non-page-chat messages the listener is completely passive.

Page selection (`nav_select` / `update_navset`) takes the first-class path: a standard Shiny input binding on `<shiny-chat-page>` — no interception needed.

## What changed

**Reserved home value renamed to `"__home__"`**
- Previously the internal home page value was `"home"`, which collided with any user-supplied page that might want that value. Renamed to `"__home__"` (page-value namespace only; sidebar/toolbar-source keys are unchanged).
- `"home"` is now available as a regular page value.

**`ChatPageInputBinding` and navigation methods**
- Registers a Shiny input binding (`shinychat.chatPage`) on `<shiny-chat-page>` that reports the active page value and fires `shiny-chat-page:change` events on selection.
- `ChatPageElement` gains `navSelect(value)`, `navHide(target)`, and `navShow(target)` methods backing the standard bslib tabset messages.
- `navHide()` of the active page returns home first (even if the control is already hidden, e.g. a selected `nav_panel_hidden()`), then hides the control. Hiding home is an error.
- `navShow()` reveals a hidden control in its configured position. `nav_show(select = TRUE)` also selects the page.
- Unknown targets and hide-home surface as client error notifications via `showClientMessage`.

**`nav_panel_hidden()` pre-rendered hidden**
- R now pre-renders `bslib::nav_panel_hidden()` controls in their configured position with the `hidden` attribute, so `nav_show()` reveals them in the right spot. The visible label falls back to the panel's value when no title is supplied. Previously, hidden panels were omitted entirely from the nav controls.

**Removed `shinychat.page_chat_home` custom message**
- The paired navigation examples now use standard `bslib::nav_select("chat_page", "__home__")` / `shiny.ui.update_navset("chat_page", "__home__")` instead of the shinychat-specific message.

**`PageChatController` for Python Playwright tests**
- New `PageChatController` in `pkg-py/src/shinychat/playwright/_page_chat.py` wraps `<shiny-chat-page>` and provides typed sub-locators (`loc_header`, `loc_sidebar`, `loc_nav`, `loc_main`, etc.) and helper methods (`active_page`, `select_page`, `return_home`, `open_nav_offcanvas`, `expect_sidebar_open`/`closed`, etc.). All three page-chat Playwright test files were refactored to use it.

**bslib dependency bumped to >= 0.12.0**
- The R package now requires bslib >= 0.12.0, matching the `nav_panel_hidden()` API used by the pre-rendered hidden controls. The nav-visibility test fixture falls back to toolbar buttons on older bslib so tests don't skip on the declared minimum.

## Test coverage

| Layer | What's covered |
|-------|---------------|
| JS unit (`page-entry.test.ts`) | Input binding round-trip, `navSelect`/`navHide`/`navShow`, hidden-page selection, `nav_menu()` children, `home` as a user page value, error cases |
| R shinytest2 (`apps/page-chat-nav-visibility`) | `nav_select` (including hidden panels and menu children), `nav_hide` (including hiding the active hidden page → returns home), `nav_show` / `nav_show(select = TRUE)`, hide-home and unknown-value errors, input binding round-trip — driven from an offcanvas so controls stay reachable on any active page |
| Python Playwright | `update_navset` selection, input value assertions, all three page-chat test files refactored to `PageChatController` |
| R unit (`test-page-chat.R`) | `nav_panel_hidden()` pre-rendered in position with `hidden` attribute, label fallback to value, `aria-labelledby` wiring, `__home__` reserved value validation |

## Future work

`nav_insert()` and `nav_remove()` are explicitly deferred — they can be done in a separate or future PR.


gadenbuie:
@cpsievert Just wanted to get your read on this before I fold it into the main `page_chat()` PR. (Python CI failures are from that PR, I'll fix in there once I merge.)

pkg-r/R/page_chat.R
```diff
@@ -998,7 +1012,10 @@ page_chat_nav_item <- function(item, pages) {
     ))
   }
   if (identical(item$type, "hidden")) {
-    return(NULL)
+    return(htmltools::tagAppendAttributes(
+      page_chat_nav_control(pages[[item$page_index]]),
+      hidden = NA
+    ))
   }
```
cpsievert, line 1019: How come the py side doesn't need logic like this?

gadenbuie: Because py-shiny doesn't have `ui.nav_panel_hidden()`. https://shiny.posit.co/py/api/core/ui.nav_panel.html

gadenbuie:
I'm going to merge so I can keep moving the main PR forward. But feel free to comment on this if you spot anything else

PR #339, opened 2026-08-24
fix asynchronous title generation

### LINKED ISSUE #338: Title generation warning
I've recently been seeing:

```
Warning: Title generation failed
Caused by error:
! HTTP 400 Bad Request.
ℹ messages.0.content: Input should be a valid array
  [invalid_request_error]
```

shinychat title generation calls:

```r
titler$chat_async(excerpt_text, echo = "none")
```

But [ellmer’s `chat_async()` has no `echo` argument](https://github.com/tidyverse/ellmer/blob/5677bc121a74b4b8d40556c0b49710e90b4bee22/R/chat.R#L408-L427). It treats `echo = "none"` as named message content, causing Anthropic’s `content` payload to serialize as an object instead of an array.

simonpcouch:
Closes #338.

### DIFF
--- a/pkg-r/R/chat_history_title.R
+++ b/pkg-r/R/chat_history_title.R
@@ -63,7 +63,7 @@ generate_title <- function(title_fn, client, recorded_turns) {
       )
       excerpt_text <- paste(excerpt, collapse = "\n\n")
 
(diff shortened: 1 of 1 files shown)


cpsievert:
Thanks!

PR #343, opened 2026-08-25
Public conversation ID, emitted on OTel spans as `gen_ai.conversation.id`

### LINKED ISSUE #307: feat: stable conversation identity for R chat_server()
ShinyChat currently creates `ConversationRecord$id` only after the first completed response is persisted. That is too late to correlate the model work that produced it with the saved conversation.

Expose a stable active conversation ID before managed model work begins, retain it through retries, restores, and switches, and record it on a ShinyChat response span. This gives Commons and similar audit or trajectory consumers a simple, history-store-independent way to identify the conversation a response belongs to without adding application context to ellmer.

<details>

<summary>Approved design plan</summary>

# ShinyChat stable conversation identity design

**Status:** Approved.

**Scope:** R ShinyChat's managed `chat_server()` path, plus the future Commons
telemetry-consumer contract.

## Purpose

ShinyChat already persists a stable ID on each `ConversationRecord`. That ID is
created too late for telemetry: the record does not exist until the first
completed response is saved, after the model call has finished.

This design moves conversation identity earlier in the lifecycle. A new
conversation receives an ID on its first user submission, before model work
begins. The same ID later becomes `ConversationRecord$id`.

ShinyChat exposes the active ID as application state and records it on an
OpenTelemetry span around managed responses. Commons can use that span to
group telemetry by the ShinyChat conversation the user actually selected.

## Goals

- Give a new conversation a stable ID before its first model call.
- Preserve that ID through failure, cancellation, retry, save, restore, edit,
  branch navigation, and client replacement.
- Expose the active ID through the object returned by `chat_server()`.
- Record the ID automatically around model work managed by `chat_server()`.
- Keep ellmer unaware of ShinyChat history identity.
- Let Commons prefer ShinyChat's saved-conversation identity while retaining
  its existing standalone fallback.

## Non-goals

- Add an exchange, response, message, or node ID.
- Model edit-branch lineage in Commons.
- Change `ConversationRecord` storage schema.
- Read ShinyChat's `ConversationStore` from Commons.
- Replay arbitrary ShinyChat UI state in a Commons trajectory.
- Add a public identity API for standalone `chat_enable_history()` users.
- Add Python parity in this work.
- Add compatibility behavior for unreleased telemetry.

## Current behavior

The R `HistoryController` uses `record = NULL` to represent an unsaved draft.
It creates the first `ConversationRecord` in `on_response()`, after the
browser has reported the completed assistant message:

```text
user submits
  -> ellmer model and tool work
  -> ShinyChat renders the response
  -> browser reports the completed messages
  -> HistoryController creates ConversationRecord$id
```

This timing is correct for saving UI state, but too late for correlating model
telemetry with the saved conversation.

`chat_server()` performs model work in an `ExtendedTask`. It currently passes
the client, UI ID, user input, and stream controller into that task. It does
not pass a history identity.

Commons independently allocates one conversation ID for the lifetime of its
client. A `commons_conversation_turn` span records that ID. When one Commons
client is used across restored or switched ShinyChat conversations, the
Commons ID does not identify the saved conversation.

## Terminology

- **Empty draft:** A new chat with no user submission. It has no ID.
- **Identified draft:** A chat that has received a user submission but has not
  yet been saved. It has an active conversation ID and no record.
- **Saved conversation:** A `ConversationRecord` whose `id` equals the active
  conversation ID.
- **Managed response:** Model work started by `chat_server()` in its built-in
  user-submission path.

## 1. Identity lifecycle

Conversation identity belongs to the active conversation, not to a completed
response or a storage write.

### Empty draft

Immediately after initialization or "New chat":

```text
active conversation ID = NULL
record = NULL
```

ShinyChat does not allocate or persist IDs for empty chats.

### First submission

Before invoking the model, `chat_server()` asks the current history controller
to ensure an active ID:

```text
active conversation ID = c_42
record = NULL
```

Repeated calls to ensure the ID return `c_42`. They never allocate another ID
for the same active draft.

### Failure, cancellation, and retry

A failed or cancelled model call does not clear the active ID. A later retry
or submission remains part of `c_42`.

The identified draft remains session-local until a response is saved. It does
not appear in the history store or drawer.

### First successful save

`HistoryController$on_response()` creates the first `ConversationRecord` with
the existing active ID:

```text
active conversation ID = c_42
record$id = c_42
```

`new_conversation_record()` must accept an explicit ID instead of always
allocating one internally.

### Restore and switch

Loading a stored record sets the active ID to `record$id` before another
submission can begin. Switching between records changes the active ID to the
selected record's ID.

A failed load leaves both the current record and active ID unchanged.

### Edit and branch navigation

Edits and sibling navigation change the active path inside one
`ConversationRecord`. They do not change the conversation ID.

### New chat and active deletion

Starting a new chat or deleting the active conversation clears the active ID
to `NULL`. The next user submission allocates a new ID.

### Client replacement

`chat$set_client()` re-registers history. It must carry the current active ID
into the replacement controller so the conversation does not become
temporarily unidentified.

If the active conversation already has a record, normal initialization
restores that record and confirms the same ID. If it is an unsaved identified
draft, the replacement controller inherits the draft ID without creating a
record.

## 2. Controller state and invariants

`HistoryController` gains reactive active-ID state separate from `record`.
The controller provides internal operations with these semantics:

- `conversation_id()` reactively returns one character ID or `NULL`.
- `ensure_conversation_id()` returns the current ID, allocating one when
  necessary.
- Activating a stored record sets both `record` and the active ID.
- Clearing the active conversation clears both `record` and the active ID.
- Seeding a replacement controller sets the draft ID without creating a
  record.

The central invariant is:

> When `record` is not `NULL`, the active conversation ID equals `record$id`.

Restore, switch, initialization, deletion, new-chat, and first-save paths must
use the shared controller operations. They must not assign `record` and active
identity independently.

The ID format remains the existing `c_`-prefixed value produced by
`new_conversation_id()`.

## 3. Public R API

The object returned by `chat_server()` gains:

```r
chat$history$conversation_id()
```

This is a Shiny reactive expression.

It returns:

- `NULL` when history is disabled,
- `NULL` for an empty draft,
- the identified draft ID after the first submission, and
- the stored record ID for a saved or restored conversation.

The reactive remains stable across history-controller replacement performed
by `chat$set_client()`. Internally, `chat_server()` keeps a reactive reference
to the current controller and delegates the public expression to it.

This design does not change the public return value of
`chat_enable_history()`. Standalone users do not receive a new identity API in
this work.

## 4. Managed submission flow

`chat_server()` changes its existing user-input observer:

1. Resolve the current history controller.
2. Call `ensure_conversation_id()` synchronously when history is enabled.
3. Capture the returned ID as a scalar.
4. Pass that scalar into the response `ExtendedTask`.
5. Open telemetry using only the captured scalar.

The task must not re-read `chat$history$conversation_id()` while it runs. A
history switch, new-chat action, or controller replacement must not relabel
in-flight work.

With history disabled, the captured ID is `NULL` and the task follows its
current uninstrumented path.

## 5. OpenTelemetry span

For a managed response with a conversation ID, ShinyChat opens one span:

```text
name: shinychat.response
kind: internal
attribute:
  gen_ai.conversation.id = c_42
```

The span begins before `client$stream_async()` and ends after `chat_append()`
has finished consuming the response stream. This lifetime places Commons,
ellmer model calls, and ellmer tool-loop spans beneath the same ShinyChat
response span.

Conceptually:

```text
shinychat.response                    gen_ai.conversation.id = c_42
└── commons_conversation_turn         gen_ai.conversation.id = commons-client-7
    └── ellmer chat and tool spans
```

The span name is fixed and low-cardinality. ShinyChat records no message
content, title, user identity, or rendered UI in its attributes.

The implementation uses `promises::with_otel_span()` so the active span
survives promise callbacks and closes after asynchronous stream consumption.
This requires:

- `promises >= 1.5.0`, and
- a direct `otel` package dependency for the ShinyChat tracer.

With no configured tracing provider, the wrapper is a no-op. ShinyChat must
not require an `otelsdk` provider or exporter.

ShinyChat emits no identity-bearing response span when history is disabled or
before an ID exists. Greeting generation and conversation-title generation
are not managed user responses and do not receive this span.

## 6. Commons consumption contract

Commons keeps its current client-lifetime conversation ID for standalone use.
It does not need to receive a new ellmer argument or read ShinyChat runtime
state directly.

When reconstructing trajectories, Commons resolves a chat span's conversation
identity in this order:

1. The nearest enclosing `shinychat.response` span carrying a valid
   `gen_ai.conversation.id`.
2. The current Commons wrapper's `gen_ai.conversation.id`.
3. The trace ID fallback.

This precedence is explicit because the inner Commons span already carries a
different value under the same semantic-convention attribute name.

The ShinyChat ID groups model and audit telemetry by saved conversation.
Commons' existing ID continues to group calls in apps that do not use managed
ShinyChat history.

This contract does not attach provenance by message or branch identity.
Conversation-level grouping and exchange-level audit association remain
separate concerns.

### Model failure or cancellation

The response span closes with the captured conversation ID. The active draft
ID remains available for retry. No history record is required.

### Store failure

A failed first store write does not allocate another ID. If the controller has
already constructed an in-memory record, it keeps that record and its ID even
though the conversation is not yet available from the store or history drawer.
The next valid save attempt persists the same record and ID.

The controller must preserve the identity invariant when record construction
succeeds but persistence fails.

### Restore failure

The existing load-before-mutate rule remains. ShinyChat changes neither the
current record nor the active ID when a target record cannot be loaded.

### Navigation during in-flight work

The response task uses the ID captured at invocation. Later reactive state
cannot change the span's identity.

This design does not otherwise change whether ShinyChat permits switching or
editing while a response is in flight.

### OpenTelemetry failure

Tracing is observational. Tracer creation, span activation, attribute
recording, or span closure must not prevent the model call or UI stream from
running.

### Controller unit tests

- A new controller has no conversation ID.
- `ensure_conversation_id()` allocates once and returns the same value on
  repeated calls.
- Failure or cancellation without a save retains the draft ID.
- The first saved record uses the preallocated ID.
- Later saves keep the same ID.
- Restore and switching adopt the stored record ID.
- Edit and sibling navigation retain the ID.
- New-chat and active deletion clear the ID.
- A failed switch leaves the current ID unchanged.
- A replacement controller inherits both saved and unsaved active IDs.
- Whenever a record exists, its ID equals the active ID.

### `chat_server()` integration tests

- `chat$history$conversation_id()` is initially `NULL`.
- The model client observes a non-`NULL` ID before its first managed call.
- The public reactive and eventual stored record report the same ID.
- Failed and retried calls report the same ID.
- Restoring and switching update the public reactive.
- `chat$set_client()` preserves the ID.
- With `history = FALSE`, the reactive remains `NULL`.

### OpenTelemetry tests

- A managed response produces one `shinychat.response` span when tracing is
  enabled.
- The span carries the captured `gen_ai.conversation.id`.
- Commons and ellmer spans are descendants of the response span.
- The span remains active through lazy stream consumption.
- Failure and cancellation close the span without changing its ID.
- History-disabled responses do not produce identity-bearing ShinyChat spans.
- Tracing-disabled responses behave identically to current chat responses.

### Future Commons integration tests

- Restore a saved conversation and continue it in a new session.
- Confirm the new spans group under the saved ShinyChat conversation ID.
- Switch between two saved conversations on one Commons client.
- Confirm the two trajectories remain separate even though the Commons
  fallback client ID is the same.
- Confirm standalone Commons still groups by its existing ID when no
  `shinychat.response` ancestor exists.

## 9. Expected file boundaries

ShinyChat R:

- `pkg-r/R/chat_history_types.R`: accept a preallocated record ID.
- `pkg-r/R/chat_history.R`: active-ID state and lifecycle operations.
- `pkg-r/R/chat_app.R`: public reactive, submission-time capture, controller
  replacement, and response span.
- `pkg-r/DESCRIPTION`: `promises` version floor and direct `otel` dependency.
- Existing R history and chat-server test files: controller, public API, and
  telemetry coverage.

Commons, in a later integration:

- `R/trajectory-read.R`: ShinyChat response-span identity precedence.
- Trajectory tests: restore, switching, and standalone fallback.

No JavaScript, ellmer, Python ShinyChat, or history-store migration is part of
this design.

### Allocate a record only after the response

This is current behavior. It cannot identify the model work that creates the
first saved response.

### Allocate an ID when the user clicks "New chat"

This creates identities for empty abandoned drafts. Allocation on first
submission gives telemetry the ID in time without polluting history state.

### Use only a public reactive

This is explicit but does not automatically transport the ID to Commons inside
the managed model call. Apps could omit or mis-scope the integration.

### Use only an OpenTelemetry span

This transports identity but hides useful application state behind tracing.
The public reactive remains the source of truth; telemetry copies it.

### Add an ellmer request argument

Ellmer does not own ShinyChat history identity. Adding the argument would
broaden ellmer's API for context it does not otherwise use.

### Add a ShinyChat-to-Commons lifecycle hook

This avoids a general response span but creates a package integration protocol,
shared mutable context, and asynchronous cleanup requirements.

### Use Shiny's existing `ExtendedTask` span

That span is conditional on Shiny reactivity collection and is an internal
Shiny implementation detail. A named ShinyChat span provides a stable,
domain-specific boundary.

### Expose standalone `chat_enable_history()` identity now

There is no current consumer requiring a second public API. The internal
controller lifecycle can support that extension later without changing the
managed `chat_server()` contract.

## Success criteria

The design is successful when:

1. The first managed model call has a stable ShinyChat conversation ID.
2. The same ID survives retry and becomes the saved record ID.
3. Restore, switch, edit, branch navigation, and client replacement preserve
   the intended conversation identity.
4. `chat$history$conversation_id()` reports the active identity reactively.
5. Managed response telemetry carries the ID without changing ellmer.
6. Commons can prefer the ShinyChat ID and keep its existing standalone
   fallback.
7. No history store, JavaScript, Python, or message/node identity changes are
   required.
</details>

cpsievert:
Closes #307. Depends on posit-dev/chatlas#398 and tidyverse/ellmer#1106, which add the client-side `conversation_id` support this PR consumes.

## Motivation

shinychat conversations have no identity that's visible outside the history store. This PR gives every conversation a stable ID and exposes it two ways:

1. **Publicly, in app code**: a new reactive accessor — `chat$history$conversation_id()` in R, `chat.history.conversation_id()` in Python — returning the active conversation's ID (`NULL`/`None` when history is disabled or the chat is still empty).
2. **In telemetry**: the ID is handed to the chat client, which records it as `gen_ai.conversation.id` on its own `invoke_agent`/`chat` OpenTelemetry spans — the standard attribute for this purpose in the [OTel GenAI semantic conventions](https://opentelemetry.io/docs/specs/semconv/registry/attributes/gen-ai/). shinychat itself emits no OTel spans.

## Downstream impact (Commons)

chatlas and ellmer already emit GenAI spans; they just had no conversation identity to correlate by. With the client recording the ID on its own spans, Commons can group all model work (token usage, latency, errors) by conversation using the standard attribute — no shinychat-specific knowledge required, and first responses that fail or are cancelled are still attributed. shinychat deliberately stays out of OTel: no tracer, no span lifecycle, no otel dependencies — just one assignment before the model call.

The Commons-side consumption is a follow-up in the Commons repo, not this PR.

## Semantics

- The ID is allocated at the first user submission — before any model work — so it's available to app code and telemetry from the very first response.
- It's stable across retries, restores, conversation switches, and `set_client()` / `chat.client.set()`, and the saved `ConversationRecord` adopts it.
- The ID is set on the client from a scalar captured at submission time, so in-flight work is never relabeled by later switches or client swaps.
- No ID is set when history is disabled. Greetings run before any ID exists. Conversation title generation clones the client after the ID is set, so title-generation spans also carry the ID — arguably correct, since they are model calls belonging to the conversation.
- Minor behavior change: `on_active_id_change` fires at first submission rather than first save, so with `restore_mode = "url"` the `?shinychat_conversation_id=` param appears while the first response streams (reloading such a URL falls through restore gracefully).

## Implementation

`HistoryController` gains active-identity state separate from the saved record (a submitted-but-unsaved draft has an ID and no record), with one invariant: **when a record exists, its `id` equals the active ID**. All restore/switch/init/delete/new-chat paths route through shared ops (`activate_record()` / `clear_active()`), and a single writer fires change notifications only on actual change.

- **R**: a `reactiveVal` tracks the live controller so the public reactive survives `set_client()`'s history re-registration; the replacement controller inherits the ID only for unsaved drafts (seeding a saved ID would risk overwriting the stored record). The ID is assigned to the client's `conversation_id` binding before the model call (capability-checked until the ellmer release lands).
- **Python**: the ID is assigned to the client's `conversation_id` property before the model call; requires `chatlas >= 0.22.0` (0.22.0 is released on PyPI; `uv.lock` resolves it from the registry).

## Testing

- R: 25 tests in `test-conversation-id.R`; Python: 20 tests in `test_conversation_id.py` mirroring the R matrix (lifecycle, integration, client-binding handoff).
- Full pytest suite (225 passed), ruff, and pyright pass; R suite passes (2 pre-existing `test-greeting.R` failures also on main).
- Note: 6 Python files fail `ruff format --check`, all pre-existing on `main` and untouched here to avoid unrelated churn.


pyproject.toml
```diff
@@ -182,3 +182,7 @@ skip-magic-trailing-comma = false
 line-ending = "auto"
 docstring-code-format = true
 docstring-code-line-length = "dynamic"
+
+[tool.uv.sources]
+# TODO: drop once chatlas 0.22.0 is released
+chatlas = { git = "https://github.com/posit-dev/chatlas", branch = "feat/conversation-id" }
```
cpsievert:
```suggestion
```

pkg-r/R/chat_app.R
```diff
@@ -286,6 +295,21 @@ check_ellmer_chat <- function(client) {
...
+# ellmer) are left alone: identity features keep working, just without
+# telemetry.
+# TODO: replace the capability check with an ellmer version floor once the
+# binding is in a released ellmer version.
+set_client_conversation_id <- function(client, id) {
+  if ("conversation_id" %in% names(client)) {
+    client$conversation_id <- id
+  }
+  invisible()
+}
```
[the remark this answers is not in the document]
cpsievert: This check is correct as written: `client` is an R6 *instance* (not the class generator), and `names()` on an R6 instance enumerates public members and active bindings (it delegates to `ls()` on the underlying environment). Verified with R6:

```r
C <- R6::R6Class("C", public = list(x = 1, f = function() {}), active = list(ab = function(v) 1))
names(C$new())
#> [1] ".__enclos_env__" "ab"              "x"               "clone"           "f"
```

So the `conversation_id` active binding on ellmer Chat instances is detected and the handoff works. The R tests in `test-conversation-id.R` also cover this path (the mock client is an R6 instance with a `conversation_id` active binding).

pkg-py/src/shinychat/_history.py
```diff
@@ -256,6 +268,65 @@ def __init__(
...
         self._over_budget_warned: bool = False
 
+    # -- active conversation identity --------------------------------------
+
+    def conversation_id(self) -> str | None:
+        """
+        Reactive read of the active conversation ID: ``None`` for an empty
+        draft, otherwise the ID the eventual ``ConversationRecord`` will (or
+        already does) carry. Allocated on first user submission, before
+        model work.
```
gadenbuie: "Allocated on first user submission, before model work" overclaims slightly: that holds for blessed `chat_app()` apps, but standalone history users allocate at first save. The public `ChatHistory.conversation_id()` docstring words this more accurately — suggest aligning this one with it.

pkg-r/tests/testthat/test-conversation-id.R
```diff
@@ -0,0 +1,690 @@
...
+  # is never reallocated.
+  expect_null(new$record)
+  expect_identical(shiny::isolate(new$conversation_id()), id)
+  expect_identical(new$ensure_conversation_id(), id)
+
+  new$on_response(cid_make_turns())
+  expect_identical(new$record$id, id)
+})
+
+test_that("a replacement controller inherits a saved conversation ID", {
```
gadenbuie: Question for you: this seeds a saved conversation's ID onto a fresh controller, but production `set_client()` only seeds drafts (`active_id` is captured only when `old_ctrl$record` is `NULL`). As written, this test can't catch the regression where someone later "simplifies" `set_client()` to always seed — the record-overwrite scenario the code comment warns about. Options: strengthen "set_client() does not seed a saved conversation's ID" (line 557) to assert the negative against the store, or drop this hypothetical. Your call.

pkg-r/R/chat_app.R
```diff
@@ -544,11 +598,21 @@ chat_server <- function(
...
-      last_input(session$input[[paste0(id, "_user_input")]])
+      user_input <- session$input[[paste0(id, "_user_input")]]
+      last_input(user_input)
+
+      # Resolve the active conversation ID before model work begins and set
+      # it on the client as a scalar: later history switches, new-chat
+      # actions, or client swaps must not relabel in-flight work.
+      hist_ctrl <- history_controller()
+      if (!is.null(hist_ctrl)) {
+        set_client_conversation_id(client, hist_ctrl$ensure_conversation_id())
```
gadenbuie: Same assumption note as the Python site: this protects in-flight work only under the one-active-stream-per-client contract. A future admission model with overlapping submissions (or a client shared across chats) could relabel a still-running stream's spans. One line here would keep the guarantee from being read as unconditional.

pyproject.toml
```diff
@@ -27,7 +27,7 @@ Changelog = "https://github.com/posit-dev/shinychat/blob/main/pkg-py/CHANGELOG.m
 [project.optional-dependencies]
 providers = [
     "anthropic;python_version>='3.11'",
-    "chatlas[mcp]>=0.20.0",
+    "chatlas[mcp]>=0.22.0",
```
gadenbuie, line 30: Sidenote: the PR description still says chatlas is "temporarily resolved from posit-dev/chatlas#398 via tool.uv.sources until release," but 0.22.0 is on PyPI and uv.lock resolves it from the registry — no temporary pin remains. Worth updating the description so reviewers don't go hunting for a pin that no longer exists.

pkg-r/R/chat_app.R
gadenbuie: Parity gap with Python: there, `ensure_conversation_id()` runs inside `handle_user_input`'s try/except, so a failure in `on_active_id_change` (e.g. url-mode `send_navigate`) surfaces through the standard error path. Here an error escapes the observer and kills the response differently. Low impact in practice — `send_navigate` failing implies a dead session — but a `tryCatch` would close the gap.

pkg-py/src/shinychat/_chat.py
```diff
@@ -519,6 +519,22 @@ async def _on_user_submit(
...
+                conversation_id = (
+                    await history_controller.ensure_conversation_id()
+                    if history_controller is not None
+                    else None
+                )
+                # Older chatlas / non-chatlas clients lack this binding;
+                # their telemetry simply goes without the ID.
+                client = chat_client.value
+                if hasattr(client, "conversation_id"):
+                    client.conversation_id = conversation_id
```
gadenbuie, line 538: Worth a one-line comment stating the assumption this relies on: the scalar-on-client handoff protects in-flight spans only because the wire contract allows one active stream per client. If overlapping submissions ever become possible — or two `Chat` instances share a client — a later submission could overwrite the scalar mid-stream and cross-label the earlier stream's spans.

pkg-r/R/chat_history.R
```diff
@@ -64,6 +64,56 @@ HistoryController <- R6::R6Class(
       invisible()
     },
 
+    # Reactive read of the active conversation ID: `NULL` for an empty draft,
+    # otherwise the ID the eventual `ConversationRecord` will (or already
```
gadenbuie, line 68: Same note as the Python counterpart: "Allocated on first user submission, before model work" holds for `chat_server()` apps, but standalone `chat_enable_history()` users allocate at first save. Suggest aligning with the more carefully worded public `history$conversation_id()` documentation.

gadenbuie (approved):
LGTM! We had a few comments and suggestions that should all be relatively easy to resolve.

pkg-py/src/shinychat/_history.py
cpsievert: Fixed in 5f9edb0f: reworded the internal `conversation_id()` docstring to note that standalone history users allocate at first save, not first submission. Same fix applied to the R counterpart in `chat_history.R`.

pkg-r/R/chat_history.R
cpsievert: Fixed alongside the Python site in 5f9edb0f.

pkg-py/src/shinychat/_chat.py
cpsievert: Fixed in 5f9edb0f: the scalar handoff comment now states the one-active-stream-per-client assumption and the cross-labeling risk.

pkg-r/R/chat_app.R
cpsievert: Fixed in 5f9edb0f: same one-stream-per-client caveat added to the R comment.

pyproject.toml
cpsievert: Fixed: the PR description now states chatlas 0.22.0 is released on PyPI and `uv.lock` resolves it from the registry; no temporary pin remains.

pkg-r/R/chat_app.R
cpsievert: Fixed in 2b9ccd73: `set_client_conversation_id()` is now wrapped in `tryCatch`, surfacing failures (e.g. url-mode `send_navigate`) via the standard notification path instead of escaping the observer, matching Python.

pkg-r/tests/testthat/test-conversation-id.R
cpsievert: Took your first option in f12a7ffe: dropped the hypothetical test and strengthened "set_client() does not seed a saved conversation's ID" with store-level negative assertions right after the swap (record untouched, store still holds exactly one record), so an always-seed regression fails there.

PR #345, opened 2026-08-25
Default assistant icon off, with an avatar-independent pending indicator

### LINKED ISSUE #310: Default to no assistant icon
And make it easy to get it back

### LINKED ISSUE #340: Preserve pending-response feedback when the assistant icon is disabled
### Problem

When the assistant icon is disabled with `chat_ui(icon_assistant = FALSE)` / `icon_assistant=False`, the UI loses the visual indicator that the server has received the user message and is waiting for the response stream to start.

With the assistant icon enabled, the animated activity indicator in the assistant avatar provides this feedback. When the avatar is disabled, the chat can appear idle during the interval between message submission and the first streamed content, even though the request is in progress.

### Expected behavior

Disabling the assistant avatar should not remove pending-response feedback. Shinychat could render a small, avatar-independent working/activity state that appears after submission and disappears when the response begins, or expose a supported hook for applications to provide one. It should avoid a brief flash for very fast responses.

### Context

Commons currently works around this with CSS on the empty pending message row, displaying a delayed `Working...` state when no assistant icon is present. See posit-dev/commons#144. This seems like behavior that may belong in shinychat itself.

### Reprex

```r
chat_ui(
  "chat",
  icon_assistant = FALSE
)
```

The same issue applies to the Python API with `icon_assistant=False`.

gadenbuie:
Fixes #310
Fixes #340

## Summary
Two related changes to the assistant avatar behavior, implemented together so disabling the icon doesn't regress pending-response feedback:
- `icon_assistant` now defaults to no icon in both R (`chat_ui()`) and Python (`chat_ui()`/`Chat`), rather than the built-in robot icon. `icon_assistant = TRUE` / `icon_assistant=True` remains the single-argument way to restore the robot icon. Implemented by normalizing `icon_assistant` to `FALSE`/`False` inside `chat_ui()` before the shared `resolve_icon_attr()`/`_resolve_icon_attr()` helper runs, so per-message `icon=` inheritance semantics (where `NULL`/`None` means "inherit the container default") are untouched.
- When the assistant icon is disabled, the chat no longer goes dead between message submit and the first streamed token: the existing `dots_fade` pulsing-dots animation (previously shown only in the icon slot) now renders inline in the message row after a 400ms delay, so fast responses never flash it in. Icon-on behavior is unchanged.

https://github.com/user-attachments/assets/440e047c-e453-40d8-9984-ad669667bb59

## Verification
Run any chat app with `chat_ui("chat")` (R) or `chat_ui("chat")`/`Chat(...)` (Python) with no `icon_assistant` argument — the assistant avatar no longer appears. Send a message: a pulsing three-dot indicator appears inline in the empty response row shortly after submit, then is replaced by the streamed content. Pass `icon_assistant = TRUE`/`icon_assistant=True` to restore the previous robot-icon behavior (dots then render in the icon slot as before).


cpsievert (approved):
Thanks!

PR #350, opened 2026-08-26
Expose send-button state, icon, and styling hooks

### LINKED ISSUE #341: Expose send-button state, icon, and styling hooks
### Summary

Make the chat submit button easier to customize without targeting shinychat’s private DOM structure or replacing its SVG with application CSS.

### Proposed API

- Add a state attribute to the submit button, for example `data-state="empty|ready|pending|cancel"`.
- Make the ready-state submit icon configurable.
- Add public CSS variables for supported submit-button styling, including colors, backgrounds, borders, hover/active states, and icon sizing.
- Add pseudo-private CSS variables for internal state and layout details.
- Define and reuse an internal `--_input-control-inset` property for the send/attach control positioning within the composer.

### Context

Commons currently has to override shinychat’s internal button classes, hide the existing SVG, and inject a replacement icon through a CSS pseudo-element and mask in order to create a quieter outlined send control. Public state and styling hooks would allow this design to be implemented without coupling to private class names or markup.

See posit-dev/commons#144 for the current implementation.

gadenbuie:
Closes #341

## Summary

Make the chat submit button easier to customize without targeting shinychat's private DOM structure or replacing its SVG with application CSS.

**`data-state` attribute.** The submit button now carries `data-state="empty|ready|pending|cancel"`, so apps can target it by state without inspecting internal classes.

**`icon_send` parameter.** A new `icon_send` argument on `chat_ui()` / `Chat()` lets callers swap the ready-state submit icon (SVG from bsicons/faicons, or any HTML). The button provides a filled circular surface by default (state-colored background, white icon); the supplied icon replaces only the glyph inside it.

**Public CSS variables.** New `--shiny-chat-btn-send-*` custom properties expose button size, background, color, border, and hover/active states, plus per-state color overrides (`--shiny-chat-btn-send-color-ready`, `-empty`, `-cancel`). A pseudo-private `--_btn-send-state-color` alias resolves to the current state's color on the button itself, enabling ghost/outline styles with a single variable reference.

**Internal layout variable.** A shared `--_input-control-inset` property normalizes send/attach control positioning within the composer.

## Verification

```r
library(shiny)
library(bslib)
library(shinychat)

# Custom send icon with default filled surface
ui <- page_fillable(
  chat_ui("chat", icon_send = bsicons::bs_icon("send-fill"))
)
  chat_server("chat", ellmer::chat("claude"))
}
shinyApp(ui, server)
```

For a ghost/outline style, target the button directly to move the state color from the background to the foreground:

```css
:root .shiny-chat-btn-send {
  --shiny-chat-btn-send-bg: transparent;
  --shiny-chat-btn-send-color: var(--_btn-send-state-color);
  --shiny-chat-btn-send-border: 1px solid var(--_btn-send-state-color);
}
```


gadenbuie:
Review round addressed in individual commits (df267f05..6d063c88):

- **docs(python)**: repaired the `chat_ui()` docstring — it had been re-indented with `icon_send` spliced in at the wrong level, nesting every subsequent parameter inside its description. The `icon_send` docs are now self-contained (CSS variable reference + ghost-style example) instead of pointing at the R docs. Verified with `quartodoc build`.
- **fix(a11y)**: pending/cancelling send-button states now get `aria-disabled` and `tabIndex={-1}`; they were previously inert only via `pointer-events: none`, leaving a focusable button that silently swallowed keyboard activation.
- **test(js)**: pinned the send-button `data-state` machine (empty/ready/pending/cancel/cancelling), the single-button invariant, inert-state a11y attributes, and `iconSend` rendering rules.
- **fix(css)**: default hover/active dimming moved from a `filter: brightness()` declaration into the `var()` fallback via `color-mix()`, so an explicitly set `--shiny-chat-btn-send-bg-hover`/`-active` is used verbatim.
- **docs(r)**: dropped an undefined `stream` from the `chat_ui()` example, removed a bare `#` splitting the roxygen block, and listed `faicons` in `Config/Needs/website` (matching the `bsicons` treatment).
- **feat!**: dropped the `icon_send = TRUE` sentinel — `NULL`/`None` (default) or `FALSE` select the default arrow; `TRUE` now raises a descriptive error in both R and Python.

Known pre-existing issue (not from this branch): `ChatInput.test.tsx > "removing a focused attachment moves focus to the next one"` is flaky — reproduced with these changes stashed. Filed as a follow-up.

pkg-py/src/shinychat/_chat.py
```diff
@@ -2561,11 +2591,46 @@ def chat_ui(
...
+    icon_send
+        The icon to use for the chat input's ready-state submit button. Can be a
+        HTML or a tag in the form of :class:`~htmltools.HTML` or
+        :class:`~htmltools.Tag`. If `None` (the default) or `False`, a default
+        arrow icon is used. The button provides a filled circular surface
+        (state-colored background, white icon); the supplied icon replaces only
+        the glyph inside it.
+
+        The button's appearance is controlled by CSS variables set on the chat
+        container or any ancestor:
```
cpsievert, line 2608: Not that we have to do it here, but I think we might want to consider making a web page dedicated to the CSS variables, and just linking to that page.

Part of why I say that is these doc strings are getting quite long, so you have to scroll a lot to get a sense of the high level argument descriptions. Maybe a cheap way to address that in the short term would be to have a dedicated section after the arguments that goes through CSS variables.

cpsievert (approved):

PR #353, opened 2026-08-26
fix: require two tildes for strikethrough in markdown rendering

### LINKED ISSUE #349: Require two tildes to trigger strikeouts in markdown
Models often write text like

```
first thing (~$1.50) and second thing (~$10)
```

Which causes the markdown to render as

> first thing (~~$1.50) and second thing (~~$10)

I think the easiest fix will be to configure the markdown renderer to require two tildes to activate the strikeout.

gadenbuie:
Fixes #349

## Summary

`remark-gfm` enables single-tilde strikethrough by default, so text like `(~$1.50)` renders as strikethrough. This passes `{ singleTilde: false }` to `remark-gfm` via a shared options constant (`remarkGfmOptions`) so only `~~text~~` triggers strikethrough.

The constant lives in a new `js/src/markdown/remarkGfmOptions.ts` module (rather than `processors.ts`) to avoid a circular import — `processors.ts` already imports from `rehypeGroundedAsides.ts`. All three `remark-gfm` call sites now import and use the shared constant:

1. `markdownProcessor` (assistant content) in `processors.ts`
2. `userMarkdownProcessor` (user content) in `processors.ts`
3. `groundedSpanProcessor` in `rehypeGroundedAsides.ts`

A regression test (`js/tests/markdown/strikethrough.test.ts`) verifies that single tildes render as literal text, double tildes render as `<del>`, and the exact bug scenario `(~$1.50)` / `~/Documents` are preserved — for both `markdownProcessor` and `userMarkdownProcessor`.

| Before | After |
|--------|-------|
| file-6ee55d31f634ee6b60e376e971d8163b | file-f7634dfc9b17f470138103ab2141965a |

## Verification

```r
library(shiny)
library(bslib)
library(shinychat)

ui <- page_fillable(
  chat_ui("chat", messages = list(
    list(
      role = "assistant",
      content = "
Single tildes should be literal:
- First thing (~$1.50) and second thing (~$10)
- See ~/Documents for details
- About ~5 items remain

Double tildes should strike through:
- ~~This is struck through~~
- ~~$1.50~~ (struck through on purpose)
"
    )
  ))
)

server <- function(input, output, session) {}

shinyApp(ui, server)
```


gadenbuie:
FYI `singleTilde` is the only parse-side option `remark-gfm` exposes. The `Options` type is re-exported directly from `micromark-extension-gfm-strikethrough`, which has exactly one field:

```typescript
export interface Options {
  singleTilde?: boolean | null | undefined
}
```

So `singleTilde: false` is the only parsing adjustment we can make.

cpsievert (approved):

PR #354, opened 2026-08-26
fix: don't let `remove_loading` finalize an in-flight stream

### LINKED ISSUE #336: R pkg: slash-command dispatcher's unconditional `remove_loading` truncates a still-streaming response
If we run the following Shiny app, and type `/work` into the chat:

* `client$stream(content)`: the assistant responds with a joke.
* `client$stream_async(content)`: the assistant responds with nothing.

```r
library(shiny)
library(bslib)
library(shinychat)

ui <- page_fillable(
  chat_ui("chat", placeholder = "Type / for commands, or chat away...")
)

  client <- ellmer::chat_anthropic()
  chat <- chat_server("chat", client = client)

  chat$slash_command("work", "Do some work", function(content) {
    # `content@text` in practice *some work*
    content@text <- "Tell me a joke."

    # stream <- client$stream(content)
    stream <- client$stream_async(content)

    chat_append("chat", stream)
  })
}

shinyApp(ui, server)
```

[`chat_server()`'s slash-command dispatcher](https://github.com/posit-dev/shinychat/blob/c1654aa2e13c979e52a16edace094d30680fa4dd/pkg-r/R/chat_app.R#L568-L618) sends a client-side `remove_loading` action unconditionally, right after this slash-command handler returns, and that action [force-finalizes](https://github.com/posit-dev/shinychat/blob/c1654aa2e13c979e52a16edace094d30680fa4dd/js/src/chat/state.ts#L951-L970) whatever streaming message is currently open on the client, even mid-stream.

[`chat_append()`](https://github.com/posit-dev/shinychat/blob/c1654aa2e13c979e52a16edace094d30680fa4dd/pkg-r/R/chat.R#L764-L776) hands the stream to [`chat_append_stream()`](https://github.com/posit-dev/shinychat/blob/c1654aa2e13c979e52a16edace094d30680fa4dd/pkg-r/R/chat.R#L1014-L1030), whose underlying coroutine unconditionally [runs its first lines](https://github.com/posit-dev/shinychat/blob/c1654aa2e13c979e52a16edace094d30680fa4dd/pkg-r/R/chat.R#L1074-L1094) (opening the streaming message) synchronously, before `chat_append()` returns. As mentioned, this synchronous part opens the streaming message on the client. If `stream` is an async generator (from `stream_async()`; <https://github.com/tidyverse/ellmer/blob/19be478ebf1a2e5d2db96a8aeaca71592c8d3f26/R/chat.R#L471-L491> and <https://github.com/tidyverse/ellmer/blob/19be478ebf1a2e5d2db96a8aeaca71592c8d3f26/R/chat.R#L647>), only that opening part finishes before `$append()` returns; the `for (msg in stream)` loop that follows, and the `chunk_end` action it eventually sends, are deferred until response chunks actually arrive (<https://github.com/posit-dev/shinychat/blob/c1654aa2e13c979e52a16edace094d30680fa4dd/pkg-r/R/chat.R#L1074-L1122>).

By contrast, if `stream` were a synchronous generator (e.g., from `$stream()`), that same `for (msg in stream)` loop could complete during this same slash-command handler invocation, so no in-progress streaming message would remain open when the handler returns.

With `stream_async()` we now face an ordering race: the slash-command handler returns before the response stream has completed, and `chat_server()` immediately emits `remove_loading`. `remove_loading` finalizes the already-opened streaming message, truncating the real, still-arriving response down to an empty finalized message. That message -- no content, not streaming, not a placeholder -- [emits no row at all](https://github.com/posit-dev/shinychat/blob/c1654aa2e13c979e52a16edace094d30680fa4dd/js/src/chat/ChatMessage.tsx#L301-L305).

The real answer never joined that message. Every chunk arriving after the forced finalize is discarded, because the reducer returns the state unchanged when no streaming message is open (<https://github.com/posit-dev/shinychat/blob/c1654aa2e13c979e52a16edace094d30680fa4dd/js/src/chat/state.ts#L593-L595>) and the trailing [`chunk_end`](https://github.com/posit-dev/shinychat/blob/c1654aa2e13c979e52a16edace094d30680fa4dd/js/src/chat/state.ts#L851-L873) is likewise a no-op. `remove_loading` had already re-enabled the composer, so the turn simply looks finished.

The response itself is intact. The loop in shinychat's `chat_append_stream_impl()` continues advancing the ellmer async generator despite the client-side premature finalization, and eventually drives it to exhaustion. Consequently, `$append()` resolves with the full text, and ellmer records the user turn and the assistant's complete reply in `client`. Its [streaming path](https://github.com/tidyverse/ellmer/blob/19be478ebf1a2e5d2db96a8aeaca71592c8d3f26/R/chat.R#L945-L986) seeds a partial turn before the first chunk via [`TurnAccumulator$begin_turn()`](https://github.com/tidyverse/ellmer/blob/19be478ebf1a2e5d2db96a8aeaca71592c8d3f26/R/chat.R#L1158-L1163) and swaps in the finished one once the stream ends via [`TurnAccumulator$complete_turn()`](https://github.com/tidyverse/ellmer/blob/19be478ebf1a2e5d2db96a8aeaca71592c8d3f26/R/chat.R#L1173-L1185). Only the browser is left without it, so the conversation carries on with an answer that was never shown to the user.

**Python side.** [`pkg-py/_chat.py`'s `_on_slash_command`](https://github.com/posit-dev/shinychat/blob/c6121f15192472d058241a9ad987d1b35a84dbbc/pkg-py/src/shinychat/_chat.py#L435) does `await _utils.wrap_async(reg.handler)(...)` and sends `remove_loading` from a `finally`, so an async Python handler is genuinely awaited before the indicator drops.

An R handler's return value is neither inspected nor awaited. If that value is a promise representing unfinished work, the dispatcher nevertheless treats the handler's synchronous return as completion.

[`remove_loading` does three things](https://github.com/posit-dev/shinychat/blob/c1654aa2e13c979e52a16edace094d30680fa4dd/js/src/chat/state.ts#L951-L970):

* drop the loading placeholder
* force-finalize `state.streamingMessage`
* set `inputDisabled: false`

The slash-command dispatcher needs a way to clean up after handlers that "produce no output": drop the loading placeholder and re-enable the composer. Without that, a handler that appends nothing — or one that errors before it does, the case the dispatcher's own [`tryCatch()`](https://github.com/posit-dev/shinychat/blob/c1654aa2e13c979e52a16edace094d30680fa4dd/pkg-r/R/chat_app.R#L575-L610) exists to handle — would leave the indicator up and the composer disabled, so the send cannot simply be removed. But `remove_loading` also force-finalizes any open stream. That extra effect is not valid merely because the handler invocation has returned. In this reproduction, an open `state.streamingMessage` indicates that asynchronous response work is still in progress.

In `pkg-r/R`, `remove_loading` is sent only from the slash-command dispatcher. At that call site, its finalization branch is harmless when no stream is open, but destructive when one is: an open `state.streamingMessage` means the handler has started a response that has not yet finished. Python sends `remove_loading` from two places — the `finally` of `_on_slash_command` and its shared exception handler `_raise_exception` — and finalizing an open message after an error is intentional there.

Releasing the composer early has a second consequence. During a normal turn it stays locked from [`chunk_start`](https://github.com/posit-dev/shinychat/blob/c1654aa2e13c979e52a16edace094d30680fa4dd/js/src/chat/state.ts#L576-L591) until `chunk_end`, so the user cannot start another turn mid-stream. Here `chunk_start` locks it during the coroutine's eager segment and `remove_loading` unlocks it in the same flush, with the response still arriving. A second `/work` can therefore be submitted before the first finishes, and two concurrent `stream_async()` calls on one `client` both mutate its turn history, which ellmer [documents as unsafe](https://github.com/tidyverse/ellmer/blob/19be478ebf1a2e5d2db96a8aeaca71592c8d3f26/vignettes/streaming-async.Rmd#L51). Guarding against this currently falls to the app.

Unfortunately, I don't have a solution at this point; I only had a few minutes to get these observations written down.

cpsievert:
## The bug

In an R Shiny app, register a slash command whose handler streams a response
asynchronously — most commonly via ellmer's `client$stream_async()`:

```r
chat$slash_command("work", "Do some work", function(content) {
  stream <- client$stream_async(content)
  chat_append("chat", stream)
})
```

The assistant's response never appears. The loading indicator vanishes, the
turn looks finished, and no error is shown — but the message list stays empty.
Worse, ellmer still records the full reply in the conversation history, so on
the next turn the model responds as if the user saw an answer they never saw.
(Synchronous streaming via `client$stream()` works fine, which made this easy
to miss.)

Reported in #336, with an accurate diagnosis. Fixes #336

## Why it happens

Three mechanisms interact badly:

1. **Submitting a slash command puts the client into a "waiting" state**: a
   loading placeholder is shown and the composer is disabled. Something
   server-side must eventually resolve that state.

2. **R's slash-command dispatcher sends a `remove_loading` action the moment
   the handler returns.** It has to — a handler that produces no output (or
   errors) would otherwise leave the placeholder up and the composer locked
   forever. Python's dispatcher does the same thing, but from a `finally`
   block *after* `await`ing the handler, so by then no stream is ever open.
   R's Shiny observers are synchronous and can't `await`; when a handler
   starts an async stream, it returns while the response is still arriving.

3. **`remove_loading` did more than its name implies.** Besides removing the
   placeholder, it force-finalized any open streaming message and re-enabled
   the composer. Meanwhile, `chat_append()` opens the streaming message
   *eagerly* (`chunk_start` is sent synchronously) while the content arrives
   later. So the sequence was: stream opens empty → handler returns →
   `remove_loading` finalizes the empty stream (which renders as nothing) and
   unlocks the composer → every subsequent chunk is silently dropped.

The premature composer unlock had a second consequence: a user could submit
another command while a stream was still in flight, starting two concurrent
`stream_async()` calls on one ellmer client — which ellmer documents as
unsafe.

## The fix

Narrow the client's `remove_loading` reducer: when a stream is open, only
remove the loading placeholder — leave the stream, the composer lock, and
`cancelRequested` alone. The stream's own lifecycle resolves everything, as it
does on every normal turn: chunks keep appending, and `chunk_end` finalizes
the message and re-enables the composer.

This is safe because **every opened stream is already guaranteed a terminating
`chunk_end`**: R's `chat_append_stream()` sends one from its error handler,
and Python's `_append_stream` sends one from a `finally` block. That means the
finalize branch of `remove_loading` was redundant in every legitimate
scenario — the only time it ever fired with a stream open was this bug.

The alternative fixes all try to make the *server* smarter about when to send
`remove_loading`, but in R the dispatcher fundamentally cannot know when a
handler's async work finishes: promises start eagerly and detach, so the only
handle on "work is done" is a promise the handler may or may not return.
Tracking in-flight streams would require session-scoped state; chaining on the
handler's return value breaks silently for fire-and-forget handlers (the
idiomatic Shiny style). The client, by contrast, can trivially see whether a
stream is open — so making the client interpret `remove_loading` narrowly
fixes the bug for every handler style at once, with no server-side bookkeeping.

As a bonus, this also fixes a latent Python variant: a Python slash-command
handler that fire-and-forgets a stream without awaiting it was truncated by
the `finally`'s `remove_loading` in the same way.

## What changed

- `js/src/chat/state.ts`: the `remove_loading` reducer no longer finalizes an
  open streaming message, unlocks the composer, or resets `cancelRequested`
  while a stream is in flight. With no stream open, behavior is unchanged.
- `js/tests/chat/state.test.ts`: updated reducer tests — new coverage for the
  open-stream case; finalization-behavior tests now trigger via `chunk_end`
  (the legitimate finalization path).
- `pkg-r/tests/testthat/.../slash-command-async-stream`: new shinytest2
  regression test that drives a real browser through a slash command with an
  async streaming response (fails before this change, passes after), plus a
  synchronous control case.
- Rebuilt JS assets copied to both packages per the repo's asset-distribution
  workflow.

## Testing

- New shinytest2 regression test: fails on the old build (response swallowed),
  passes with this change.
- Full JS suite: 1222 tests pass; lint/typecheck clean.
- Full R suite: passes except two pre-existing `test-greeting.R` bookmark
  failures that also fail on `main`.
- Python `test_chat.py` and `test_page_chat.py` suites pass.


gadenbuie (approved):
LGTM! Thanks!

PR #355, opened 2026-08-26
fix: proxy public CSS vars through private --_ vars for :root override

### LINKED ISSUE #351: Fix CSS scoping issues
Audit shinychat's public CSS variables so that anything with `--shiny-chat-*` can be set on `:root` and doesn't need to be scoped. E.g. the workaround below should not be necessary

```css
/* These rules load before shinychat's own stylesheet (both arrive via the
 * page theme), so overrides of shinychat classes need strictly higher
 * specificity than shinychat's selectors to win regardless of order. */
shiny-chat-container[id] { 
  --shiny-chat-user-message-bg: var(--bs-tertiary-bg, #f4f5f7);
  --shiny-tool-card-spinner-color: var(--bs-secondary-color, #6c757d);
}
```

Extracted from https://github.com/posit-dev/commons/pull/169

gadenbuie:
Fixes #351

## Summary

Public `--shiny-chat-*` CSS variables that had element-level defaults were impossible to override from `:root` because the element's own declaration always won. This PR introduces private `--_` proxy variables for all affected tokens, keeping the default in the `var()` fallback so `:root` overrides take precedence.

**Batch 1 — Proxied theming tokens (13 variables):** Each public variable that was set with a default on an element/class selector now has a `--_` proxy. The proxy reads the public variable in its `var()` fallback; all consumption sites read from the private proxy instead. No `:root` block is needed — the default stays in the fallback. Affected variables: 

* `--shiny-chat-border`
* `--shiny-chat-user-message-bg`
* `--shiny-chat-attachment-size`
* `--shiny-chat-greeting-max-width`
* `--shiny-chat-greeting-color`
* `--shiny-chat-footer-font-size`
* `--shiny-chat-footer-color`
* `--shiny-chat-messages-padding-bottom`
* `--shiny-chat-thinking-header-color`
* `--shiny-chat-thinking-border-color`
* `--shiny-chat-thinking-content-color`
* `--shiny-chat-resize-handle-hit-target`
* `--shiny-chat-resize-handle-indicator-color`

**Batch 2 — Renamed runtime-state variables (5 variables):** JS-computed runtime state variables (not theming tokens) renamed from `--shiny-chat-*` to `--_*` across SCSS and TSX: 

* `--shiny-chat-drawer-width` → `--_drawer-width`
* `--shiny-chat-greeting-offset` → `--_greeting-offset`
* `--shiny-chat-composer-offset` → `--_composer-offset`
* `--shiny-chat-drawer-layout-width` → `--_drawer-layout-width`
* `--shiny-chat-drawer-layout-gap` → `--_drawer-layout-gap`.

**Page sidebar resizer fix:** The page sidebar resizer (`.shiny-chat-page-sidebar-resizer`) is a `shiny-chat-resize-handle` element, so it reads `--_resize-handle-hit-target` directly rather than the public variable.

## Verification

Set any public `--shiny-chat-*` variable on `:root` and confirm it overrides the component default without needing higher specificity:

```css
:root {
  --shiny-chat-user-message-bg: var(--bs-tertiary-bg, #f4f5f7);
  --shiny-chat-border: 2px solid var(--bs-border-color, #e9ecef);
}
```

Before this PR, these `:root` overrides had no effect. After, they take precedence over the built-in defaults.


cpsievert (approved):
Thanks!

PR #358, opened 2026-08-27
fix(py): flush queued appends after message stream context

gadenbuie:
## Summary
When `append_message()` is called while a bare `message_stream_context()` is active, Chat queues the append to preserve message ordering. Previously, a bare context ended its stream but never drained that queue, so the queued message was never sent to the browser.

Drain pending appends after the root stream context sends its end chunk. Nested contexts continue to share their parent stream and do not flush independently.

Add regression coverage that opens a bare streaming context, queues an append, exits the context, and verifies the queued message is delivered.

## Verification

```bash
uv run pytest pkg-py/tests/pytest/test_chat.py -q -k 'message_stream_context_flushes_queued_appends or stream'
uv run ruff check pkg-py/src/shinychat/_chat.py pkg-py/tests/pytest/test_chat.py --config pyproject.toml
uv run pyright pkg-py/src/shinychat/_chat.py
```

### DIFF
--- a/pkg-py/src/shinychat/_chat.py
+++ b/pkg-py/src/shinychat/_chat.py
@@ -1133,6 +1133,7 @@ class Chat:
                     chunk="end",
                     stream_id=stream_id,
                 )
+                await self._flush_pending_messages()
 
     async def _append_message_chunk(
         self,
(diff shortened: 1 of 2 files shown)


cpsievert (approved):
This is a sensible bugfix, but also worth highlighting that part of #311 was dropping support for concurrent streams within the same chat instance (thus getting rid of the queueing/flushing logic). 

Assuming we don't take the "meat" of #311, that part might still be worth splitting off into it's own PR.

PR #359, opened 2026-08-27
fix(history-store): atomic writes and tolerant reads

gadenbuie:
## Problem

FileConversationStore splits each conversation across a small `record.json` file and append-only turn/UI journals. The record references journal sequence IDs, so these files must advance as one logical update.

Previously, both implementations appended journal entries before serializing and replacing `record.json`. A serialization or rename failure could leave new journal data on disk that the record did not reference, while in-memory write state could prevent a safe retry. The UI write gate also compared only message counts, so changed content with the same count was not persisted. R's default JSON precision additionally truncated doubles.

## Solution

The store now prepares all serialized payloads before mutating journals, writes the replacement record to a temporary file, appends journals, then atomically replaces `record.json`. Append and rename failures restore both journals to their pre-write byte sizes and leave staged write state uncommitted.

UI persistence now uses a content digest instead of a message count. Journal reads remain tolerant of damaged entries but warn with the file and line number. R persisted JSON uses 17 significant digits for exact double round-trips.

The implementation documents its boundary: it protects against ordinary I/O failures, but does not provide power-loss durability or cross-process coordination.

## Verification

- `uv run pytest pkg-py/tests/test_history_store.py pkg-py/tests/test_history_matrix.py pkg-py/tests/test_history_types.py pkg-py/tests/test_history_controller.py -q` (`188 passed`)
- `cd pkg-r && Rscript -e "devtools::test(reporter = 'summary')"` (passed; 3 dependency skips and 2 existing warnings)
- `uv run pyright`
- `air format --check pkg-r/`

### DIFF
--- a/.github/workflows/verify-js-built.yaml
+++ b/.github/workflows/verify-js-built.yaml
@@ -46,8 +46,8 @@ jobs:
       - name: 🏭 Build
         run: make js-build
 
-      - name: 🔄 Sync history matrix fixture
-        run: make history-matrix-sync
+      - name: 🔄 Sync test fixtures into R package
+        run: make r-sync-fixtures
 
       - name: 🧐 Check for uncommitted changes
         run: |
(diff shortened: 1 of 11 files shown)


cpsievert (approved):
Thanks!

PR #363, opened 2026-08-28
expose pending indicator styling

simonpcouch:
Related to https://github.com/posit-dev/commons/pull/144. This aims to give commons an entry point to overwrite the circular `...` pending indicator with a different UI element.

Exposing as CSS hooks rather than an `pending_indicator` HTML argument so that commons could implement its indicator entirely via `commons_theme()`.

Adopting in commons (PR headed there shortly) looks like:

https://github.com/user-attachments/assets/c11355dc-19e6-44b2-a186-8d249fc8e712


cpsievert:
I would be open to adopting `Working...` as shinychat's default UI. It does look nice up against thinking and tool calls. That said, I think we'd want a spinner (or a pulsing dot) in the `::before` contents, not the diamond? That or a shimmer animation on the text?

Also, before we take any action here, I'll let @gadenbuie chime in with thoughts since he was here most recently

gadenbuie:
The reason I didn't add "Working ..." initially was primarily because in Shiny I usually try to prefer universal designs that don't introduce English-only strings. (We're not perfect at this, but we try!) It's one part internationalization headache; one part "this word works for app X but is out of place in app Y".

That said, if we really want "Working ..." to fit in with shinychat's other affordances, like thinking and tool calling, I think we should redesign the pending indicator to fully align with the design of those elements. Then the extension point isn't to disable our icon and layer on styles, it's just to change the word that shows up in the element.

Unfortunately, I can't recommend that we use the `content` property or try to have users customize this via CSS, because the word itself should appear in the DOM. There's also a bunch of related ARIA practices that we should consider to make sure it works correctly.

I'm in favor of making those changes in the long run, but I don't think it's necessarily worth the effort (in shinychat) right now while we have much larger and meaningful tasks in flight that we want to land. If you have something in commons that works, however hacky, I think you should keep it for now and we can revisit this post-conf, at which point you could ideally delete any hack interim approaches wholesale.

simonpcouch:
Sure, that makes sense. Thank you!

PR #365, opened 2026-08-28
fix(py): tear down ChatHistory on same-ID Chat reconstruction

gadenbuie:
## Context

In Shiny for Python, a dynamic `Chat` can be reconstructed with the same id within a session (e.g. when dynamic UI re-renders). `Chat.__init__` guards against duplicate registrations with `CHAT_INSTANCES`: the previous instance is popped and `destroy()`ed before the new one registers.

## The bug

`Chat.destroy()` tore down the chat's own effects and bookmarking, but not its `ChatHistory`. `ChatHistory._start()` registered its `_on_session_end` cleanup with `session.on_ended()` and discarded the returned unregister callback, and none of its eight reactive effects (history init, save-on-response, and the select/new/rename/delete/edit/navigate input handlers) were tracked anywhere.

So on each same-ID reconstruction, the old history's input effects stayed **active** on the same input ids — both the stale and the replacement controller would answer select/new/rename/etc., racing to drive the same client chat element — and the old controller and its closures stayed **retained** by the session until session end.

## The fix

- `ChatHistory` now tracks its effects and the `on_ended` unregister callback.
- New `ChatHistory._teardown()` destroys the effects, unregisters the session-end callback, and runs the cleanup session end would have run (bookmark-stamp cancel + `controller.cancel_pending()`). No-op if history never started; safe to call twice.
- `Chat.destroy()` now calls `self.history._teardown()`.
- The session-end closure nulls the tracked handles when it fires, so a `destroy()` after session end doesn't re-run cleanup.

Note that destroying the effects (not just unregistering the session-end callback) is load-bearing — without it the old controller stays live on the shared input ids. One known residual: py-shiny itself registers every effect's `destroy()` with `session.on_ended()` as a strong reference and discards *that* unregister callback, so destroyed effect shells remain session-retained until session end regardless of what we do here. That's platform behavior in py-shiny, out of scope for this PR.

## Verification

New regression tests in `pkg-py/tests/test_chat_history.py` (run against a minimal non-stub mock session that actually executes `ChatHistory._start()`): same-ID reconstruction unregisters the prior history's session-end callback, destroys its effects, and runs `cancel_pending()`; repeated (3x) reconstruction retains exactly one history registration; simulated session teardown runs cleanup exactly once, even with a later `Chat.destroy()`; and `destroy()` is safe when history never started.

The two reconstruction tests fail on the unfixed code (verified by stashing the fix). Full Python unit suite: 542 passed. pyright: 0 errors.

### DIFF
--- a/pkg-py/src/shinychat/_chat.py
+++ b/pkg-py/src/shinychat/_chat.py
@@ -2006,6 +2006,7 @@ class Chat:
         """
         self._destroy_effects()
         self._destroy_bookmarking()
+        self.history._teardown()
 
     def _destroy_effects(self):
         for x in self._effects:
--- a/pkg-py/src/shinychat/_history.py
+++ b/pkg-py/src/shinychat/_history.py
@@ -38,6 +38,7 @@ from ._history_types import (
 if TYPE_CHECKING:
     from htmltools import HTML, Tag, TagList
     from shiny.module import ResolvedId
+    from shiny.reactive._reactives import Effect_
 
     from ._chat import Chat
     from ._chat_types import ChatGreeting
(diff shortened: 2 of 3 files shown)


cpsievert (approved):
Nice catch, thank you!

PR #371, opened 2026-08-31
fix(js): hold the greeting while a history restore may be pending

cpsievert:
## Problem

With history enabled, a returning user's greeting renders immediately on
page load and then vanishes when the server restores their conversation.
The flash scales with restore latency: imperceptible locally, but on hosted
deployments (e.g. Posit Connect Cloud) the greeting sits on screen for
seconds before disappearing — the app shows the wrong UI on every reload.

This is the client-side half of the race #275 fixed server-side. The initial
greeting is serialized as an attribute into the page HTML, so the client
renders it at mount, before the websocket delivers the restore decision that
#275 made authoritative.

## Fix

When history is enabled and a stored conversation ID exists, the greeting
starts in a new `"held"` status and is withheld until the first
`history_update` releases it: dropped if a conversation was restored
(matching the server's existing replay behavior), revealed otherwise. Two
guards cover the edges:

- A "Restoring conversation…" pill appears if the hold exceeds 500ms, so
  slow restores don't read as a blank, broken chat.
- A 15s safety timeout releases the greeting if no `history_update` ever
  arrives (stale ID with history disabled server-side, failed connect) —
  showing it late beats hiding it forever.

<details>
<summary>Reproducing (self-contained app, no API key needed)</summary>

Save as `app.py` and run `uv run shiny run app.py`. `SlowStore` adds 3s to
conversation loads to simulate hosted latency:

```python
from __future__ import annotations

import asyncio
from typing import Any, AsyncGenerator
from unittest.mock import MagicMock

import chatlas
from chatlas import Turn
from chatlas._turn import AssistantTurn

from shinychat import Chat, chat_greeting, chat_ui
from shinychat.types import FileConversationStore, HistoryOptions

STORE_DIR = "/tmp/shinychat-greeting-hold"

GREETING = chat_greeting("## Welcome!\n\nThis greeting should not flash on restore.")


class SlowStore(FileConversationStore):
    """FileConversationStore with simulated hosted-environment latency."""

    async def get(self, *args: Any, **kwargs: Any) -> Any:
        await asyncio.sleep(3)
        return await super().get(*args, **kwargs)


class EchoChatClient(chatlas.Chat):
    """A fake LLM that just echoes the input back, no API key required."""

    def __init__(self) -> None:
        provider = MagicMock()
        provider.name = "echo"
        provider.model = "echo"
        super().__init__(provider)

    async def stream_async(
        self, *args: Any, **kwargs: Any
    ) -> AsyncGenerator[str, None]:  # type: ignore[override]
        user_input = str(args[0]) if args else ""
        self._turns.extend(
            [
                Turn(role="user", contents=user_input),
                AssistantTurn(contents=f"echo: {user_input}"),
            ]
        )

        async def _gen() -> AsyncGenerator[str, None]:
            yield f"echo: {user_input}"

        return _gen()


def app_ui(request: object) -> ui.Tag:
    return ui.page_fillable(chat_ui("chat", greeting=GREETING))


def server(input: Inputs, output: Outputs, session: Session) -> None:
    Chat(
        id="chat",
        client=EchoChatClient(),
        greeting=GREETING,
        history=HistoryOptions(
            store=SlowStore(dir=STORE_DIR),
            scope="demo-user",
            title=None,
        ),
    )


app = App(app_ui, server)
```

Steps:

1. Load the app; the greeting is visible.
2. Send any message (the fake client echoes it), saving a conversation.
3. Reload. On `main`: the greeting shows for ~3s before the restored
   conversation replaces it. On this branch: no greeting, just the
   restoring indicator, then the conversation.
4. No-restore paths: "+ New chat" re-shows the greeting; deleting
   `/tmp/shinychat-greeting-hold` and reloading (stale ID) reveals the
   greeting once the server reports nothing to restore.

</details>

## Notes for reviewers

- JS-only; no server changes. Restore → "New chat" → greeting reappears is
  unchanged (the server re-sends it).
- Known tradeoff: fresh visits with a stale stored ID get the greeting one
  server round trip later; the pathological case waits up to 15s. A
  server-driven `restore_error` signal for true failures was considered and
  deferred as follow-up.
- Built assets in `pkg-r/` and `pkg-py/` included per repo convention.

## Testing

New `HeldGreeting.test.tsx` suite (7 tests): held vs. immediate display,
release on no-restore `history_update`, drop on restore (asserting the
greeting is gone, not merely still held), indicator timing, and the safety
timeout.


cpsievert:
Merging away so commons can take advantage sooner than later

PR #372, opened 2026-08-31
Order history conversations by created_at for stable ordering

cpsievert:
## Problem

The history drawer sorted conversations newest-first by `updated_at`. Since selecting a conversation triggers a save that bumps `updated_at`, simply opening an older conversation moved it to the top of the list and refreshed its timestamp — a confusing experience.

## Changes

- Conversation stores (Python `FileConversationStore`/`InMemoryConversationStore`, R equivalents) now sort `list()` results by `created_at` instead of `updated_at`.
- The history drawer groups ("Today / Last 7 days / Previous") and labels conversations by `created_at`, so positions and timestamps stay stable when opening or updating a conversation.
- Added regression tests in both languages asserting that updating a conversation does not reorder the list.

## Note

Store eviction (`max_store_bytes`) removes conversations from the end of `list()`, so it now evicts the oldest-*created* conversation rather than the least-recently-*updated* one. Happy to decouple eviction ordering from display ordering if we'd rather keep LRU eviction.

## Testing

- `pytest pkg-py/tests/test_history_store.py test_history_controller.py test_history_matrix.py` — 154 passed
- `devtools::test(filter = "chat_history")` — 423 passed
- `make js-test` — 1241 passed


cpsievert:
Merging away so commons can take advantage sooner than later

PR #373, opened 2026-09-01
feat: structured content types for tool UI, HTML, and web activity

gadenbuie:
This PR changes how trusted, non-text content reaches the chat UI. Tool cards, web search activity, and server-rendered HTML used to travel as markup embedded in the message string, which the client scanned back out. They now travel as typed JSON blocks in a structured channel that only the server can write to. This closes a model-output XSS class by construction, and it deletes a lot of string-scanning machinery in the process.

This is a big diff (134 files), but the idea behind it is small. This description builds the idea up first, then maps the diff, then suggests a review path. Reading this before the code should make the code much easier to follow.

This PR is stacked on top of #360, which closed the raw-HTML island sink. The diff you see here is relative to that branch. This PR closes the second sink from the same security review: spoofed tool elements. The model-output XSS story is complete when both ship.

## The problem

A chat message used to be a list of content strings. If the server wanted to show a tool card, it serialized a `<shiny-tool-result>` element into the string. The client parsed the string, found the tag, and rendered a card from its attributes.

The client could not tell who wrote the tag. A model could emit this as ordinary assistant markdown:

```html
<shiny-tool-result value-type="html" custom-display="true" value="<script>…">
```

The client scanned it out like any other tool element and reached `innerHTML` with no user interaction. Tool `title` and `icon` attributes were a second sink through `dangerouslySetInnerHTML`. Filtering can't fix this reliably, because the channel itself mixes model text and server markup.

## The design in three steps

**Step 1: split the channels.** A message is now an ordered list of segments of two kinds. String segments (markdown, html, text, thinking) carry text, including everything a model can produce. Structured blocks (`tool_request`, `tool_result`, `web_search`, `web_search_results`, `web_fetch`, `html_block`) are typed JSON envelopes that only the server can construct.

**Step 2: make the envelope the trust signal.** The client never infers trust from markup found in a string. If a block exists, the server sent it, so its fields are server-attested. HTML-bearing fields (`tool_result.value` with `value_type: "html"`, `title`, `icon`, `footer`, `html_block.content`) render through one audited `RawHTML` sink. Text fields are escaped. Code fields render as code blocks. A forged `<shiny-tool-result>` in markdown is now just text, because nothing scans strings for tool markup anymore.

**Step 3: keep presentation on the client.** The server sends an ungrouped, source-ordered stream of blocks. The client converts each block to its render model on arrival and handles display concerns: adjacent tool blocks merge into one tool loop, adjacent web blocks merge into one activity, search results pair with searches by adjacency. No grouping wrapper exists on the wire.

Two properties round out the design. Every block carries a `version` field, so a client that predates a block's version skips it with a warning instead of breaking the message. And restore re-derives blocks from stored chat turns rather than re-parsing persisted markup, so old UI state is never a trust input.

## What changed, area by area

**Wire protocol** (`js/src/transport/types.ts`, plus Python and R emitters). `MessagePayload.segments` is now a union of string segments and structured blocks. Streaming `chunk` actions only target string segments. Structured blocks arrive complete, either inline in `message.segments` or mid-stream through a new `block_insert` action. One deliberate current limitation: tool blocks don't stream incrementally. The protocol has room for it, but no consumer needs it yet.

**Client conversion** (`js/src/chat/chat-entry.ts`, `tool-model.ts`, `web-activity-model.ts`, `html-block-model.ts`). Blocks are validated and converted to render-model blocks on arrival. Unknown types and unsupported versions warn and skip. Non-`html_block` structured blocks in a user-role message are dropped, since tool UI is never legitimate there. Grouping re-expresses the old `routeToolBlocks` and `rehypeGroupWebActivity` logic over typed blocks instead of parsed markup.

**Tool card trust** (`ToolCard.tsx`, `ToolResult.tsx`, `RawHTML.tsx`). All server-attested HTML fields now render through the single `RawHTML` component, replacing ad-hoc `dangerouslySetInnerHTML` calls. `RawHTML` injects via `innerHTML` inside a React-owned wrapper so Shiny bindings survive re-renders, and it manages its own `bindAll`/`unbindAll` scoped to its element.

**MarkdownStream alignment** (`js/src/markdown-stream/`, `_markdown_stream.py`, `markdown-stream.R`). The streaming-output widget consumes the same structured blocks and the same render components as chat. Its flat wire protocol gained an optional `block` field (a message carries `content` XOR `block`). Tool blocks are intentionally unsupported there; `html_block` and web blocks work, and inline asides work through the existing rehype pipeline.

**Island machinery retirement.** Trusted raw HTML used to travel inside markdown as `<shinychat-raw-html>` island tags, with a disguise/escape plugin pair protecting the parse. With `html_block` carrying that content, the wrapper tags, the per-message HAST round-trip, and most of the island machinery are deleted. What survives is the spoof guard: any island tag that shows up in markdown is model-authored forgery, and `rehypeNeutralizeIslands` reduces it to visible inert text before the component map runs.

**Uniform replace semantics.** A streaming `chunk` with `operation="replace"` now wipes the whole in-flight message (strings, thinking, and structured blocks) and appends the new content. Previously the reducer tried to preserve structured blocks across replaces, which produced a decision tree of special cases. Replace means replace.

**Restore, both languages.** Python adopts R's turns-based restore. A stored `Turn` normalizes its contents into an ordered interleaving of string runs and blocks, and `StoredMessage.wire_segments()` re-interleaves them so wire order matches source order. Persisted UI markup is discarded and never re-parsed for trust. Python also gains a client-authoritative HTML-dependency round-trip: deps travel with the message snapshot, so they can be re-registered on restore in a fresh browser session. One caveat: this restore path is an intermediate state, not the final design. An exchange-tree history rewrite on another branch will save server-emitted messages in a more robust form, and the restore machinery here is built to be replaced by it.

**Deletions.** `routeToolBlocks` string scanning, the `ROUTABLE_CONTENT_TYPES` gate, tool-marker parsing in `tool-protocol.ts`, the web-activity rehype grouping plugins, and the tool-tag component-map bridges are gone. The untrusted component map still maps tool and web tags to `EscapedIsland` as defense in depth, so a forged tag renders as visible text rather than an empty DOM element.

## Suggested review path

The diff is large but layered. This order follows the data:

1. `memory-bank/content-rendering.md` first. It describes the architecture this PR lands, with a key-file map.
2. `js/src/transport/types.ts` for the wire types. Every block type is declared here with its fields.
3. `js/src/chat/chat-entry.ts` (`messagePayloadToData`) for the arrival conversion, then `tool-model.ts` and `web-activity-model.ts` for grouping.
4. `js/src/chat/state.ts` for the reducer: the `block_insert` case and the uniform replace path.
5. `js/src/chat/RawHTML.tsx` and `ToolCard.tsx` for the single-sink trust rule.
6. Server emitters: `pkg-py/src/shinychat/_chat_normalize_chatlas.py` and `_chat_types.py`, then `pkg-r/R/contents_shinychat.R` and `chat.R`.
7. Restore: `pkg-py/src/shinychat/_chat_types.py` (`StoredMessage.wire_segments`) and `pkg-r/R/chat_restore.R`.
8. MarkdownStream: `js/src/markdown-stream/markdown-stream-entry.ts` and `MarkdownStream.tsx`.

The dist directories (`pkg-r/inst/lib/`, `pkg-py/src/shinychat/www/`) are rebuilt assets. Skip them; they mirror `js/dist/`.

## Testing

- Spoof regression tests (`js/tests/chat/spoofed-tool-markup.test.tsx`): forged tool results, requests, icon/title sinks, and web carriers in assistant markdown all render inert, across presentation variants and streaming paths. These were written against the old code first, and the web-carrier cases failed there as predicted.
- Unit suites: vitest (1289 passed), pytest (603 passed, non-Playwright), testthat (1883 tests).
- Playwright: deterministic end-to-end coverage for streaming and non-streaming tool flows, web search/fetch display, error states, custom-display results, and restore with structured blocks.
- Every review finding that the suites missed got a regression test in the same commit as its fix.

## Compatibility notes

Old persisted chat UI state is discarded on restore and re-derived from turns. No client shim, no segment migration. R's stored-UI record format bumped to v2 with an exact-match gate; v1 records fall back to turn re-derivation. Python's records never contained island wrappers and stay at v1. The legacy `<shinychat-raw-html>` tag is accepted as input during a compatibility window and renders as neutralized text.


pkg-py/src/shinychat/_history.py
```diff
@@ -157,46 +165,114 @@ def __init__(
...
-        if record.nodes[nid].turns[0].get("role") == "user"
-    ]
+    new_node_ids: list[str] = []
+    n_derived = 0
+    for g in new_groups:
+        node_id = record.append_linear(g)
+        new_node_ids.append(node_id)
+        derived = derive_stored_ui_message(g, session=session)
+        if derived is not None:
+            record.nodes[node_id].ui = [derived]
```
cpsievert: Before this PR, Python persisted the message content that was actually sent to the browser. Now the
stored UI is derived from the turns, which only works for transforms done entirely through `contents_shinychat()`.

As a result, we no longer restore these situations correctly:

1. Any apps using LangChain, Pydantic AI, etc. for response generation.
2. Any apps using `transform_assistant_response`.
3. Any use of `append_message()` to provide things like status messages.
4. Any generator that "wraps" the LLM's response stream to modify it in some way.

What makes this feel like an own-goal rather than a necessary sacrifice: we already build exactly the state we'd want to restore from. `_send_append_message` runs every outbound message through `_as_stored_message()` — post-transform, blocks, deps, server-attested — and then throws it away. Persisting it is wiring, not new machinery.

I'm not necessarily looking to block this PR on it (overall this is great work I'm largely on board with this direction), but I'd like us to agree on the follow-up before release:

- **Shorter term:** restore the server-side send-time capture we had before
  #272 — persist that `StoredMessage` as the primary `ui` source, with
  turns-derivation demoted to the fallback it's good at.
- **Longer term:** I'd like R to adopt the same mentality, and #311 is the
  natural vehicle — a shared server-side transcript for both languages would
  end this class of divergence for good.

pkg-py/src/shinychat/_history.py
```diff
@@ -157,46 +165,114 @@ def __init__(
...
+        d.pop("attachments", None)
+    if not d.get("blocks"):
+        d.pop("blocks", None)
+    d["version"] = STORED_UI_VERSION
+    return d
+
+
+def derive_stored_ui_message(
+    group: list[dict[str, Any]],
+    session: Any = None,
```
cpsievert: Please do more to avoid `Any`.

Also, `TypedDict` is generally preferable `dict` in most situations. There might even be pre-exisiting types you could make use of

pkg-py/src/shinychat/_chat_normalize_chatlas.py
```diff
@@ -39,6 +40,14 @@
 )
 
 
+def _serialize_htmltools(
+    value: TagChild,
+) -> Optional[SerializedHTML]:
+    if value is None:
+        return None
+    return serialize_htmltools(value)
```
cpsievert: Get rid of this helper

pkg-py/src/shinychat/_chat_types.py
```diff
@@ -22,11 +25,158 @@
...
+
+# One entry of `MessagePayload.segments`: a string segment or a structured
+# block (discriminated by the presence of `type`).
+MessagePayloadSegment = Union[StringSegment, StructuredBlock]
+
+
+def is_structured_segment(
+    seg: MessagePayloadSegment,
+) -> TypeGuard[StructuredBlock]:
+    return "type" in seg
```
cpsievert: Doesn't this also need `isinstance(seg, dict)`?

pkg-py/src/shinychat/_history.py
gadenbuie: > **Restore, both languages.** Python adopts R's turns-based restore. ... One caveat: this restore path is an intermediate state, not the final design. An exchange-tree history rewrite on another branch will save server-emitted messages in a more robust form, and the restore machinery here is built to be replaced by it.

Yeah, this part of the PR is intentionally limited so that we're not re-inventing the parts that will come with #311 or the alternative version that I have cooking locally. The goal was to minimize the diff, which I also understand would mean that we shouldn't land this PR as-is.

I think we should wait for the complete message history story to come together first; or if needed we can end up reverting #272 and we can re-align this branch at that point.

pkg-py/src/shinychat/_html_islands.py
```diff
@@ -57,36 +60,59 @@ def is_trusted(child: TagChild) -> bool:
...
 
-    Returns a list of TagChild items ready to be serialized.
+
+def derive_island_parts(content: TagChild | TagList) -> list[IslandPart]:
+    """
+    Partition trusted tag content around `data-shinychat-react` elements.
+
+    Runs without the attribute become `IslandBlockPart` parts (trusted HTML
+    + deps → `html_block` envelopes). Elements with the attribute become
+    `IslandResidualPart` string runs. This is the single derivation shared
```
cpsievert: Partitioning trusted content on `data-shinychat-react` alone means anything without the attribute — including plain strings — falls into the non-React run and gets rendered via `TagList(...).render()`, which escapes plain strings. So for mixed content like `TagList("**markdown** with <b>html</b>", some_react_tag)`, the string portion is doubly broken: its raw HTML is escaped, and since it ships as a trusted `html_block`, its markdown is never processed either.

The streaming paths avoid this by splitting on trust first (`split_content_by_trust()` in `_markdown_stream.py` / `markdown-stream.R`), which routes bare strings to the untrusted markdown path. But the message-construction paths skip that split and call the island derivation directly: `ChatMessage.__init__` here in Python, and in R `build_html_island_segments()` (plus the static `chat_ui()` variant and `render_island_string()` for greeting/drawer surfaces). The same string is therefore untrusted markdown when streamed but trusted, escaped HTML when sent as message content — and R's mixed-list path (`build_wire_segments()`) keeps bare strings as markdown, making the `tagList()` case inconsistent even within R.

The greeting surface has the same problem on the Python side too: `ChatGreeting.__init__` renders non-string content via `TagList(content).render()` (`_chat_types.py`), so a `TagList` greeting containing a bare string gets escaped as well.

Could we run the trust split in the message-construction paths too, so bare strings stay markdown string parts instead of being folded into `html_block` islands?

cpsievert (commented):
Likely more to follow this afternoon, but getting this to you now

pkg-py/src/shinychat/_history.py
cpsievert: Great, sounds good.

pkg-py/src/shinychat/_history.py
gadenbuie: Done in 9ebbea5d — `TurnDict`/`StoredUiMessage` `TypedDicts` now cover the turn-derivation pipeline, and session params are `Session | None`. One tradeoff worth being explicit about: `TurnDict` is a closed `TypedDict` (our `typing_extensions` floor lacks PEP 728 `extra_items`), so type checkers reject subscript access to undeclared provider keys (e.g. `turn["turn_id"]`) even though the runtime contract deliberately permits arbitrary extra keys — everything still passes through `Turn.model_validate`, and callers needing undeclared keys can `cast`. I accepted that because the `TypedDict` documents the keys this codebase actually reads (`role`/`contents`/`content`) and failing closed is the right default for new code. I kept `dict[str, Any]` where the data is genuinely dynamic: the pydantic record models (`TypedDict` extra-key semantics are unsafe for persisted data), the user-supplied `TitleFn` callback, and the save/restore + bookmark-state signatures.

pkg-py/src/shinychat/_chat_normalize_chatlas.py
gadenbuie: Done in 481a8131 — the helper is gone. serialize_htmltools now accepts None and passes it through, and both icon field_serializers call it directly. (The other caller, the pydantic fallback= in _chat_bookmark.py, never receives None since pydantic serializes it natively, so it is unaffected.)

pkg-py/src/shinychat/_chat_types.py
gadenbuie: Yes — and it was a real bug, not just hygiene: "type" in seg on a bare string is substring matching, so a string segment containing the word type would be misclassified as a structured block by the TypeGuard. Fixed in 80a5913b with isinstance(seg, dict) first, plus a regression test in the same commit. (R is safe here: "type" %in% names(seg) is FALSE for a bare string because names() is NULL.)

pkg-py/src/shinychat/_html_islands.py
gadenbuie: Some history on this contract, since it informed where we landed: mixed `TagList` content did work at one point. py-shiny #1868 (Mar 2025) introduced both the docstring contract and `TagChild` support, and its implementation sent rendered TagList HTML out as `content_type="markdown"` — so the client markdown pipeline processed bare strings while tags passed through as raw HTML. The `` ````{=html}` `` fenced-block wrapping (`e118f44a`, Sep 2025) inadvertently broke this by making the whole payload literal. (R is the exception: the contract text was copied into its docs, but the implementation sent `tagList` as `content_type="html"` from day one.)

For this PR we've corrected the docs to match current behavior — `TagList`/`tagList()` content is an HTML container, strings inside are literal text, `HTML()` is the explicit spelling for trusted raw strings. We explored restoring the old behavior via a trust split in construction, but it pulled in a surprising amount of machinery (per-part content types, coalescing, a flattening-boundary XSS regression) for something that previously worked only incidentally and fragilely — a string right after a tag without a blank line got swallowed by CommonMark's HTML-block rules.

On the MessageSegment-like API: we're fully on board, and that's the direction we want to take. This branch already has the unshipped plumbing for it — Python's `ChatMessage(parts=[...])` and R's `list()` content both treat bare strings as markdown segments interleaved with typed blocks, and the wire format and client already handle mixed segment lists. What's left is making it a first-class, documented API (segment constructors, greeting support), which we've filed as a follow-up to land after this PR. The greeting-with-a-select-input use case is exactly what it should serve.

pkg-py/src/shinychat/_chat.py
```diff
@@ -2772,22 +2905,40 @@ def chat_ui(
...
+        for msg in normalized_messages:
+            message_tags.append(
+                Tag(
+                    "shiny-chat-message",
+                    *msg.html_deps,
+                    content=msg.content,
+                    # The assistant default must not leak onto user messages,
+                    # which render `message.icon` directly (no assistant
+                    # fallback chain).
+                    icon=icon_attr if msg.role != "user" else None,
```
cpsievert: Probably worth a general pass to reduce/eliminate redundant commentary

pkg-py/src/shinychat/_chat.py
```diff
@@ -2582,9 +2714,10 @@ def chat_ui(
             * To prevent interpreting as markdown, mark the string as
               :class:`~shiny.ui.HTML`.
         * A UI element (specifically, a :class:`~shiny.ui.TagChild`).
-            * This includes :class:`~shiny.ui.TagList`, which take UI elements
-              (including strings) as children. In this case, strings are still
-              interpreted as markdown as long as they're not inside HTML.
+            * This includes :class:`~shiny.ui.TagList`, which takes UI elements
+              (including strings) as children. TagList content is treated as
+              HTML: strings inside it are literal text (HTML-escaped), not
+              markdown. Use :class:`~shiny.ui.HTML` for trusted raw HTML strings.
```
cpsievert, line 2683: Maybe worth mentioning `ChatMessage(parts=[...])` to achieve a mix of contents

pkg-py/src/shinychat/_chat.py
gadenbuie: lol yeah it's shocking how much effort went into reviewing and cleaning up comments (there are at least a half dozen commits if you're interested in seeing how much was cut and cleaned up). This comment looks like a recent addition; I can take another pass at the newer stuff

pkg-r/R/chat_history_types.R
```diff
@@ -85,6 +85,298 @@ record_meta <- function(record, size_bytes) {
...
+      session = session
+    )
+    messages <- c(messages, list(stored))
+  }
+  messages
+}
+
+# Check whether a stored UI list carries the current version. Older or
+# unversioned UI returns FALSE and is discarded at replay time.
+is_stored_ui_versioned <- function(stored) {
```
cpsievert, line 304: Opinion: I'd lean toward dropping this fallback. v1 records can exist for dev-build users, but migration looks cheap — replay already re-derives from turns when stored UI is NULL, so a short script that strips the stale `ui` field from the JSONL conversation files would cover it. That trades a second restore path we maintain forever for a one-time migration note. Caveat: records without stored turns would degrade to text-only.

js/src/chat/tool-protocol.ts
```diff
@@ -1,262 +1,7 @@
...
-      continue
-    }
-    if (m[1] === undefined) continue
-    const raw = m[2] ?? m[3] ?? m[4] ?? ""
-    attrs[m[1].toLowerCase()] = decodeEntities(raw)
-  }
-  return attrs
-}
-
 export function isTruthyAttribute(
```
cpsievert: Observation: I can't find any callers of this after the tool-markup scanning was removed — looks like an orphan of the refactor. Remove?

pkg-py/src/shinychat/_chat_normalize_chatlas.py
```diff
@@ -536,14 +537,129 @@ def tool_result_contents(x: "ContentToolResult") -> Tagifiable:
...
+        deps.extend(icon_ui["dependencies"])
+
+    return block, deps
+
+
+def tool_request_message(request: Tagifiable) -> ChatMessage:
+    """Wrap shinychat's rich tool-request card in a block-carrying message."""
+    if isinstance(request, ToolRequestComponent):
+        # The default rich path emits the structured `tool_request` envelope.
+        # The tagify code is retained for the legacy/none overrides.
```
cpsievert: Recommendation: I think we should just delete the legacy guards (`is_legacy()` and the branches it gates). `pyproject.toml` already requires `chatlas[mcp]>=0.22.0`, so the chatlas < 0.11.1 check can't ever be true under the declared dependency floor — the legacy branches and the `.tagify()` fallback paths they gate are unreachable. Removing them also drops the `chatlas._version.version_tuple` peek, which relies on chatlas internals. If there's a reason the floor and the guards disagree, worth a comment either way, but as-is this looks like dead code we can shed in this PR.

js/src/chat/chat-entry.ts
```diff
@@ -44,7 +49,63 @@ const CHAT_TOOLBAR_TAG = "shiny-chat-input-toolbar"
...
 const CHAT_DRAWER_TAG = "shiny-chat-drawer"
 
-function parseInitialMessages(container: HTMLElement): ChatMessageData[] {
+/**
+ * Parse the `data-initial-messages` attribute: a JSON array of message
+ * payloads the server embeds when any initial message carries structured
+ * blocks. Each entry replays through `messagePayloadToData()`. Returns null
+ * when malformed so the caller can fall back to static tags.
+ */
+function parseInitialMessagesAttr(
```
cpsievert, line 60: Observation (non-blocking): the initial-segment JSON parsing here and in `markdown-stream-entry.ts` look like near-duplicates with slightly different validation. Since the two surfaces are deliberately sharing the content model now, would a shared parser help keep them from drifting? Fine as a follow-up.

pkg-py/src/shinychat/_markdown_stream.py
```diff
@@ -104,7 +122,10 @@ async def _mock_task() -> str:
 
     async def stream(
```
cpsievert, line 129: Observation (non-blocking): the trust-split → island-parts → deps aggregation block in the middle of this method is doing a lot relative to the rest of it. An extracted helper might make the streaming flow easier to scan. Not asking for a refactor in this PR — flagging in case it bothers you too.

cpsievert (commented):
This is excellent work. Moving trust from scanned markup to server-constructed envelopes eliminates an entire XSS class by design, the versioned blocks with client-side warn-and-skip are exactly the right forward-compat posture, and `content-rendering.md` made a 134-file diff genuinely approachable. I'm on board with the direction and don't see anything here that should block merge.

The inline comments in this round are all aimed at simplification — compat shims that may have outlived their purpose, a couple of orphans the refactor left behind, and one or two spots where extraction might pay off. None are requirements; treat them as places to consider trimming. Given the size of the diff, I suspect there are more opportunities like these lurking, so it may be worth a dedicated cleanup pass (here or as a fast follow-up) before release.

pkg-r/R/chat_history_types.R
gadenbuie: I went the other way on this. Instead of dropping the fallback, I reversed the v1→v2 bump in 63c7022. The bump only existed to guard against dev-build records from this branch that embed pre-af81 `<shiny-chat-raw-html>` wrappers, but I never dogfooded the branch with history enabled, so no v1 records exist anywhere, my machine included. Both languages now ship at `STORED_UI_VERSION = 1`. No second restore path, no migration note, and your text-only caveat is moot since there are no legacy records to degrade.

I'd still keep the gate. The marker is separate from the record-level `schema_version` on purpose. Turns are canonical and stored UI is a derived cache, so this is the cache's schema version. It fails closed to re-derivation instead of rejecting the record outright. Its real moment comes after release, when the client stops understanding something old stored UI contains. The legacy island machinery is "retained for now" per the epic, and its eventual removal is very likely the first real bump. The v1→v2 exercise was a rehearsal for that, and a gate that's already in place and tested is worth more than the ~12 lines deleting it would save. I removed the v1 replay test; the generic gate coverage for unversioned, stale, future, and malformed records stays.

js/src/chat/tool-protocol.ts
gadenbuie: Done in 9db55dba. `isTruthyAttribute` had no callers after the tool-markup scanning removal, so the module is gone. `ToolResultOpenStyle` moved to its only consumer, tool-model.ts.

pkg-py/src/shinychat/_chat_normalize_chatlas.py
gadenbuie: Done in 9b190f74. The floor and the guards agreed, so the legacy branches were dead code. Deleting `is_legacy()` also dropped the `chatlas._version.version_tuple` internals peek, the `packaging.version` import, and a test that monkeypatched its way into the unreachable path.

js/src/chat/chat-entry.ts
gadenbuie: Done in fad1a0db. Both entry points now share `parseJsonArray` from js/src/utils/json.ts. Bonus: the stream entry used to fall back silently on malformed JSON and now warns.

pkg-py/src/shinychat/_markdown_stream.py
gadenbuie: Done in 64364cb6. The trust-split → island-parts → deps block is now `MarkdownStream._send_trusted_segment()` in Python and `send_trusted_segment()` in the R port. Behavior unchanged.

pkg-py/src/shinychat/_chat.py
gadenbuie: Done in 3ad47731. The `messages` docs now point to `ChatMessage(parts=[...])` for interleaving markdown and UI in one message, flagged as provisional since the segment API isn't final yet.

pkg-py/src/shinychat/_chat_types.py
```diff
@@ -414,25 +691,186 @@ def html_deps(self) -> list[SerializedDep] | None:
...
+            segments=segments,
             attachments=message.attachments,
+            blocks=blocks,
+            block_positions=positions or None,
         )
+
+
+def serialize_html_deps(
+    deps: list[HTMLDependency] | None,
+    session: Any,
```
cpsievert: There are still a couple `session: Any` that (I think) should be `session: Session`?

pkg-py/src/shinychat/_history_client.py
```diff
@@ -32,20 +54,24 @@ def _turns_client(self) -> ClientWithTurns:
...
-            return [serialize_chatlas_turn(t) for t in turns]
-        return list(turns)
-
-    def get_turns_grouped(self) -> list[list[dict[str, Any]]]:
+            return cast(
+                list[TurnDict], [serialize_chatlas_turn(t) for t in turns]
+            )
+        # Non-chatlas clients return arbitrary JSON dicts; they are turn
+        # dicts by construction (see TurnDict's docstring).
+        return cast(list[TurnDict], list(turns))
```
cpsievert: The `cast()`s here can be dropped

pkg-py/src/shinychat/_chat.py
```diff
@@ -1509,7 +1527,13 @@ async def _send_append_message(
             action: ChatAction = {"type": "chunk_start", "message": msg_payload}
             await self._send_action(action, message.html_deps)
         elif chunk == "end":
-            if content:
+            if message.blocks:
+                await self._send_message_parts(message, operation)
+            elif has_mixed_content_types(
+                cast(list[ContentSegment], message.segments)
+            ):
```
cpsievert: It'd be preferable to widen the input type to `has_mixed_content_types()` rather than `cast()` (this pattern occurs a couple times in this file).

pkg-py/src/shinychat/_chat_normalize.py
```diff
@@ -452,6 +465,62 @@ def _(chunk: Turn):
     pass
 
 
+BurstCited: TypeAlias = tuple[list[WebSearchSource], dict[str, WebSearchSource]]
+
+
+def _attach_cited_sources(
+    contents: list[Any], blocks: list[StructuredBlock]
+) -> None:
```
cpsievert: The citation fallback is solving a real problem, but `_attach_cited_sources()` doesn't fit this module's design: it re-walks the raw turn contents in a second pass and reaches back into already-built blocks by positional correspondence (i-th results-less burst → i-th `web_search` block), which silently mis-pairs citations if block emission ever changes. It also triplicates the FIFO request/response pairing rule (here, R's `attach_cited_sources`, and the client's `applyWebBlock`), and it only runs on the complete-turn path, so replayed history shows fallback sources that the live stream never does.

Instead, since chatlas search requests/responses already share an id, I'd put that id on the `web_search`/`web_search_results` wire blocks and let the client pair them deterministically rather than FIFO-guessing. Citations could then travel as their own wire block referencing the search id, making "show citations only until results arrive" a purely client-side policy — the client already handles late-arriving results blocks, so streaming works for free, and this post-processing (plus its R twin) gets deleted entirely.

pkg-py/src/shinychat/_chat_types.py
```diff
@@ -22,11 +25,161 @@
...
+# block (discriminated by the presence of `type`).
+MessagePayloadSegment = Union[StringSegment, StructuredBlock]
+
+
+def is_structured_segment(
+    seg: MessagePayloadSegment,
+) -> TypeGuard[StructuredBlock]:
+    # The isinstance guard matters at runtime: `in` on a bare string is
+    # substring matching, so a string segment containing "type" would
+    # otherwise be misclassified as a structured block.
```
cpsievert:
```suggestion
```

pkg-py/src/shinychat/_chat_types.py
```diff
@@ -221,6 +390,8 @@ def __init__(
         role: Role = "assistant",
         content_type: "ContentType | None" = None,
         attachments: "list[Attachment] | None" = None,
+        blocks: "list[StructuredBlock] | None" = None,
+        parts: "list[str | StructuredBlock] | None" = None,
     ):
```
cpsievert, line 506: `ChatMessage.__init__` and `StoredMessage.from_chat_message()` need a look together. Right now `content`, `blocks`, and `parts` are accepted as three independent constructor args with no reconciliation at construction time, and precedence only gets decided later in `from_chat_message()`: if `parts` contains a string, `content`/`blocks` are silently discarded; otherwise `parts` is silently discarded. `ChatMessage(content="hello", blocks=[block_a], parts=["world", block_b])` sends `"world"` + `block_b` — `"hello"` and `block_a` never reach the wire, with no validation or warning. We should either raise when `parts` and standalone `content`/`blocks` are both meaningfully supplied, or drop `content`/`blocks` as separate constructor args and route everything through `parts`.

Separately, the non-string-`content` branch of `__init__` (~70 lines) is doing four distinct things inline — partitioning `TagChild` content into string/HTML-block parts via `derive_island_parts`, resolving the residual string content and content_type, merging those parts with the caller-supplied `blocks`, and coalescing adjacent string runs while deciding whether `self.parts` should even be set. That's a lot of sequential, stateful logic to hold in your head for what's supposed to be a data container's constructor.

Easy wins for breaking it up:
- `partition_tagchild_content(content) -> (content_parts, block_deps_by_index, deps)` — the `derive_island_parts` walk. Note: this loop's block-construction step (`{"type": "html_block", "version": 1, "content": ...}` + conditional `html_deps`) is duplicated verbatim in `_markdown_stream.py`'s `_send_trusted_segment()` and its static Tag-builder. Worth factoring a shared `build_html_block()` into `_html_islands.py` alongside `derive_island_parts()` and having all three sites — including this new helper — call it, rather than adding a fourth copy.
- `resolve_residual_content(content_parts, content_type) -> (content, content_type)` — the residual-html/content_type branch.
- `coalesce_parts(parts) -> list[str | StructuredBlock]` — the adjacent-string-merging loop, which is also generically useful (it's the same shape of logic as the citation-merging code elsewhere).
- `merge_parts(content_parts, supplied_blocks) -> list[str | StructuredBlock] | None` — the merge + "only set parts when multi-part" decision.

None of these need shared state beyond their arguments, so they can be plain module-level functions, independently testable, and would shrink `__init__` to a straight-line sequence of calls.

pkg-py/src/shinychat/_chat_types.py
```diff
@@ -281,8 +537,21 @@ def __init__(
...
+            # envelopes. Bare data-shinychat-react elements stay inline in
+            # the string and resolve through the client's component map.
+            #
+            # The payload is a single HTML string rendered by the client via
+            # innerHTML, so TagList(content).render() is used directly: it
+            # renders trusted tags to HTML and escapes bare strings, which is
+            # the safe-and-correct behavior for this single-string payload
+            # (escaping prevents an XSS sink from untrusted bare strings).
+            # Markdown processing for mixed greetings needs a segments
+            # channel (follow-up, tracked separately as shinychat#2dzc).
```
cpsievert: Still seeing some pretty large commentary that maybe isn't worth it's weight

cpsievert (commented):
Mostly polish/clarity things as well as one correctness issue

pkg-py/src/shinychat/_chat_normalize_chatlas.py
```diff
@@ -536,14 +526,129 @@ def tool_result_contents(x: "ContentToolResult") -> Tagifiable:
...
+        icon_ui = TagList(component.icon).render()
+        block["icon"] = str(icon_ui["html"])
+        deps.extend(icon_ui["dependencies"])
+
+    return block, deps
+
+
+def tool_request_message(request: Tagifiable) -> ChatMessage:
+    """Wrap shinychat's rich tool-request card in a block-carrying message."""
+    if isinstance(request, ToolRequestComponent):
```
cpsievert: Optional, low priority: `tool_request_contents`/`tool_result_contents` can return an empty `TagList()` as a "nothing to render" sentinel, and `tool_request_message`/`tool_result_message` accept the overly broad `Tagifiable` to match. Since `ChatMessage` already accepts `None` as content, `Optional[ToolRequestComponent]`/`Optional[ToolResultComponent]` would say what's actually happening more directly. Not blocking — happy to leave for a follow-up.

pkg-py/src/shinychat/_chat.py
```diff
@@ -397,6 +407,9 @@ async def _mock_task() -> str:
             # TODO: deprecate messages once we start promoting managing LLM message
             # state through other means
             async def _append_init_messages():
+                # `chat_ui(messages=)` initial messages ride the DOM, so only
+                # the deprecated `Chat(messages=)` constructor arg is appended
+                # here.
```
```suggestion
```

pkg-py/src/shinychat/_chat_types.py
gadenbuie: b7e4f9f9 — went with Session | None since history derivation calls these session-free.

pkg-py/src/shinychat/_history_client.py
gadenbuie: Half done in b7e4f9f9: dropped the cast on the non-chatlas path (list[Any] is assignable). The chatlas cast is load-bearing — pyright won't assign dict[str, Any] to the TurnDict TypedDict, and re-typing serialize_chatlas_turn cascades into a Jsonifiable incompatibility downstream — so I kept it, now with a comment explaining why.

pkg-py/src/shinychat/_chat.py
gadenbuie: b7e4f9f9 — widened has_mixed_content_types() to Sequence[_SegmentBase]; both call sites now pass message.segments directly.

pkg-py/src/shinychat/_chat_types.py
gadenbuie: b7e4f9f9

pkg-py/src/shinychat/_chat_types.py
gadenbuie: You're right, and thanks for pointing this out because I think the code smells here point at something bigger than the missing validation. There's a second discard edge too: `ChatMessage(content="hi", parts=[block])` takes the flat path in `from_chat_message()`, which reads `self.blocks` and drops the parts' blocks entirely.

The validation raise felt like treating a symptom, so I traced how we got here. `content`, `blocks`, and `parts` each entered within a few days on this branch, each one reasonable on its own, and the constructor ended up with three overlapping representations of the same content with precedence decided downstream. This function ended up as a sort of frankenstein while trying to resolve spontaneous issues in the plan, not out of intentional design.

So rather than reconciling the three representations, I'm planning to delete two of them. The plan is to make ChatMessage segments-native, with one internal ordered list of string segments (per-segment `content_type`) interleaved with structured blocks. That's the shape the wire, the streaming accumulator, and StoredMessage already use. `content`/`blocks`/`parts` become input representations or derived views. Each representation is converted to segments at construction, and conflicting representations raise an error instead of being silently discarded.

In this PR:

-   The constructor becomes a normalizer for each input form: bare strings become stamped string segments, TagChild content walks the `derive_island_parts` partition, `parts=` passes through. Conflicting combinations raise, with regression tests for both discard edges.
-   `StoredMessage.from_chat_message` becomes the single downstream conversion point.
-   The shared `build_html_block()` extraction you suggested, covering all three hand-built sites including the two in `_markdown_stream.py`.

I'm skipping the four-function `__init__` extraction, at least for now. Once the constructor is a thin normalizer over segments, that structure gets rearranged anyway, and the follow-up that makes the segment list a first-class documented API will finish the job. I'd rather not polish a shape we're about to replace. Also out of scope: the streaming accumulator and StoredMessage's `block_positions` encoding. The latter changes the persisted schema, which deserves its own versioning decision.

Getting this model right is the point of this PR, so thanks for pushing on it.

pkg-py/src/shinychat/_chat_types.py
gadenbuie: Done in e21e1f89

pkg-py/src/shinychat/_chat_types.py
gadenbuie: Done in 0baf3a3f. ChatMessage is now segments-native. Its single internal content representation is an ordered segment list (string ContentSegments interleaved with structured blocks), the same shape as the wire and StoredMessage. content/blocks/parts are derived views, and the constructor normalizes each input form exactly once. Mutually exclusive input representations now raise at construction instead of reconciling downstream, which fixes the silent-discard combos (parts+strings dropping content/blocks, blocks-only parts dropping its blocks, TagChild content overwriting caller parts). StoredMessage.from_chat_message remains the single downstream conversion.

The shared build_html_block() you flagged is factored into _html_islands.py (e8a7db2a), and all three duplicated sites now call it. The four-function __init__ split (partition_tagchild_content / resolve_residual_content / coalesce_parts / merge_parts) is deliberately not in this change. The constructor's three-way branch is now thin enough that the extraction buys little, and we've deferred it to the first-class segment API follow-up, where the constructor will grow anyway.

R's ChatMessage port follows the same segments-native model.

pkg-py/src/shinychat/_chat_normalize.py
gadenbuie: Done, with one change from the plan. When we went to wire the id through, we found chatlas and ellmer don't share a first-class id between search request and response. Shared ids exist only in provider `extra` payloads: Anthropic pairs the request `id` with the response's `tool_use_id`, OpenAI gives the request an id but returns no results payload, and Google provides no ids at all. No provider links a citation to a search id, either.

So the wire carries ids where they exist and pairs deterministically where they don't. `web_search` blocks gained an optional `id` and `web_search_results` an optional `search_id`, both populated from `extra`. The client pairs results to searches by id when present and otherwise to the earliest pending search, so a missing or unmatched id can never misattribute results by position. Citations travel as their own `web_search_citations` block, and the client attaches them to the most recent search in the message, which matches the old complete-turn behavior. "Show citations only until results arrive" is now purely client-side policy, as you suggested.

`_attach_cited_sources` and its R twin `attach_cited_sources` are deleted, and `web_search.cited_sources` is gone from the wire. The citation handler emits on both the streaming and replay paths, so replayed history now matches the live stream.

Commits: 18d01bc5 (client pairing), 861a9b29 (Python), 03fc91d0 (R), b4650346 (review fixes).

pkg-py/src/shinychat/_chat_types.py
```diff
@@ -401,93 +431,160 @@ def __init__(
...
+                raise ValueError(
+                    "`parts` and `blocks` are mutually exclusive: `parts` is "
+                    "the ordered segment list and already carries its blocks."
+                )
+            if not isinstance(content, str) or content:
+                raise ValueError(
+                    "`parts` cannot be combined with `content`: the segment "
+                    "list is the complete content spelling. Pass "
+                    "`content=''`."
+                )
```
cpsievert, line 536: I'm slightly surprised that I'm not seeing the same checks being applied in the property setters?

pkg-py/src/shinychat/_chat_types.py
```diff
@@ -401,93 +431,160 @@ def __init__(
...
+        parts: list[str | StructuredBlock], content_type: ContentType
+    ) -> list[ContentSegment | StructuredBlock]:
+        # Bare strings become string segments stamped with the message
+        # content_type (markdown by default) — the segment-list API is the
+        # deliberate way to mix markdown and UI in one message, unlike
+        # TagList content which is an HTML container. Blocks pass through in
+        # order; adjacent string runs coalesce. Markdown runs join with a
+        # paragraph break (direct concatenation is unsafe at a markdown
+        # seam); other content types concatenate directly so the author
+        # controls the exact bytes.
```
cpsievert: `_parts_to_segments()` seems like a better name here. And the comment is giving AI (in fact a lot of comments and docstrings are)

pkg-py/src/shinychat/_html_islands.py
```diff
@@ -15,6 +15,23 @@
...
+) -> HtmlBlock:
+    """Build an ``html_block`` structured block (a trusted raw-HTML island).
+
+    ``html_deps`` are accepted in whatever serialized form the caller has
+    (session-processed or raw ``as_dict()``); omitted when empty.
+    """
+    block: HtmlBlock = {"type": "html_block", "version": 1, "content": html}
+    if html_deps:
+        block["html_deps"] = html_deps
+    return block
```
cpsievert: This is one of the downsides of embracing `TypedDict` for stuff you need to do downstream things with.

Not saying this has to change, but smells a bit. If we keep it, I might prefer it to be named `html_block()`, live in `_chat_types.py` (next to the TypedDict), and drop the docstring (in favor of the docstring on `HtmlBlock`)

pkg-py/src/shinychat/_chat_types.py
```diff
@@ -387,6 +407,16 @@ class ChatMessageDict(TypedDict):
...
 class ChatMessage:
+    """A chat message whose content is a single ordered segment list.
+
+    ``segments`` (string :class:`ContentSegment`s interleaved with structured
+    blocks) is the one internal content representation — the same shape as
+    the wire and :class:`StoredMessage`. ``content``, ``blocks``, and
+    ``parts`` are derived views; the constructor's input spellings compile to
+    segments once and are mutually exclusive (``parts=`` raises when combined
+    with ``blocks=`` or non-empty ``content``).
+    """
```
cpsievert, line 496: Maybe worth making this docstring user friendly (i.e., follow numpy conventions and document all the params)?

cpsievert (commented):
Getting close!

pkg-py/src/shinychat/_html_islands.py
```diff
@@ -57,36 +77,59 @@ def is_trusted(child: TagChild) -> bool:
...
-    <shiny-chat-raw-html> wrappers.
+    html: str
+    deps: list[HTMLDependency]
 
-    Returns a list of TagChild items ready to be serialized.
+
+IslandPart = Union[IslandBlockPart, IslandResidualPart]
+
+
+def derive_island_parts(content: TagChild | TagList) -> list[IslandPart]:
```
cpsievert, line 81: It seems one implicit assumption behind the logic here is that `content` is not a string (in that case, it's interpreted as markdown). It might help to be more clear about that by checking for `str` and raising early?

pkg-py/src/shinychat/_chat_normalize_chatlas.py
gadenbuie: Done in 2153a5a2. The contents functions now return Optional[ToolRequestComponent]/Optional[ToolResultComponent] (None for the none display override), and the message wrappers accept the same. The None case still becomes ChatMessage(content=TagList()) inside the wrapper. An empty TagList is the spelling that compiles to a zero-segment message, while content=None falls into the island path and emits an empty html_block. So the sentinel lives on, but contained to one spot.

pkg-py/src/shinychat/_chat.py
gadenbuie: Applied in c369dc89.

pkg-py/src/shinychat/_chat_types.py
gadenbuie: Added in 4fa37333. The parts setter now raises when the message carries blocks, since setting parts replaces the segment list outright and would silently drop them. No guard on the content setter, since it keeps blocks trailing, same as the constructor allowing content+blocks together. (Python-only; R builds messages functionally and has no setters.)

pkg-py/src/shinychat/_chat_types.py
gadenbuie: Renamed in 802a87e1 and the comment is tightened. As for the broader pass over comments and docstrings, agreed, and I'll handle it in a targetted pass.

pkg-py/src/shinychat/_html_islands.py
gadenbuie: Went with your suggestion in 3832520c. html_block() now lives in _chat_types.py next to the TypedDict, docstring dropped. As a side benefit, _html_islands.py no longer imports from _chat_types.py, so the module dependency runs one way.

pkg-py/src/shinychat/_chat_types.py
gadenbuie: Agreed. I'll do a broader docstring and comment pass later tonight or tomorrow morning.

pkg-py/src/shinychat/_html_islands.py
gadenbuie: Done in 4fa37333, both languages. derive_island_parts() now raises on a plain (non-HTML) string. On the R side, render_island_string() (greetings, drawer, static message tags) accepts bare strings on purpose and escapes them, so it now handles that case itself before calling the derivation. Both test suites had pinned the old loose behavior, so those tests changed too.

pkg-py/src/shinychat/_chat_types.py
gadenbuie: Done in c3266eedfeb2cfdd2d771be9e05d8352a3755b26, 2c9a442af5f8f894d3a16e33b927251cb233ecc5, and 167938dfd6a2a82103069acc59feb9e8ac2c32f9.

pkg-py/src/shinychat/_chat_normalize_chatlas.py
```diff
@@ -536,14 +530,136 @@ def tool_result_contents(x: "ContentToolResult") -> Tagifiable:
...
+        block["icon"] = str(icon_ui["html"])
+        deps.extend(icon_ui["dependencies"])
+
+    return block, deps
+
+
+def tool_request_message(
+    request: Optional[ToolRequestComponent],
+) -> ChatMessage:
+    """Wrap a tool-request card in a message that carries a structured block."""
```
cpsievert, line 571: Both here and in `tool_result_message()`, I think you could drop all the commentary, and also do an early return for `None`

pkg-py/src/shinychat/_chat_types.py
```diff
@@ -215,39 +401,219 @@ class ChatMessageDict(TypedDict):
...
+                ContentSegment(content=content, content_type=resolved_type),
+                *supplied_blocks,
+            ]
+        else:
+            # TagList/tag content is an HTML container: bare strings inside
+            # it are escaped text nodes (via TagList().render()), NOT
+            # markdown. To mix markdown and UI in one message, use `parts`.
+            # The shared derive_island_parts() partition turns non-React runs
+            # into html_block structured blocks; bare React elements are
+            # rendered and concatenated as residual string segments.
```
cpsievert: Optional: consider a `content_to_segments()` helper here (to help make the `__init__()` feel more scannable)

pkg-py/src/shinychat/_chat_types.py
```diff
@@ -397,6 +766,14 @@ class StoredMessage(BaseModel):
...
     segments: list[StoredSegment]
     attachments: list[Attachment] = []
+    # Server-authored structured blocks carried by this message. Stored
+    # separately from the string `segments`; `wire_segments()` recombines
+    # them.
+    blocks: list[StructuredBlock] = []
+    # Parallel to `blocks`: how many string `segments` precede each block in
+    # the source content. `wire_segments()` re-interleaves instead of
+    # appending all blocks after the string segments.
+    block_positions: list[int] | None = None
```
cpsievert: **1. Stored messages: parallel arrays where everything else interleaves**

Message content now lives in three shapes — `ChatMessage.segments` (interleaved), the wire format (interleaved), and `StoredMessage` (strings + a parallel `blocks`/`block_positions` array that `wire_segments()` re-interleaves with pointer arithmetic). `StoredMessage` is the odd one out, and the parallel encoding is where the complexity concentrates: position recording that's usually discarded, a silent flat-layout fallback on length mismatch, and invariant comments long enough to signal the shape is fighting the reader. R mirrors all of it. Same word, two meanings, one file apart — `ChatMessage.segments` (mixed) vs `StoredMessage.segments` (strings-only). The alternative — rename the mixed list to `parts` everywhere, wire field included — is possible but not better: both end states are equally coherent, the rename would break restoring bookmarks saved by already-released apps (v0.6.1 persists `segments`-keyed message JSON), and aligning the shapes fixes the underlying split for free.

**Why now:** `blocks`/`block_positions` don't exist in any released version, and the only released persisted shape (string-only segments) deserializes unchanged into the interleaved shape. After this ships, the same change needs real migration; before, it's free.

**Fix:** make `StoredMessage.segments` interleaved (`list[StoredSegment | StructuredBlock]`), like `ChatMessage` already is. `wire_segments()` collapses to an identity accessor, `from_chat_message()` becomes a straight copy, and `content`/`blocks` become derived properties. Concrete payoff: `initial_message_payload()` currently builds a `StoredMessage` only to strip its serialized `html_deps` back out of every block (a `cast`-after-dict-comprehension, and an untyped `dict[str, Any]` return that hand-matches `MessagePayload`); with one canonical segment list it projects directly and can return `MessagePayload`.

pkg-py/src/shinychat/_chat_normalize.py
```diff
@@ -418,27 +445,34 @@ def _(chunk: ContentThinkingDelta):
...
-            # is concatenated, so per-item dependencies would otherwise be
-            # dropped and the item's UI would arrive unstyled and unscripted.
             deps += item.html_deps
+            offset = len(blocks)
+            blocks.extend(item.blocks)
+            for idx, block_deps in item._block_html_deps.items():
+                block_dep_objs[offset + idx] = block_deps
+            if item.content:
+                if parts and isinstance(parts[-1], str):
+                    parts[-1] += item.content
```
cpsievert: **2. Turn-merge string joining diverges from R**

The Python `Turn` merge pre-joins adjacent string items with no separator; R's `merge_ellmer_turn_group()` joins with a paragraph break, which is also the seam contract `_parts_to_segments()` already documents. The pre-join can produce the exact markdown-seam bug this branch fixed elsewhere — and makes the two languages render the same turn differently.

**Fix:** delete the manual pre-join; `parts.append(item.content)` — `_parts_to_segments()` already coalesces with the right separator. One-line deletion, restores parity with R; pin with a multi-item `Turn` test.

pkg-py/src/shinychat/_history.py
```diff
@@ -157,46 +168,116 @@ def __init__(
...
+    so the result carries structured ``blocks`` and ``block_positions``.
+    Returns ``None`` when the group has nothing to display.
+    """
+    msg = normalize_turn_group(group)
+    if msg is None:
+        return None
+    return _stored_ui_dict(as_stored_message(msg, session))
+
+
+def derive_node_ui(
```
cpsievert, line 198: **3. `derive_node_ui()` re-implements role derivation**

`_history.py` hand-rolls the tool-result/role logic that `_turn_dict_effective_role()` already is — and R factors it out the same way (`ellmer_turn_effective_role()`). Two copies of this rule is how the "a tool-result turn displays as assistant" invariant drifts.

**Fix:** call the helper (dropping its underscore also dissolves the private aliased import and the doubled import block).

pkg-py/src/shinychat/_markdown_stream.py
```diff
@@ -23,12 +34,28 @@
     "text",
 ]
 
+# Mirrors `asStreamBlock` in js/src/markdown-stream/markdown-stream-entry.ts.
+# The client drops other types with a warning, so reject server-side.
+_STREAM_BLOCK_TYPES = frozenset(
```
cpsievert, line 39: **4. Stream-block allowlist maintained in three places**

The same five-type set is stated independently in Python (`_STREAM_BLOCK_TYPES`), R (`STREAM_BLOCK_TYPES`), and TS (`asStreamBlock`), with comments pointing at each other. Nothing in CI catches a missed update; forgetting one side makes a block stream fine in one language and error in the other.

**Fix:** one test per package pinning the exact set, so adding a block type fails a test naming the places to update.

pkg-py/src/shinychat/_chat.py
```diff
@@ -1201,6 +1213,7 @@ async def _append_message_chunk(
 
         if operation == "replace":
             msg.content = stream_content
+            msg.parts = None
```
cpsievert: **5. Dead call: `msg.parts = None` in the replace path**

After `msg.content = stream_content`, the content setter has already rebuilt segments into flat layout; the `parts = None` line re-runs the identical rebuild and cannot change anything.

**Fix:** delete the line and update the test that exercises it. Keep the parts setter itself — only this call site is redundant.

pkg-py/src/shinychat/_chat.py
```diff
@@ -1509,7 +1524,11 @@ async def _send_append_message(
             action: ChatAction = {"type": "chunk_start", "message": msg_payload}
             await self._send_action(action, message.html_deps)
         elif chunk == "end":
-            if content:
+            if message.blocks:
```
cpsievert: **6. Duplicated send triage**

The `blocks` / mixed-content-types / plain-chunk ladder is copy-pasted across the `end` and mid-stream arms of `_send_append_message`. R routes the same two arms through one helper (`send_wire_segment_actions()`).

**Fix:** extract the shared triage into one private helper, mirroring R's structure. Behavior-preserving.

pkg-py/src/shinychat/_history_client.py
```diff
@@ -1,9 +1,28 @@
...
+    """A JSON-serialized chat turn dict.
+
+    Matches the shape of ``chatlas.Turn.model_dump(mode="json")`` and the
+    plain dicts that non-chatlas clients return from ``get_turns()``. The
+    keys this codebase reads are ``role`` (a string), ``contents``
+    (chatlas content-item dicts), and ``content`` (plain-text content on
+    generic dicts). Extra keys are allowed at runtime.
+    """
+
+    role: str
```
cpsievert: **7. Minor: `TurnDict.role` declared required but read defensively**

`role: str` is typed as required, yet every consumer reads `turn.get("role", ...)` — pyright sees a contract the code doesn't believe exists.

**Fix:** `role: NotRequired[str]` (one line; matches how the code already reads it and how `contents`/`content` are declared).

pkg-py/src/shinychat/_chat_types.py
```diff
@@ -22,11 +36,172 @@
...
+    WebSearchBlock,
+    WebSearchResultsBlock,
+    WebSearchCitationsBlock,
+    WebFetchBlock,
+    HtmlBlock,
+]
+
+# One entry of `MessagePayload.segments`: a string segment or a structured
+# block (discriminated by the presence of `type`).
+MessagePayloadSegment = Union[StringSegment, StructuredBlock]
```
cpsievert: **8. `MessagePayloadSegment` means different things in Python and TS**

In TS, `MessagePayloadSegment` is the string-only segment (`{content, content_type}`, `js/src/transport/types.ts:21`); in Python it's the *union* of string segment and structured block. Same identifier, a member of the union on one side and the union itself on the other — actively misleading when tracing a wire issue across the mirror implementations (TS's union is `SegmentPayload`, types.ts:155).

**Fix:** each side adopts the sibling's name — Python's union becomes `SegmentPayload`, TS's string type becomes `StringSegment`. Type-alias-only renames, no runtime change; afterward the wire names mean exactly the same thing in both languages.

js/src/chat/state.ts
```diff
@@ -42,7 +66,12 @@ export interface ThinkingBlock {
   durationMs?: number
   streaming: boolean
 }
-export type MessageBlock = ContentBlock | ThinkingBlock | ToolLoopBlock
+export type MessageBlock =
```
cpsievert: **9. Client render model shares a name with the wire concept**

On the wire, "block" means a `StructuredBlock`; in the client's internal render model, `MessageBlock` is a different concept (`content` / `thinking` / `tool_loop` render state). The codebase already coined the right term — `htmlBlockToRenderBlock()` (`js/src/chat/html-block-model.ts:43`) — it just hasn't applied it to the type.

**Fix:** rename `MessageBlock` → `RenderBlock`. Internal type, free; keeps "block" unambiguous between the wire and the render model. Leave `MessagePayload.segments` and the `blocks` field names as-is — those are wire/persisted format.

cpsievert (commented):
A round of feedback centered on clarity and simplicity

pkg-py/src/shinychat/_chat_normalize_chatlas.py
gadenbuie: Done — both functions now early-return on `None` and the commentary is gone.

pkg-py/src/shinychat/_chat_types.py
gadenbuie: Done — extracted `content_to_segments()`; the `__init__` island branch is now about six lines.

pkg-py/src/shinychat/_chat_types.py
gadenbuie: Done, both languages. `StoredMessage.segments` is now interleaved (`list[StoredSegment | StructuredBlock]`), `wire_segments()` collapsed to a per-entry projection, `from_chat_message()` is an order-preserving copy, and `content`/`blocks`/`html_deps` are derived views. R's `build_stored_message_from_content()`/`restore_history_message()` got the same change — the re-interleave machinery is deleted there too. No `STORED_UI_VERSION` bump since this format only exists on this branch. One wire delta worth knowing: blocks-only stored messages in R now lead with the empty string segment, matching Python's existing behavior.

pkg-py/src/shinychat/_chat_normalize.py
gadenbuie: Good catch — the pre-join is deleted (`parts.append(item.content)` only), so `_parts_to_segments()` coalesces with the paragraph break, restoring parity with `merge_ellmer_turn_group()`. Pinned with a multi-item Turn test; one existing test had pinned the old no-separator behavior and was updated to the seam contract.

pkg-py/src/shinychat/_history.py
gadenbuie: Done — `NotRequired[str]`, matching how every consumer already reads it. (This surfaced a few direct `turn["role"]` accesses in tests, now `.get()`.)

pkg-py/src/shinychat/_markdown_stream.py
gadenbuie: Done — one pin test per package (pytest, testthat, vitest), each naming the other two locations to update. Also verified the three sets are currently identical: no drift today.

pkg-py/src/shinychat/_chat.py
gadenbuie: Confirmed and deleted — the parts setter with `None` just re-runs the content setter's rebuild, which the preceding line already did. The exercising test is updated; the parts setter itself stays.

pkg-py/src/shinychat/_chat.py
gadenbuie: Went one step further than extraction: `_send_message_parts()` was already structurally identical to `send_wire_segment_actions()`, so we deleted the ladder entirely and route both arms through it (renamed `_send_wire_segment_actions()` — the rename and the unification align the Python and R send paths, which now share the same structure). Wire changes this introduces: homogeneous segments go out as one chunk per segment instead of one collapsed chunk; replace is uniformly wipe-then-append (previously, mixed segments under replace sent per-chunk `replace` actions); empty mid-stream chunks are no longer emitted. Tests re-pinned accordingly.

pkg-py/src/shinychat/_history_client.py
gadenbuie: Done — `NotRequired[str]`, matching how every consumer already reads it. (This surfaced a few direct `turn["role"]` accesses in tests, now `.get()`.)

pkg-py/src/shinychat/_chat_types.py
gadenbuie: Done on both sides: Python's union is now `SegmentPayload`, and TS's string-only type is `StringSegment`. (TS already had `SegmentPayload` for the union, so the names now line up across the wire.)

js/src/chat/state.ts
gadenbuie: Done — `MessageBlock` is now `RenderBlock` throughout the client render model. Member type names and the wire field names (`MessagePayload.segments`, `blocks`) are unchanged.

cpsievert (approved):

PR #377, opened 2026-09-03
fix(js): restore the chat send/stop input icons to their original size and shape

cpsievert:
## Why

The chat send button's arrow icon has been through three states since the
send-button redesign, and never quite landed back where it started:

1. **Before #350**: the icon *was* the button — a single self-contained
   `arrow-up-circle-fill` SVG (a filled circle with the arrow knocked out
   of it), rendered at 24px. Bold, clearly visible.
2. **#350** ("Expose send-button state, icon, and styling hooks") separated
   the button surface from the icon, to support `icon_send` customization
   and CSS styling hooks. As part of that, the bare arrow glyph became
   `arrow-up-short` at 16px. Garrick caught that this read as noticeably
   smaller and bumped it to 18px in the same PR — but `arrow-up-short`'s
   own viewBox has significant empty margin around the arrow (it's
   designed as a compact, inline-text-sized glyph), so the bump only
   partially closed the gap. The arrow still looked considerably smaller
   than before.
3. **This PR**: rather than pick another stock Bootstrap icon and guess
   again, the fix extracts the *exact* arrow contour that was already
   inside the original `arrow-up-circle-fill` path (previously drawn as a
   knockout/hole in that combined shape) and uses it as its own
   standalone glyph. Bumped the icon size to 22px to match. Verified by
   checking out the pre-#350 button, rendering both versions with
   Playwright, and comparing them side by side — the shape, weight, and
   size now match closely.

(#355 was investigated as a possible cause too, but it only touched CSS
variable proxying for `:root` overrides — unrelated to icon size.)

## What changed

- `js/src/utils/icons.ts` — bare arrow glyph now uses the arrow contour
  extracted from `arrow-up-circle-fill` instead of `arrow-up-short`.
- `js/src/chat/chat.scss` — send-button icon-size fallback 18px → 22px
  (this also happens to now match the attach button's existing default).
- `js/src/chat/ChatInput.tsx`, `ChatMessage.tsx` — updated to the renamed
  icon export (shared by the send button and the edit-message resend
  button).
- `js/tests/chat/ChatInput.test.tsx` — updated icon selector.
- Rebuilt `js/dist` and synced into `pkg-r`/`pkg-py`.
- Updated the "18px" icon-size mentions in the R/Python `icon_send` docs
  and regenerated `chat_ui.Rd`.

All existing `icon_send` customization and CSS variable hooks from #350
are untouched — this only changes the *default* glyph and its size.

## Test plan

- [x] `js` unit tests pass (1248 tests)
- [x] Rendered the pre-#350 button and this PR's button via Playwright
      and compared them side by side
- [ ] Reviewer: sanity-check the send button and the edit-message resend
      button in a running app


gadenbuie (approved):
LGTM! I'm assuming you've visually checked it out and like the appearance (I haven't, but the changes look good).

One small question: do we need to do the same thing for the stop icon too since it was a similar change (circle with knocked-out shape → filled shape)?

cpsievert:
> do we need to do the same thing for the stop icon too 

Good catch. Addressed in [35a41a9](https://github.com/posit-dev/shinychat/pull/377/commits/35a41a94b32f7bfb8d01527453ac18fa58294978)

> I'm assuming you've visually checked it out and like the appearance 

Yes, but to be clear, I'm mainly aiming at restoring what we currently have released since the stroke, fill, and size proportions seem sensible. Long term, I think it might be worth us revisiting the actual fill and stroke colors (more transparent/muted?), but I'm not considering that a priority for now.

PR #380, opened 2026-09-03
fix: revert browser-owned chat message state

gadenbuie:
## Summary
- Revert PR #272 semantically rather than by reverting its merge commit. Later history, attachment, and restore changes overlap with #272, so a mechanical revert produced conflicts and would have removed follow-on fixes.
- Restore server-owned message state in the JavaScript client, Python package, and R package. User submissions, appended messages, streamed completions, slash-command echoes, bookmarks, and history now read from or update server state instead of a client-reported transcript.
- Remove the `shinychat.messages` input, the `_messages` bookmark exclusion, and the `ui_offset` bookkeeping. Delete tests that require the retired browser snapshot protocol.
- Keep replay support for UI captured in existing stored history records. New R history records no longer save browser-rendered UI snapshots and instead restore from recorded turns.
- Remove the post-replay R save suppression because it was only needed for delayed browser echoes. With server-side completion saves, it would discard the first real response after restoring a conversation.
- Document the restored Python `chat.messages()` contract in the unreleased changelog because #272 shipped in Python 0.6.1.

## Verification
- `make js-lint`
- `make js-test`
- `make py-check-tests FILTER='bookmark or messages_surfaces_attachments'`
- `make r-check-tests FILTER='history'`

## Follow-up
PR #379 should rebase onto this result and keep its server-side transcript/tree behavior without reintroducing `_messages` or `ui_offset`.


pkg-py/src/shinychat/_chat.py
```diff
@@ -1568,6 +1561,7 @@ async def _restore_bookmark_message(self, message_dict: Any) -> None:
                 stacklevel=2,
             )
             return
+        self._store_message(stored)
```
cpsievert, line 1564: Reverting the protocol stops new client-reported UI from reaching history records, but records saved between #272 and this PR still carry it, and restore re-serves that stored UI as server-trusted — this path sends its `html_deps` along as-is (R does the same in `restore_history_message()`, `chat.R:1111`). A modified client could have forged those deps at save time, so arbitrary script URLs saved into a record would load on replay. #272 never shipped in a tagged release, so only dev-install stores are exposed, but shared-scope stores (e.g. Connect deployments) could hold poisoned records. Can we settle on an approach here — a migration that strips stored `ui` (or at least its `htmlDeps`), or an explicit note accepting the residual? For context, #373's stored-UI version gate discards unversioned stored UI on replay, which would neutralize this once it lands.

pkg-py/CHANGELOG.md
```diff
@@ -42,6 +42,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Bug fixes
 
+* Reverted the browser-owned chat message protocol from #272. Chat message state is again stored on the server, and `chat.messages()` updates when the server receives a user submission, appends a message, or completes a stream.
```
cpsievert: This bullet and the 'Changes' entry above tell opposite stories: line 37 documents #272's browser-sourced protocol as the current behavior, while this bullet reverts it. Since #272 never shipped in a tagged release (v0.6.1 predates it), I'd delete the line-37 entry outright rather than describe both the change and its rollback. The `input[f"{id}_user_input"]` sub-bullet at line 39 is a real change this PR keeps — but its rationale ('co-batch with the message snapshot') no longer applies once the snapshot is gone; reword around the seq-nonce dedup reason.

pkg-r/R/chat_app.R
```diff
@@ -520,17 +520,17 @@ chat_server <- function(
...
+
+    # Preserve the active conversation across the controller replacement.
+    # A completed response has already created a record, while an in-progress
+    # draft has only an allocated ID.
     old_ctrl <- get_session_chat_bookmark_info(
       session,
       paste0(id, ".history-controller")
     )
-    active_id <- if (!is.null(old_ctrl) && is.null(old_ctrl$record)) {
+    active_record <- if (!is.null(old_ctrl)) old_ctrl$record else NULL
```
cpsievert, line 532: Fine as-is — just worth a sentence in the PR description: this swap-time record preservation is new behavior, not part of the revert (pre-#272 had no swap preservation; only the unsaved-draft ID seeding existed). It's needed because saves moved back to stream completion, so a completed response now has a live record at swap time. Without the note, future readers may assume this restored old behavior.

cpsievert (commented):
Verified against the pre-#272 code: the revert is faithful, the deliberate keeps (store format, attachment MIME validation, regular-priority user input) are right, and all three test suites plus pyright pass locally.

One decision before merge (inline on `_restore_bookmark_message()`): records saved while #272 was live still replay their stored UI — including possibly client-forged `htmlDeps` — as trusted. Migrate or document.

Remaining points are doc / dead-code cleanup (not on diff lines, so collected here):

- `pkg-py/src/shinychat/_history.py` — comments in `replay_ui()` (~line 586) and `on_response()` (~line 374) still describe the client report; `_messages_for_bookmark()` now reads the synchronous server accumulator. The logic is right, the explanations are stale.
- `pkg-r/R/chat_history_types.R:148` — `record_ui_count()` has no non-test callers after this revert; delete it with its test block.
- `pkg-r/NEWS.md:58` — the 'co-batch with the client's message snapshot' rationale is stale (behavior kept; reword), and there's no R entry for the rollback or for the removal of the documented `input$<id>_messages`.
- `pkg-r/R/chat_restore.R:115` — drop the `_messages` bookmark exclusion; that input no longer exists.
- `memory-bank/content-rendering.md:54` — the 'HTML Dependencies: Client-Authoritative Round-Trip' section documents deleted machinery (`buildMessagesSnapshot()`, `messages_input_value()`).

Heads-up: two `test-greeting.R` failures reproduce identically on `main` in my local setup — pre-existing, worth a separate issue.

pkg-py/src/shinychat/_chat.py
gadenbuie: Addressed in 5feb6012. History replay now keeps the stored rendered content but removes serialized html_deps before sending the message to the client. R history replay now drops htmlDeps as well, so legacy browser-supplied dependency URLs are not loaded. I added a regression test for the Python path.

pkg-py/CHANGELOG.md
gadenbuie: Addressed in 5feb6012. I removed the browser-owned message-state entry and kept the user-input note with the sequence-nonce deduplication rationale.

pkg-r/R/chat_app.R
gadenbuie: Addressed in 5feb6012. I added the note in pkg-r/R/chat_app.R explaining that swap-time record preservation is new behavior because saves now happen at stream completion.

pkg-py/CHANGELOG.md
```diff
@@ -34,14 +34,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+
+    * The user-submission input (`input[f"{id}_user_input"]`) remains a persistent regular input, allowing sequence-nonce deduplication across submissions. It remains excluded from bookmarks.
+
```
cpsievert:
```suggestion
```

gadenbuie:
Addressed the remaining review feedback in 98e4f8fb:

- Updated stale Python history comments and the HTML dependency documentation.
- Removed the unused R `record_ui_count()` helper and obsolete snapshot-parser tests.
- Removed the `_messages` bookmark exclusion.
- Corrected the R NEWS rationale and documented removal of `input$<id>_messages`.

pkg-r/NEWS.md
```diff
@@ -55,7 +55,9 @@
 
 * The `last_input` reactive returned by `chat_server()` now mirrors the shape of `input$<id>_user_input`: a string when attachments are disabled, and a list of ellmer `Content` objects when enabled.
 
-* `input$<id>_user_input` is now a persistent regular input rather than an event input, so it retains its last submitted value between submissions instead of resetting to `NULL`. This lets it co-batch with the client's message snapshot in a single reactive flush. It remains excluded from bookmarks.
+* `input$<id>_user_input` is now a persistent regular input rather than an event input, so it retains its last submitted value between submissions instead of resetting to `NULL`. This supports sequence-nonce deduplication across submissions. It remains excluded from bookmarks.
+
+* Reverted the browser-owned chat message state introduced during development of #272. Chat messages are again kept on the server, and the internal `input$<id>_messages` input has been removed.
```
cpsievert:
```suggestion
```

pkg-py/src/shinychat/_history.py
```diff
@@ -581,12 +581,24 @@ async def replay_ui(self, record: ConversationRecord) -> None:
...
+                    "segments": [
+                        {
+                            key: value
+                            for key, value in segment.items()
+                            if key != "html_deps"
+                        }
+                        for segment in message_dict.get("segments", [])
+                    ],
+                }
+                await self.chat._restore_bookmark_message(safe_message)
```
cpsievert: Sorry, I didn't review my 🤖 enough before posting the review. 🤦 

I think the guidance it provided leading to https://github.com/posit-dev/shinychat/pull/380/changes/5feb601206b2df5ef0c00f7ba933755e61695f5a was being overly paranoid. 

For the issue it addresses to acutally impact anyone, all 3 of these would have to be true:

1. An app ran from main (never a released version) between Jul 24 and now.
2. Its history store was shared across users, like a Connect deployment. In a per-user store, the forged content only ever replays for the attacker themselves.
3. During the window, a hostile user deliberately saved a message with forged script URLs into that shared store.


That said, I think it's probably best to just revert 5feb601206b2df5ef0c00f7ba933755e61695f5a

PR #381, opened 2026-09-03
Soft-deprecate `chat_ui(messages=)`, error with history enabled

gadenbuie:
## Summary

Startup messages passed via `messages=` can't be recorded by the conversation-history feature, but the previous behavior (deprecation warning + silently dropping content) left hand-rolled-history users without a working capability and without a clear signal.

- **R**: `chat_ui(messages=...)` is soft-deprecated via `lifecycle::deprecate_warn("0.5.0", ...)`: a deprecation warning is emitted and the startup messages still render. `chat_app(messages=...)` now throws an error when `history = TRUE` (the default), since the check is static and fails immediately.
- **Python**: `chat_ui(messages=...)` emits a deprecation warning and still renders. `Chat(messages=...)` raises a `ValueError` when `history` is not `False`.
- The error/warning guides developers to the supported alternatives: use `greeting=` for a startup message, use `chat_append()` / `Chat.append()` to replay messages from the server (replayed messages are captured by the client-side history), or set `history=False` if managing conversation state yourself.
- Docs, examples, vignettes, and man pages updated accordingly.

## Verification

Python: `Chat(messages=["hi"])` raises `ValueError`; `Chat(messages=["hi"], history=False)` warns but works; `chat_ui(messages=["hi"])` warns and renders. Full suite: 574 passed.

R: `chat_ui(messages = list("hi"))` warns via lifecycle and renders; `chat_app(history = TRUE, messages = ...)` errors with guidance to `greeting`. Chat contexts: 1239 passed (snapshot guidance updated).


pkg-py/src/shinychat/_chat.py
```diff
@@ -317,8 +322,18 @@ def __init__(
             raise TypeError("`id` must be a string.")
 
         if messages:
+            if history is not False:
+                raise ValueError(
+                    "`Chat(messages=...)` requires `history=False`: startup "
+                    "messages can't be recorded by the conversation-history "
+                    "feature. Use the `greeting` parameter for a startup "
+                    "message, use `.append()` to replay messages from the "
```
```suggestion
                    "message, use `.append_message()` to replay messages from the "
```

pkg-py/src/shinychat/_chat.py
```diff
@@ -317,8 +322,18 @@ def __init__(
...
+                    "server, or set `history=False` if you're managing "
+                    "conversation state yourself."
+                )
             warn_deprecated(
-                "`Chat(messages=...)` is deprecated. Use `.ui(messages=...)` instead."
+                "`Chat(messages=...)` is deprecated and will be removed in a "
+                "future release."
```
```suggestion
                "future release. Use `.append_message()` instead."
```

pkg-py/src/shinychat/_chat.py
```diff
@@ -2304,10 +2319,16 @@ def ui(
         Parameters
         ----------
         messages
-            A sequence of messages to display in the chat. Each message can be either a
-            string or a dictionary with `content` and `role` keys. The `content` key
-            should contain the message text, and the `role` key can be "assistant" or
-            "user".
+        Deprecated. Startup messages can't be recorded by the
+        conversation-history feature. Use ``greeting`` for a startup
+        message, ``.append()`` to replay messages from the server, or set
```
```suggestion
        message, ``.append_message()`` to replay messages from the server, or set
```

pkg-py/src/shinychat/_chat.py
```diff
@@ -2304,10 +2319,16 @@ def ui(
...
+        ``history=False`` on the ``Chat`` if you're managing conversation
+        state yourself.
+
+        A sequence of messages to display in the chat. Each message can be
+            either a string or a dictionary with `content` and `role` keys. The
+            `content` key should contain the message text, and the `role` key
+            can be "assistant" or "user".
```
cpsievert: I'd be open to removing basically the entire messages description, especially if we document them well already at `.append_message()`. I feel like we could even get rid of the part about recommending what to do instead if our errors capture that well.

pkg-py/src/shinychat/_page_chat.py
```diff
@@ -176,7 +176,8 @@ def page_chat(
         Shiny's ``"shiny"`` preset. Pass a :class:`shiny.ui.Theme` directly to
         use another preset or a completely custom theme.
     messages
-        Initial chat messages. See :func:`~shinychat.chat_ui`.
+        Deprecated. Startup messages can't be recorded by the
+        conversation-history feature. See :func:`~shinychat.chat_ui`.
```
cpsievert: Let's remove the param entirely from `page_chat()`?

pkg-py/src/shinychat/_chat.py
```diff
@@ -2730,10 +2757,21 @@ def chat_ui(
...
+    from shiny._deprecated import warn_deprecated
     from shiny.module import resolve_id
     from shiny.ui.css import as_css_unit
     from shiny.ui.fill import as_fill_item, as_fillable_container
 
+    if messages:
+        warn_deprecated(
+            "`chat_ui(messages=...)` is deprecated. Startup messages can't "
+            "be recorded by the conversation-history feature. Use "
+            "`greeting=` for a startup message, `Chat.append()` to replay "
```
```suggestion
            "`greeting=` for a startup message, `Chat.append_message()` to replay "
```

pkg-py/src/shinychat/_chat.py
gadenbuie: Done. Updated the deprecation warning to recommend `.append_message()`.

pkg-py/src/shinychat/_chat.py
gadenbuie: Done. Updated the guidance to use `.append_message()`.

pkg-py/src/shinychat/express/_page_chat.py
```diff
@@ -189,7 +184,6 @@ def page_chat(
     chat_root = _create_page_chat_root(
         id=id,
         drawer=drawer,
-        messages=messages,
```
cpsievert, line 192: I think messages could also be dropped from `def _create_page_chat_root` now

cpsievert (approved):

pkg-py/src/shinychat/express/_page_chat.py
gadenbuie: Addressed in d586a609 (already on this branch): removed `messages` from `_create_page_chat_root()` and stopped forwarding it. The explicit `page_chat(messages=...)` guard remains to provide a clear migration error.

pkg-py/src/shinychat/_chat.py
```diff
@@ -282,7 +282,12 @@ async def handle_user_input(user_input: str):
         client with empty turns is passed so the greeting can be LLM-generated
         without polluting conversation history.
     messages
-        Deprecated. Use `chat.ui(messages=...)` instead.
+        Deprecated. Startup messages can't be recorded by the
+        conversation-history feature, so ``messages`` requires
```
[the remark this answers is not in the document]
gadenbuie: Fixed in 53ab8d4b: the Python `Chat` and `chat_ui` docstrings now explicitly say the history restriction/deprecation guidance applies to non-empty startup messages. Focused Python tests pass (3 tests).

pkg-r/R/chat_app.R
```diff
@@ -181,6 +177,12 @@ chat_app <- function(
   check_ellmer_chat(client)
   dots <- rlang::dots_list(...)
   check_chat_app_dots(dots)
+  if (!isFALSE(history) && "messages" %in% rlang::names2(dots)) {
```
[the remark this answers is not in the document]
gadenbuie: Fixed in 53ab8d4b: `chat_app()` now rejects startup messages only when `dots$messages` is non-NULL, so `messages = NULL` behaves like omitting the argument. Added regression coverage; the focused R test file passes (38 tests).

PR #386, opened 2026-09-05
fix: silently drop NULL fields in tool result display lists

cpsievert:
## Problem

`as_tool_result_display()` warns on `NULL` field values, but named NULLs are an idiomatic R byproduct of conditionally including a field:

```r
# Tool author wants a title only for some actions
display = list(
  title = if (action == "update") title,  # NULL when action != "update"
  markdown = md,
  open = TRUE
)
```

This produces a named `NULL` (`list(title = NULL, ...)`), which fails field validation:

```
Warning: Invalid `@extra$display` format for `ContentToolResult` from `querychat_query()` ...
x title must be a single string or HTML content, not NULL; ignoring.
```

Seen in the wild with querychat's `querychat_query` tool.

## Repro

```r
shinychat:::as_tool_result_display(list(title = NULL, open = TRUE))
# Before: warning "title must be a single string or HTML content, not NULL; ignoring."
# After:  no warning; title treated as absent
```

## Fix

Drop `NULL` fields at the top of `as_tool_result_display()`, before unknown/invalid field validation. Genuinely invalid values (wrong type, `NA`, non-scalar) still warn as before.

### DIFF
--- a/pkg-r/R/contents_shinychat.R
+++ b/pkg-r/R/contents_shinychat.R
@@ -836,8 +836,6 @@ as_tool_result_display <- function(display, error_context = NULL) {
     return(structure(list(), class = "shinychat_tool_result_display"))
   }
 
-  # Named NULLs arise from idiomatic conditional fields like
-  # `list(title = if (cond) title)`; treat them as absent.
   display <- display[!map_lgl(display, is.null)]
 
   unknown <- setdiff(names(display), tool_result_display_fields)
--- a/pkg-r/R/contents_shinychat.R
+++ b/pkg-r/R/contents_shinychat.R
@@ -836,6 +836,10 @@ as_tool_result_display <- function(display, error_context = NULL) {
     return(structure(list(), class = "shinychat_tool_result_display"))
   }
 
+  # Named NULLs arise from idiomatic conditional fields like
+  # `list(title = if (cond) title)`; treat them as absent.
+  display <- display[!map_lgl(display, is.null)]
+
   unknown <- setdiff(names(display), tool_result_display_fields)
   if (length(unknown) > 0) {
     cli::cli_warn(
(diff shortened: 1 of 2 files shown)


cpsievert:
Additional context on why this is worth taking before the next release:

In the currently released version (r/v0.4.0), `get_tool_result_display()` only checks `is.list(display)` — there is no per-field validation, so `list(title = NULL, ...)` passes through silently and the `NULL` is ignored downstream.

The strict per-field validation that rejects `NULL` (`title must be a single string or HTML content, not NULL`) was introduced in 352e025d (#283), which is not contained in any R release tag. So the warning this PR removes is a regression relative to released behavior, but one that only exists on main and has never shipped.

Net effect of this PR: the next release preserves v0.4.0's tolerance for `NULL` fields, while keeping the new (useful) warnings from #283 for genuinely invalid values (wrong types, `NA`, non-scalars).

pkg-r/R/contents_shinychat.R
```diff
@@ -836,6 +836,10 @@ as_tool_result_display <- function(display, error_context = NULL) {
     return(structure(list(), class = "shinychat_tool_result_display"))
   }
 
+  # Named NULLs arise from idiomatic conditional fields like
+  # `list(title = if (cond) title)`; treat them as absent.
```
cpsievert:
```suggestion
```

pkg-r/tests/testthat/test-contents_shinychat.R
```diff
@@ -1009,6 +1009,28 @@ test_that("as_tool_result_display() rejects non-scalar and NA logicals", {
   expect_null(res$show_request)
 })
 
+test_that("as_tool_result_display() silently drops NULL fields", {
+  # `list(title = if (cond) title)` is an idiomatic way for tool authors to
+  # conditionally include a field; it yields a named NULL that should be
+  # treated as absent, not as an invalid value.
```
cpsievert:
```suggestion
```

gadenbuie:
Thanks for catching and fixing this!

Issue #391, opened 2026-09-05
Conversation history: support R and Python workers without a Shiny session


JamesHWade:
I’m looking at the history work in #379 and wondering whether both the Python implementation and planned R port could support workers without an active Shiny session.

In [Deputy](https://github.com/JamesHWade/deputy), I’d like to let someone select a conversation branch in a Shiny app, continue it in a background R worker, and see the saved continuation when they return. I’d pass the user, chat and conversation identifiers to the worker from the app.

My immediate use case is R with Deputy and ellmer, but I’d like the same workflow to be available in Python with chatlas.

The separation of display history and provider turns looks useful here. I may summarize older turns for the model while keeping the original conversation available. I’d also like to save a Deputy run ID through the planned application-state support.

Would this fit the history redesign? I’d like to reuse the history operations and validation you’re building.

This also relates to #307’s conversation identity work and #61’s distinction between UI content and model turns. I’m tracking the Deputy integration in JamesHWade/deputy#66.

PR #66, opened 2025-06-03
test(examples): Add some examples leveraging pydantic-AI

karangattu:
This pull request introduces several new Shiny applications powered by `Pydantic AI` agents. Each application is designed for a specific use case, such as chat interfaces, interactive adventures, structured output handling, tool calling, and personalized workout plan generation.

PR #392, opened 2026-09-05
Prototype conversation history in sessionless Python workers

JamesHWade:
Refs #391. Stacked on #379, targeting `feat/history-exchange-tree`.

A Python worker can restore the conversation branch selected in Shiny, continue it without a session, and save the result for the user's return. The worker uses the existing exchange record, store, turn adapter, and shared capture/restore implementation. Provider turns remain independent of display messages, so compaction preserves earlier display history and alternative branches.

This remains a private prototype at `shinychat._conversation.Conversation`. It is not exported from `shinychat`, advertised in the changelog, or included in published documentation. #379 must establish public v2 enablement before this becomes a supported API. The [fixture README](https://github.com/JamesHWade/shinychat/blob/codex/sessionless-history-workers/pkg-py/tests/playwright/chat/history_worker/README.md) documents the draft setup, worker lifecycle, recovery, and corresponding R behavior contract.

Input is saved before work begins; complete assistant messages are checkpointed as they arrive; context exit captures provider turns and terminal status. Workers reject incompatible or incomplete provider state and support explicit recovery from an earlier exchange. Failed output saves can be retried without duplicating messages.

Before each write, the handle compares canonical JSON for the stored record with the snapshot last loaded or successfully saved. A change or deletion raises `ConversationConflictError`. This catches values-only changes even when `updated_at` is unchanged, preserves JSON boolean/number distinctions, and keeps the expected snapshot intact after a failed write. It detects already-stale handles; the comparison and write are not atomic. The caller still owns authorization, scheduling, and exclusive access for the entire load/work/save operation. Each returning Shiny session needs a fresh file store to avoid a stale process-local history list.

Python is implemented; the R/ellmer worker use case still requires the planned R exchange-tree port and a supported R worker API. The downstream integration is tracked in [Deputy #66](https://github.com/JamesHWade/deputy/issues/66). Application `values` uses the existing conversation-level callbacks, with consistent `ValueError` validation; exchange-local application-state hooks and multi-writer coordination remain separate work. #391 stays open.

Validation at `c57099c`:

- 1,027 Python tests pass, including stale saves across two handles, changed values and branches, deletion, changes during work, failed-write retries, JSON type distinctions, invalid values, attachments, and direct rejection of pending selections.
- All 9 affected worker/v2-restore browser tests pass using installed Chrome. The worker test closes the Shiny session, invokes a separate Python process using real chatlas with a mocked HTTP boundary, then reopens the app and verifies display history, compacted provider turns, run metadata, and preserved alternative branches.
- Ruff and Pyright pass. Wheel/sdist and API/site documentation builds pass; Quarto emits existing site warnings. Artifact checks confirm the implementation is packaged but absent from public exports and published docs.
- Independent review of the remediation has no remaining actionable findings.

The extraction, chatlas system-prompt restoration fixes, and switch-test synchronization remain separate commits. Changes needed to reconcile #379 with main belong on that base first; this feature remains stacked on it. This PR remains a draft. GitHub CI status is reported by the checks below.

### DIFF
--- a/pkg-py/docs/history-workers.qmd
+++ b/pkg-py/docs/history-workers.qmd
@@ -1,0 +1,94 @@
+---
+title: Conversation history in a worker
+---
+
+`Conversation` lets a worker continue the branch selected in a Shiny app and save the result for the user's next visit. It uses the exchange-tree history format proposed in [#379](https://github.com/posit-dev/shinychat/pull/379). This is a draft API for [#391](https://github.com/posit-dev/shinychat/issues/391); it requires v2 history and does not load legacy v1 records.
+
+The app passes its store location, resolved chat ID, trusted owner scope, and conversation ID to the worker. The worker creates its own model client. No `Chat`, Shiny session, or reactive context is needed.
+
+```python
+from chatlas import ChatOpenAI
+from shinychat import Conversation
+from shinychat.types import ConversationPartition, FileConversationStore
+
+
+async def continue_conversation(directory, scope, chat_id, conversation_id, run_id):
+    client = ChatOpenAI(model="gpt-4.1")
+    conversation = await Conversation.load(
+        FileConversationStore(directory),
+        ConversationPartition(chat_id=chat_id, scope=scope),
+        conversation_id,
+        client=client,
+    )
+
+    prompt = "Continue the analysis and write a brief conclusion."
+    async with conversation.exchange(prompt):
+        conversation.values["deputy"] = {"version": 1, "run_id": run_id}
+        response = await client.chat_async(prompt, echo="none", stream=False)
+        await conversation.append_message(await response.get_content())
+```
+
+The caller runs this function in its chosen worker system. The caller also authorizes the scope and serializes access to the conversation for the entire operation. The file store does not coordinate concurrent writers. A disconnected browser is not sufficient if the server is still processing an earlier response; the app must finish or hand off that work first. Use an explicit shared directory, not a session-dependent default store location. Create a fresh store when returning to the conversation so process-local list caches do not hide another process's changes.
+
+When the user returns, ordinary Shiny history restoration replays the saved messages and restores the model's turns. Application values use the existing callbacks:
+
+```python
+# Within the Shiny server function, after creating chat:
+from shiny import reactive
+
+run_id = reactive.Value(None)
+
+
+@chat.history.on_restore
+def restore_run(values):
+    state = values.get("deputy")
+    if state is not None and state.get("version") != 1:
+        raise ValueError("Unsupported Deputy history values")
+    run_id.set(state["run_id"] if state is not None else None)
+
+
+@chat.history.on_save
+def save_run(values):
+    values["deputy"] = {"version": 1, "run_id": run_id.get()}
+```
+
+## Branches and model context
+
+`Conversation.load()` restores the saved active branch. `conversation.active_leaf` identifies its selected exchange. To create an alternative continuation, select an earlier exchange and submit a new input:
+
+```python
+await conversation.select(previous_exchange_id)
+async with conversation.exchange("Try a different approach"):
+    # Invoke the client and append the resulting Markdown here.
+    ...
+```
+
+Selection restores provider turns before work starts. Existing descendants remain in the tree. Retain an exchange ID returned by the `exchange()` context or `active_leaf` to select it later; callers do not need to modify tree fields.
+
+Display content and provider turns are independent. A worker can replace or summarize the client's turns **inside** the exchange context. Capture stores a snapshot when the original prefix changes. Earlier messages and branches remain intact. Append only the Markdown that should appear to the reader; provider tool turns and other context are saved through the client's turn adapter. New worker output currently supports complete Markdown messages and user attachments. Existing custom display messages are preserved without re-rendering them in the worker.
+
+## Persistence boundaries
+
+The context saves the user input before yielding. Each `append_message()` saves another complete message. Context exit captures provider turns and saves the terminal status. Exceptions and cancellation propagate to the caller, while saved errors use the same bounded summaries as Shiny history. A killed process leaves its latest checkpoint pending. Persistence failures propagate too; they do not report successful completion.
+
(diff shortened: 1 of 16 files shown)


pkg-py/src/shinychat/__init__.py
```diff
@@ -1,6 +1,7 @@
 from ._attachments import Attachment, attachment_to_content
 from ._chat import Chat, UserInput, chat_greeting, chat_ui
 from ._chat_normalize import message_content, message_content_chunk
+from ._conversation import Conversation
```
JamesHWade: **Resolve v2 enablement before publishing this API.** `Conversation` is exported and documented, but it accepts only v2 records while the app still defaults to `_EXCHANGE_TREE_HISTORY_V2 = False`. The browser fixture enables it by modifying that private module flag. A reader following the worker page therefore cannot load history produced by the documented app setup. Keep the export and published documentation private until #379 supplies the supported enablement path, or explicitly make this experimental and document a complete, runnable setup. Cover that documented setup in the app-to-worker test.

pkg-py/src/shinychat/_conversation.py
```diff
@@ -0,0 +1,287 @@
...
+        alone does not capture model work or complete an exchange.
+        """
+        await self._save_record(self._record)
+
+    async def _save_record(self, record: ConversationRecordV2) -> None:
+        _validate_mapping_keys(record.values)
+        json.dumps(record.values, allow_nan=False)
+        # Store implementations may retain references. Give them a snapshot,
+        # so further work cannot change a saved checkpoint without a put().
+        await self._store.put(self._partition, record.model_copy(deep=True))
```
JamesHWade, line 334: **Reject stale handles before replacing the stored conversation.** Two handles can load the same conversation, one can save new values or a branch, and this `put()` then silently replaces that newer record with the other's older copy. I reproduced this with application values; those saves do not advance `updated_at`, so a timestamp-only comparison would still lose updates. Compare against the record last loaded or successfully saved (or a revision that every writer advances), treating deletion as a conflict too, and retain the expected state when `put()` fails. Test competing handles, including a values-only change with an unchanged timestamp. A separate `get()` followed by `put()` can detect an already-stale handle but cannot close the intervening write race; retain the exclusive-access requirement and do not describe this as multi-writer protection.

pkg-py/docs/history-workers.qmd
```diff
@@ -0,0 +1,96 @@
...
+| Select | Restore an existing exchange's provider state; preserve descendants. The next user input creates a child of the selected exchange. |
+| Exchange | Persist accepted input, checkpoint display messages, then capture provider state and terminal outcome. Preserve and propagate errors and cancellation. |
+| Compact | Capture a snapshot when model turns replace their earlier prefix, preserving the original display transcript. |
+| Application values | Share the existing conversation-level callback payload. Application keys and payload versions remain application-owned. |
+| Validation | Use the same record/state validation as the UI before mutating the client. Reject incompatible provider state for unattended continuation. |
+| Concurrency | Caller owns authorization, scheduling, and exclusive access for the full operation. No implicit worker queue or multi-writer guarantee. |
+
+R and Python share behavior, not provider-turn encodings or cross-language stored records. Deputy's integration remains gated on a supported, released R API, as tracked in [Deputy #66](https://github.com/JamesHWade/deputy/issues/66).
```
JamesHWade: **Keep downstream integration tracking in the PR description.** Retain the explanation that R and Python share behavior rather than serialized records, but move the Deputy #66 link and its project-specific release gate out of shipped package documentation. The package page should describe shinychat's supported contract; the PR description can explain the downstream motivation and remaining integration work.

JamesHWade (commented):
Review of c523c94, compared with `feat/history-exchange-tree`.

The shared capture/restore implementation is a sound basis for this draft. Keep this PR stacked on #379: it is six commits ahead of its base and none behind. The base is 13 commits behind main; update that base separately before rebasing this feature.

Before exposing this as a supported workflow, address the public v2 enablement gap and reject writes from stale worker handles. I reproduced the overwrite with two independently loaded handles. Application-values-only saves also leave `updated_at` unchanged, so comparing that timestamp alone would miss this case. Any read-before-write guard still requires exclusive access; it is not an atomic concurrency guarantee.

Move the downstream Deputy issue link from package documentation into this PR description. Minor follow-ups are consistent JSON-validation errors, attachment persistence coverage, and direct coverage of selecting a pending exchange. Keep the app-side file-store cache limitation explicit. The system-prompt fixes and switch-test synchronization would also benefit #379 independently.

This remains a partial implementation of #391: Python worker continuation is implemented, while the R/ellmer worker use case still requires the R port. Keep the PR in draft and #391 open.

Verification for this review: inspected the current diff and base relationship; reproduced stale-handle overwrite, unchanged timestamps on values-only saves, and the different validation exceptions for NaN and datetime. The full test counts in the existing PR description were not rerun for this review.

pkg-py/src/shinychat/__init__.py
JamesHWade: Fixed in c57099c. Conversation is now private: removed the package export, public documentation entries, and changelog claim. The guide moved beside the browser fixture and explicitly describes its private v2 setup. The worker browser round trip passes; artifact checks confirm the API is absent from public exports and the published site.

pkg-py/src/shinychat/_conversation.py
JamesHWade: Fixed in c57099c. Every worker save compares canonical JSON with the snapshot last loaded or successfully saved; conflicts and deletion raise ConversationConflictError. This catches unchanged-timestamp values updates and distinguishes true from 1. Failed puts retain the expected snapshot. Tests cover both stores, competing handles, newer branches, deletion, changes during work, and retry behavior. The documented whole-operation exclusive-access requirement remains because get and put are not atomic.

pkg-py/docs/history-workers.qmd
JamesHWade: Fixed in c57099c. Removed the downstream issue link and release gate from the guide; Deputy #66 is now referenced in this PR description. The guide itself moved out of published package documentation into the draft worker fixture README.

PR #394, opened 2026-09-07
fix: restore saved file attachments

gadenbuie:
Follow-up to #373

## Summary
- Preserve attachment payloads when Python derives stored history UI from turns.
- Retain R upload payloads until history persistence so restored user messages keep their rich attachment previews.
- Add Python and R regressions for attachment-history serialization.

## Verification
- `uv run pytest pkg-py/tests/test_history_controller.py -k 'extend_preserves_attachment_payload_from_ui_snapshot or replay'`
- `make r-check-tests FILTER='attachments'`
- `make r-check-tests FILTER='chat_history_types'`
- `make r-check-format`
- `make py-check-types`


cpsievert (approved):
Thanks, the linked issue appears to be incorrect, but otherwise LGTM!

PR #399, opened 2026-09-08
Add new_chat for conversation history

### LINKED ISSUE #397: `chat$clear()` doesn't coordinate with the history controller: conversations are lost and new chats get grafted onto the previous record
## Summary

With conversation history enabled, the chat handle's `clear()` method (backed by `client_clear()` in `chat_server()`) clears the chat UI and the ellmer turns, but never touches the `HistoryController`. This makes `clear()` unusable as a "start fresh" action: the conversation currently on screen is silently dropped from the history, and the controller remains attached to the old record so that subsequent saves corrupt it.

The history drawer's own "New chat" button already does the right thing — the `{id}_history_new` observer in `chat_enable_history()` calls `controller$new_chat()` — but there is no public path for an app author to do the same. `chat$clear()` is the only public "reset the chat" method, and it leaves the chat in an inconsistent state when history is on.

There is a second public surface with the same problem: the exported `shinychat::chat_clear()` clears the visible UI (and optionally resets the greeting) but also leaves the history controller — and the ellmer turns — untouched, so direct callers desync the chat the same way.

## Current behavior

`client_clear()` does exactly two things:

1. `chat_clear(id, ...)` — clears the chat UI
2. `client$set_turns(list())` (when `client_history = "clear"`) — clears the ellmer turns

It does not call `save_current()`, `clear_active()`, or `send_history_update()` on the history controller. And the exported `chat_clear()` itself is a UI-level primitive: it sends a `clear` action to the chat element and manages the greeting state, nothing more.

## Failure modes

**1. The conversation is never saved.** Records reach the store only through the controller (`save_current()` / `on_response()`). `clear()` discards the turns without a final save, so the conversation that was just on screen either never appears in the history drawer or is truncated to whatever was saved before the clear.

**2. The next conversation is saved into the previous record.** The controller keeps its `active_id` and in-memory `record`. After the clear, `on_response()` early-returns while `length(recorded_turns) <= record_turn_count(self$record)`, so the new conversation isn't recorded at all until it outgrows the old one — at which point `extend_record_linear()` grafts the new turns onto the old record. The drawer then shows two spliced conversations in one entry.

**3. `on_active_id_change(NULL)` never fires.** Listeners keyed to the active conversation (in btw, the token and cost counters) can't detect that a new conversation started, so they keep the old conversation's totals.

## Repro

Environment: shinychat 0.4.0.9000 (dev, toward 0.5.0), `page_chat()` + `chat_server()` with history enabled.

```r
library(shiny)
library(bslib)
library(ellmer)
library(shinychat)

ui <- page_chat(
  "Chat",
  id = "chat",
  toolbar = toolbar(toolbar_input_button("clear_btn", "Clear"))
)

  chat <- chat_server("chat", client = ellmer::chat_openai())
  observeEvent(input$clear_btn, chat$clear())
}

shinyApp(ui, server)
```

1. Chat for a few turns. The conversation appears in the history drawer.
2. Press Clear. The UI wipes.
3. Check the drawer: the conversation is missing or stops at the last save before the clear — never reflecting the full conversation that was on screen.
4. Send a message in the fresh chat. Nothing new appears in the drawer; once the new conversation exceeds the old record's turn count, its turns are appended to the old record's entry.

Calling `shinychat::chat_clear("chat")` directly produces the same desync (with the ellmer turns also still intact).

## Proposed behavior

When history is enabled, `chat$clear()` should behave like the drawer's "New chat" button — i.e., what `HistoryController$new_chat()` already does:

- `save_current()` first, so the conversation lands in the store
- `client$set_turns(list())` + `chat_clear()` (the current behavior)
- `clear_active()` + `on_active_id_change(NULL)`, so the next message starts a fresh record
- `notify_settled(FALSE)` + `send_history_update()`

With history disabled (no controller), `clear()` keeps today's behavior. The cleanest implementation may be for `client_clear()` to delegate to `controller$new_chat()` when a controller exists, since `new_chat()` already performs the turn and UI clearing internally.

## Which layer should own the history reset?

`chat$clear()` is the only layer that sees all three pieces of chat state (the UI, the ellmer turns, and the history record), so it's the natural owner of the full reset. That leaves two options for the exported `shinychat::chat_clear()`:

1. **Keep it a UI-only primitive, and document it as such.** Its signature (`id`, `greeting`, `session`) has no access to the client or the controller, and callers who want a full reset have `chat$clear()`. A doc note saying "clears the visible messages only — use the chat handle's `clear()` to also reset turns and history" would prevent the trap.
2. **Make it history-aware too**, so direct callers can't desync. But since it also doesn't touch the ellmer turns, coordinating history there produces a half-reset (fresh record, stale turns), which is arguably worse than clearly-scoped behavior.

We lean toward (1): fix the composed operation (`chat$clear()`), keep `chat_clear()` honestly scoped as UI-level, and document the difference. But this is your call — the main point is that today both public clear surfaces desync history, and at least one of them should do the full reset.

## Should there be an opt-out?

We considered an argument like `history = c("save", "discard", "none")`, but we don't think one is needed initially:

- The "discard this conversation" case is already served by `controller$delete()` on the saved record — a second overlapping mechanism on `clear()` would be API surface without a demonstrated need.
- Saving matches the "new chat" mental model from every other chat app; a default that silently drops the record looks like a bug, and the natural desire to reach for "don't save" is really "delete it from the drawer", which exists.

If a real need emerges later, an opt-out can be added with the current default preserved.

One naming note: `client_history = c("clear", "set", "append", "keep")` already exists on `clear()` and refers to the ellmer *turns*, not the conversation history store. Any new argument needs a different name, and the docs may want to disambiguate the two "history" concepts regardless.

---

Since history ships in 0.5.0 and hasn't reached CRAN yet, now is the cheapest moment to fix the default.

gadenbuie:
Fixes #397

## Summary

- Add `new_chat()` to the managed R and Python chat handles. It saves the active conversation, clears the rendered messages and client turns, resets the active conversation ID, and refreshes the history drawer.
- Keep `clear()` as the lower-level UI/client-turn reset. It errors when conversation history is enabled and directs callers to `new_chat()`.
- Keep the low-level UI clear APIs UI-only, and document the distinction.
- Reject `new_chat()` while a response streams in R and Python, so a late response cannot enter the new conversation.

## Verification

- `make r-check`
- `make py-check`

### DIFF
--- a/pkg-py/src/shinychat/_chat_client.py
+++ b/pkg-py/src/shinychat/_chat_client.py
@@ -100,17 +100,11 @@ class ChatClient:
         """
         Clear the chat and optionally reset the client's turn history.
 
-        When conversation history is enabled, the default operation starts a
-        new conversation through the history controller. The current record
-        is saved before the client turns, rendered messages, active
-        conversation ID, and history drawer state are reset. In that mode,
-        ``messages`` and the non-default ``client_history`` modes are not
-        supported because they can make the rendered chat, client turns, and
-        persisted conversation disagree.
-
-        When conversation history is disabled, this method retains its
-        existing behavior and ``client_history`` controls only the underlying
-        client's turns.
+        This is a lower-level operation for chats without conversation
+        history. ``client_history`` controls only the underlying client's
+        turns. When conversation history is enabled, use :meth:`new_chat`
+        instead so the rendered chat, client turns, and persisted conversation
+        remain synchronized.
 
         Parameters
         ----------
@@ -119,14 +113,11 @@ class ChatClient:
             ``client_history`` is ``"set"`` or ``"append"``, and not allowed
             with ``"clear"`` or ``"keep"``.
         greeting
-            Passed to :meth:`~shinychat.Chat.clear_messages`. With history
-            enabled, ``True`` also causes the configured greeting to be
-            resolved for the new conversation.
+            Passed to :meth:`~shinychat.Chat.clear_messages`.
         client_history
             How to handle the client's turn history:
 
-            * ``"clear"`` (default): removes all turns from the client. With
-              history enabled, this starts a new saved-history conversation.
+            * ``"clear"`` (default): removes all turns from the client.
             * ``"set"``: sets the client's turns to ``messages``.  Requires
               ``messages`` to be provided.
             * ``"append"``: appends ``messages`` to the client's existing turns.
@@ -152,17 +143,10 @@ class ChatClient:
             getattr(self._chat, "history", None), "_controller", None
         )
         if history_controller is not None:
-            if messages is not None:
-                raise ValueError(
-                    "`messages` cannot be supplied when conversation history "
-                    "is enabled; use `chat.client.clear()` to start a new "
-                    "conversation."
-                )
-            if client_history != "clear":
-                raise ValueError(
-                    '`client_history` must be "clear" when conversation '
-                    "history is enabled."
-                )
+            raise ValueError(
+                "Can't clear a chat with conversation history enabled; use "
+                "`await chat.client.new_chat()` to start a new conversation."
+            )
         if client_history in ("set", "append") and messages is None:
             raise ValueError(
                 f"`messages` must be provided when `client_history='{client_history}'`."
@@ -181,10 +165,6 @@ class ChatClient:
                 sanitize=False,
             )
 
-        if history_controller is not None:
-            await history_controller.new_chat(greeting=greeting)
-            return
-
         await self._chat.clear_messages(greeting=greeting)
 
         if messages is not None:
@@ -202,6 +182,51 @@ class ChatClient:
             turns = self._client.get_turns() + messages_to_turns(messages)
             self._client.set_turns(turns)
         # "keep" → do nothing
+
+    async def new_chat(self, *, greeting: bool = False) -> None:
+        """
+        Save the current conversation and start a new one.
+
+        This method is available only when conversation history is enabled.
+        It saves the current record, clears the client turns and rendered
+        messages, resets the active conversation ID, and updates the history
+        drawer.
+
+        Parameters
+        ----------
+        greeting
+            Passed to :meth:`~shinychat.Chat.clear_messages`. When ``True``,
+            also clears the greeting and resolves the configured greeting for
+            the new conversation.
+
+        Raises
+        ------
+        ValueError
+            If conversation history is not enabled.
+        shiny.types.NotifyException
+            If an assistant response is currently streaming. To avoid this
+            error, guard the call by checking
+            ``chat.latest_message_stream.status() != "running"`` before
+            calling :meth:`new_chat`.
+        """
+        history_controller = getattr(
+            getattr(self._chat, "history", None), "_controller", None
+        )
+        if history_controller is None:
+            raise ValueError(
+                "Can't start a new chat without conversation history enabled; "
+                "use `await chat.client.clear()` instead."
+            )
+        if self._chat.latest_message_stream.status() == "running":
+            from shiny.types import NotifyException
+
+            raise NotifyException(
+                "Can't start a new chat while a response is still being "
+                "generated. Please wait for it to finish or stop it first.",
+                sanitize=False,
+            )
+
+        await history_controller.new_chat(greeting=greeting)
 
 
 def messages_to_turns(
--- a/pkg-py/src/shinychat/_chat_client.py
+++ b/pkg-py/src/shinychat/_chat_client.py
@@ -139,9 +139,7 @@ class ChatClient:
                 f"`client_history={client_history!r}` is not valid. Expected "
                 'one of "clear", "set", "append", or "keep".'
             )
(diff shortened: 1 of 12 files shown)


pkg-py/src/shinychat/_chat_client.py
```diff
@@ -133,6 +148,21 @@ async def clear(
                 f"`client_history={client_history!r}` is not valid. Expected "
                 'one of "clear", "set", "append", or "keep".'
             )
+        history_controller = getattr(
+            getattr(self._chat, "history", None), "_controller", None
+        )
```
cpsievert:
```suggestion
        history_controller = self._chat.history._controller
```

pkg-py/src/shinychat/_chat_client.py
```diff
@@ -151,6 +181,10 @@ async def clear(
                 sanitize=False,
             )
 
+        if history_controller is not None:
+            await history_controller.new_chat(greeting=greeting)
+            return
```
cpsievert: I think it would make sense to bring this up into the first check that `history_controller is not None`?

pkg-py/src/shinychat/_chat_client.py
```diff
@@ -107,11 +119,14 @@ async def clear(
...
-            Passed to :meth:`~shinychat.Chat.clear_messages`.
+            Passed to :meth:`~shinychat.Chat.clear_messages`. With history
+            enabled, ``True`` also causes the configured greeting to be
+            resolved for the new conversation.
         client_history
             How to handle the client's turn history:
 
-            * ``"clear"`` (default): removes all turns from the client.
+            * ``"clear"`` (default): removes all turns from the client. With
+              history enabled, this starts a new saved-history conversation.
```
cpsievert: I find this phrasing:

> starts a new saved-history conversation.

A bit ambiguous. Maybe we could be a little bit more explicit that this is essentially maps to `chat.history.new_chat()`?

pkg-py/src/shinychat/_chat_client.py
```diff
@@ -169,6 +183,51 @@ async def clear(
...
+            If conversation history is not enabled.
+        shiny.types.NotifyException
+            If an assistant response is currently streaming. To avoid this
+            error, guard the call by checking
+            ``chat.latest_message_stream.status() != "running"`` before
+            calling :meth:`new_chat`.
+        """
+        history_controller = getattr(
+            getattr(self._chat, "history", None), "_controller", None
+        )
```
cpsievert:
```suggestion
        history_controller = self._chat.history._controller
```

pkg-py/src/shinychat/_chat_client.py
```diff
@@ -133,6 +139,14 @@ async def clear(
...
                 'one of "clear", "set", "append", or "keep".'
             )
+        history_controller = getattr(
+            getattr(self._chat, "history", None), "_controller", None
+        )
+        if history_controller is not None:
+            raise ValueError(
+                "Can't clear a chat with conversation history enabled; use "
+                "`await chat.client.new_chat()` to start a new conversation."
+            )
```
cpsievert, line 147: I still think the concept of changing messages (& turns) for the currently active conversation is going to be useful (e.g., compacting a conversation). 

This is also why (privately, via Slack) I was advocating for deprecating `.clear()` in favor of a new `.set_messages()` method, which would basically be the same as the current `.clear()`, but assumes you are going to provide `messages`

gadenbuie: Are you okay moving forward without working this out yet? I'm on board with the general idea but I think we'll need to think about it a little more carefully and make sure we're getting the API right in tandem with ellmer/chatlas APIs. 

One small detail, I have generally aligned on messages being UI chat state and turns being client chat state. And here on first read compaction seems to be more about changing client state independently from UI state. 

Anyway, we can work this out but I do want to find the minimal immediate fix for anyone using the history with `chat_server()`.

gadenbuie: For posterity I mentioned on Slack that `set_messages()` makes sense to me if we think we can get it together quickly

> I was advocating for deprecating `.clear()` in favor of a new `.set_messages()` method

I guess my last comment is just that if we're thinking of deprecating `.clear()`, I'd like to consider leaving it as an alias for `.new_chat()` and deprecating internal arguments. That would give most general-case users a chance to keep using `.clear()` without breaking their apps and we could the more advanced usecase users migrate in the right direction.

cpsievert: > Are you okay moving forward without working this out yet? I'm on board with the general idea but I think we'll need to think about it a little more carefully and make sure we're getting the API right in tandem with ellmer/chatlas APIs.

Yep, and the more I looked into this, the more I'm happy with the PR as it currently stands (i.e.., hard error when `history` is enabled, but no deprecation, at least for now). 

I think `.set_messages()` is probably worth doing, but I also think it should probably come after #379, especially if we want it to automatically save to history before mutation of the current conversation.

cpsievert (approved):

PR #401, opened 2026-09-09
fix: remove attachment content from restored messages

gadenbuie:
Follow-up to #394.

## Summary

- Remove model-facing image, PDF, and text attachment content from restored message bodies.
- Preserve rich attachment previews and user-entered text.
- Repair records already saved with duplicated attachment content during replay.
- Cover mixed attachments and attachment-only messages in Python and R.

## Verification

Attach an image or text file, save the conversation, and restore it. The attachment preview remains, while the message body contains only the text entered by the user.

- `make py-check-unit FILTER='history'`
- `make py-check-types`
- `make py-check-format`
- `make r-check-tests FILTER='chat_history'`
- `make r-check-format`

### DIFF
--- a/pkg-py/src/shinychat/_history.py
+++ b/pkg-py/src/shinychat/_history.py
@@ -205,7 +205,11 @@ def _strip_attachment_content(
 ) -> StoredUiMessage:
     """Remove model-facing attachment content from stored user UI."""
     attachments = message.get("attachments")
-    if message.get("role") != "user" or not attachments:
+    if (
+        message.get("role") != "user"
+        or not attachments
+        or message.get("attachment_content_stripped") is True
+    ):
         return message
 
     segments = [dict(segment) for segment in message["segments"]]
@@ -261,6 +265,7 @@ def _strip_attachment_content(
 
     cleaned = dict(message)
     cleaned["segments"] = cast(Any, segments)
+    cleaned["attachment_content_stripped"] = True
     return cast(StoredUiMessage, cleaned)
 
 
@@ -751,7 +756,9 @@ class HistoryController:
                 _strip_attachment_content(message) for message in stored
             ]
             for message_dict in stored:
-                await self.chat._restore_bookmark_message(message_dict)
+                restored_message = dict(message_dict)
+                restored_message.pop("attachment_content_stripped", None)
+                await self.chat._restore_bookmark_message(restored_message)
                 restored_count += 1
         # The restored messages are already in the server-side accumulator, so
         # start the offset after the messages restored into the record.
--- a/pkg-py/src/shinychat/_history.py
+++ b/pkg-py/src/shinychat/_history.py
@@ -5,7 +5,12 @@ import dataclasses
 import warnings
 from typing import TYPE_CHECKING, Any, Awaitable, Callable, Literal, cast
 
-from ._attachments import Attachment, validate_attachments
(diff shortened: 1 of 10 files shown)


pkg-py/src/shinychat/_history.py
```diff
@@ -195,6 +200,75 @@ def derive_stored_ui_message(
     return _stored_ui_dict(as_stored_message(msg, session))
 
 
+def _strip_attachment_content(
```
cpsievert, line 207: Worth a short note here marking this as temporary — something like:

```python
# Stopgap: turn-derived UI mixes model-facing attachment content into user
# message bodies, so we re-serialize each attachment and subtract it here.
# The exchange-tree history rewrite captures user messages as displayed
# (typed text + attachment metadata) instead, making this unnecessary.
# Delete this and `attachment_content_stripped` when it lands.
```

And a twin above `strip_stored_attachment_content` in `pkg-r/R/chat_history_types.R`. Without a pointer, this reads as ordinary cleanup logic and will outlive its purpose.

pkg-py/tests/test_history_controller.py
```diff
@@ -2346,6 +2472,112 @@ async def test_replay_discards_old_format_ui_and_rederives_from_turns():
     ]
 
 
+@pytest.mark.anyio
+async def test_replay_repairs_attachment_content_in_existing_record():
```
cpsievert, line 2558: The tests in this PR construct v1 records and call v1 helpers directly, so they'll be rewritten when the exchange-tree history lands. Could we add one test at the behavior level — submit a message with an attachment, save, restore, assert the restored body is only the typed text with the preview intact — that doesn't touch record internals? That gives the guarantee a home that survives the architecture swap.

cpsievert (approved):
Approving for release. This fixes a real bug — restored user messages show the attachment's model-facing serialization duplicated against the preview — and the test matrix (image/text/PDF, attachment-only, replay repair) covers it well.

The mechanism is fragile by nature: it re-serializes each attachment and subtracts it from stored segments by suffix matching, so any change to `attachment_to_content` or the `\n\n` separator silently breaks the repair (the `except Exception` fallback hides that). Acceptable as a stopgap since it only touches user messages with attachments, but it shouldn't become permanent.

It won't need to. On `feat/history-exchange-tree`, user input is captured as the wire message (typed text + attachment metadata) and turns are stored separately, only to rebuild model context. UI is never derived from turns, so there's nothing to strip. When that lands, please delete the strip and the `attachment_content_stripped` flag outright.

pkg-py/src/shinychat/_history.py
gadenbuie: Added matching stopgap notes above the Python and R stripping helpers in 10466626. They point to feat/history-exchange-tree and explicitly call for deleting the stripping logic and attachment_content_stripped flag when it lands.

pkg-py/tests/test_history_controller.py
gadenbuie: Added test_attachment_message_round_trips_through_history in 10466626. It saves an attachment-bearing user turn, restores it through switch_to() on a fresh controller, and asserts only the user-visible contract: typed text remains the body and attachment metadata remains for the preview. It does not construct nodes, inspect record internals, assert schema versions, or call the stripping helper.

PR #403, opened 2026-09-09
feat: add configurable thinking display threshold

gadenbuie:
I think we reasonably default to showing all thinking, but I've found in practice that the thinking display is quite noisy and hides agentic work, especially in faster models like `GLM-5.3-Flash`. This is a small PR to give `chat_ui()` an option to delay showing thinking until it has taken "long enough", e.g. `show_thinking_after_s = 5` to suppress any thinking blocks that complete in less than 5 seconds.

It can be set to <0, e.g. `-1`, to hide all thinking, which might also be useful.

Thinking is always shown on restore because we don't currently persist timing data -- this will be fixed in the exchange tree work -- so on restore the options are show (>0) or hide (<0) the thinking blocks.

## Example

```r
library(ellmer)
library(shinychat)

chat <- chat_posit(model = "zai-org/GLM-5.3")
chat_app(chat, show_thinking_after_s = 5)
```

Try this short prompt, which should normally keep thinking hidden:

> What is the capital of France?

Then try this more involved prompt, which is more likely to show thinking after five seconds:

> You have 12 coins, one of which is counterfeit and may be either heavier or lighter than the others. Using a balance scale only three times, describe a strategy that always identifies the counterfeit coin and whether it is heavier or lighter.

## Summary

- Add `show_thinking_after_s` to the R and Python `chat_ui()` APIs, with a matching `show-thinking-after-s` client attribute.
- Delay, show, or suppress reasoning panels entirely in the browser without changing the server message wire shape.
- Bound positive thresholds to 60 seconds and ship rebuilt JavaScript assets for both packages.


pkg-py/src/shinychat/_chat.py
```diff
@@ -2654,6 +2662,7 @@ def chat_ui(
     toolbar_input: Optional[TagChild] = None,
     footer: Optional[TagChild] = None,
     tool_grouping: 'Literal["none", "tool", "all"]' = "tool",
+    show_thinking_after_s: float = 0,
     drawer: bool | ChatDrawer = True,
     show_history: bool = True,
```
cpsievert: Let's put this at the end (to match R)?

Also more generally I think we need to revisit the ordering of (keyword) parameters. Currently messages is first, but it's deprecated. Also feels like useful things like `drawer` should come earlier.

cpsievert (approved):

